PR #30, opened 2018-04-19
Set proper displayName for HOC wrappers

dbismut:
The `commonComponent` component blinds the `wrappedComponent` displayName. This PR should fix this.

**Before**
capture d ecran 2018-04-19 a 19 56 32

**After**
capture d ecran 2018-04-19 a 19 56 56

_**EDIT:** this is just a quick fix at the moment, since according to [React documentation best practices](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging) a HOC shouldn't have the same name as the wrapped component. Let me know if you'd like this PR to fix this as well._

### DIFF
--- a/samples/advanced-sample-react/src/enhancers/commonComponent.js
+++ b/samples/advanced-sample-react/src/enhancers/commonComponent.js
@@ -1,7 +1,15 @@
 import { withExperienceEditorChromes } from '@sitecore-jss/sitecore-jss-react';
 import extendStyles from './extendStyles';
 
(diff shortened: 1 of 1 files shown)


kamsar:
Appreciate the PR. The core code is maintained currently in an internal repo and we have merged this code into it (appending `CommonComponent` to the name so it's unique) - so it'll be in the next public release of JSS.

Thanks for helping to make JSS better :)

PR #41, opened 2018-08-08
Fix missing files

AlbertBrand:
For sitecore-javascript-renderings example


kamsar:
This is an experimental feature and after review we're not sure there should even be any manifest/disconnected setup for this sample at all. Since it's not a traditional JSS app, per se, it's a set of JS-rendered renderings placed within a Sitecore site, there is no disconnected mode.

I'll add a review to our backlog.

AlbertBrand:
Actually you can run `jss start` after these changes and you can review the components in 'disconnected' mode, however you need to do extra work to get it properly setup with a template etc.

kamsar:
Ok the docs (https://jss.sitecore.net/docs/techniques/mvc-integration/javascript-rendering) and the sample JS renderings app (https://github.com/Sitecore/jss/tree/master/samples/sitecore-javascript-renderings) have been updated to be more correct and helpful. Import of a JS renderings component library app is not going to be supported at launch - even if it is perhaps possible.

PR #57, opened 2018-09-14
Add JSX extention to component factory script

sowhatdoido:
Adding the ability to import components with `.jsx` file extension, since it's a common convention amongst React developers.


sowhatdoido:
I wasn't sure if you guys were open to using es6 syntax in these script files, since it would require later versions of nodejs. If you decide yes, I can update the fileExists function to use the spread operator instead of the arguments object.

kamsar:
`.jsx` extensions will be supported in the final release (merged internally)

Thanks for the contribution.

PR #83, opened 2018-11-29
Register multiple component in JssModule.forChild method

agalbulev:
## Description
Hi JSS team, I am Aleksandar and I work in Virtual Affairs. I saw that you now have a feature for lazy loading of component which is great. I try to made lazy loading module with multiple components which will be great feature for us but it not possible with current implementation. I made research in your code and found that this is possible with little change which can be make on JssModule.forChild function.

## Types of changes


packages/sitecore-jss-angular/src/lib.module.ts
```diff
@@ -81,14 +82,19 @@ export class JssModule {
   }
 
   /** Instantiates a module for a lazy-loaded JSS component */
-  static forChild(component: Type<any>): ModuleWithProviders {
+  static forChild(components: Array<Type<any>>): ModuleWithProviders {
```
kamsar: This would appear to be a breaking signature change (and would thus also break the existing angular sample app, which does not pass an array). Can this be implemented such that the array is optional?

agalbulev: Yes, I done this in next commit.

packages/sitecore-jss-angular/src/lib.module.ts
```diff
@@ -81,14 +82,24 @@ export class JssModule {
...
-  static forChild(component: Type<any>): ModuleWithProviders {
+  static forChild(components: Type<any> | Array<Type<any>>): ModuleWithProviders {
+    const providers: Provider[] = [
+      { provide: ROUTES, useValue: [], multi: true },
+    ];
+
+    if (Array.isArray(components)) {
+      components.forEach((component) => {
+        providers.push({ provide: ANALYZE_FOR_ENTRY_COMPONENTS, useValue: component, multi: true });
+        providers.push({ provide: DYNAMIC_COMPONENT, useValue: component });
```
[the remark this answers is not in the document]
kamsar: That makes sense. I vaguely recall attempting to prototype placing several lazy components in a module and there was some reason why it was less than simple.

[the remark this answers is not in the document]
agalbulev: Yes, you are right. Sorry for my mistake. I made other change and start to use ComponentNameAndType in forChild because I need to have a name of the component in jss-component-factory.service. With help of component name we success loaded the correct type like is made for loaded component.

agalbulev:
Do you think that I need to improve something more? I made several test about my implementation and I think that work fine. Can some one make review of changes?

kamsar:
@agalbulev I did a quick test and this change does not seem to break the existing lazy sample, at least.

It's not entirely clear to me what the procedure would be to register multiple components, and that needs to be documented and tested in order to merge. To make this simpler, we've moved the docs into this repo so that they can be directly updated as part of PRs. The relevant documentation [is here](https://github.com/Sitecore/jss/blob/master/docs/data/routes/docs/client-frameworks/angular/angular-tips/en.md#use-lazy-loaded-components-for-infrequently-used-jss-components) - please enhance it to provide examples of lazy loading multiple components. I've merged master into your PR so the docs are available on your branch.

agalbulev:
@kamsar I added the documentation and example.

docs/data/routes/docs/client-frameworks/angular/angular-tips/en.md
```diff
@@ -61,6 +61,55 @@ If you're using the default code-generated `app-components.module.ts` that's all
...
+@NgModule({
+  imports: [
+    // secret sauce
+    JssModule.forChild([
+      { name: 'FirstComponent', type: FirstComponent },
+      { name: 'SecondComponent', type: SecondComponent }
+    ])
+  ],
+  declarations: [
+    MyComponent,
```
kamsar: Wouldn't the declarations need to be `FirstComponent, SecondComponent`?

agalbulev: In this branch you can find my test site https://github.com/agalbulev/jss/tree/feature/test_multi_lazy. Can you send to me your configuration because I don't have this problem with my test site?

agalbulev: Can you also check that the component name 'StyleguideAngularLazyLoadingComponent' is equal to the property path for this component in 'app-components.module.ts'.

kamsar: Argh I'd pasted the name of the class into the `forChild()` registration complete with the `Component` suffix. Which the JSS component name didn't have.

Chalk another one up to Angular's totally amazing error messaging :trollface:

kamsar:
@agalbulev @spike1292 ran into an issue that seems strongly correlated with this PR that it seems to have broken AOT (`jss build`) compilation. Reproduction is pretty easy: make a new Angular app with the CLI and run `jss build`. It works fine with non-AOT builds (i.e. `jss start`).

https://sitecore.stackexchange.com/questions/15903/deploy-jss-app-on-sitecore-instance

I can get it to build if I move the `JssModule.forChild` function _into a variable_:

```js
import { NgModule } from '@angular/core';
import { JssModule } from '@sitecore-jss/sitecore-jss-angular';
import { StyleguideAngularLazyLoadingComponent } from './styleguide-angular-lazy-loading.component';
import { AppComponentsSharedModule } from '../app-components.shared.module';

// this seems to make AOT happy?!
const forChild = JssModule.forChild;

@NgModule({
  imports: [
    forChild(StyleguideAngularLazyLoadingComponent),
    AppComponentsSharedModule
  ],
  declarations: [
    StyleguideAngularLazyLoadingComponent,
  ],
})
export class StyleguideAngularLazyLoadingModule { }
```

Do you guys see any reason why that should suddenly be required? Certainly other modules are using `Module.forXXX()` functions and they have no issues with AOT - and this works fine if I downgrade to `jss-angular@11.0.0` without these changes.

agalbulev:
@kamsar Yes, today I find this problem too. I think this is a problem associated with an Angular AOT compiler. A found how we can fix the problem, without using fromChild() method into lazy loading module.
```
  {
      ngModule: JssModule,
      providers: [
        {
          provide: DYNAMIC_COMPONENT,
          useValue: {
            'FirstComponent': FirstComponent,
	    'SecondComponent': SecondComponent
          }
        },
        {
          provide: ANALYZE_FOR_ENTRY_COMPONENTS,
          useValue: [
		{ name: 'FirstComponent', type: FirstComponent },
      		{ name: 'SecondComponent', type: SecondComponent }
	  ]
          multi: true
        }
      ]
    }
```

Only we need to exprot DYNAMIC_COMPONENT provider.

PR #92, opened 2018-12-11
Alter typing to enable strong typed placeholder names

### LINKED ISSUE #78: FR: Allow us to type the PlaceholdersData name
Right now the `PlaceholdersData` interface from `@sitecore-jss/sitecore-jss/types/dataModels.d.ts` is defined as:

```ts
export interface PlaceholdersData {
    [name: string]: Array<ComponentRendering | HtmlElementRendering>;
}
```

So any `name` is a valid placeholder name.  
This is already quite nice however it could be improved so we can keep placeholder names maintainable even for large projects.

The main advantage is that it would allow typescript to check if a name can be used for this component and the user gets faster access to the information about which placeholder name is usable in the current context.

In the following example it allows only to use the name "jss-reuse-example":

placeholder

Hovering will even tell you that "unknown-placeholder-name" is not supported for this component.

placeholder2

## Describe the solution you'd like

Changing the type just a little bit will allow it to be used as a generic:

```ts
export type PlaceholdersData<K extends string = string> = {
  [P in K]: Array<ComponentRendering | HtmlElementRendering>;
}
```

`K extends string = string` means that you can provide all allowed names (but it's optional and if you don't provide anything it will fallback to the current behaviour - so this would be backwards compatible)

Usage of that type:

```ts
// Allow only jss-placeholder-one or jss-placeholder-two
type DemoRestrictedPlaceholdersData = PlaceholdersData<'jss-placeholder-one' | 'jss-placeholder-two'>;
```

## Additional information

A full example including a very basic example for the `PlaceholderProps` can be seen here: https://codesandbox.io/s/0100y49vn

kamsar:
Fixes #78

@jantimon can you take a look at this? The sandbox on #78 was a 404.

This does not break the compilation of downstream consumers in TS (angular sample app, react lib, angular lib)

## Types of changes

### DIFF
--- a/packages/sitecore-jss/src/dataModels.ts
+++ b/packages/sitecore-jss/src/dataModels.ts
@@ -48,8 +48,8 @@ export interface RouteData {
  * Placeholder contents data (name: placeholder name, then array of components within that placeholder name)
  * Note: HtmlElementRendering is used by Sitecore Experience Editor
  */
-export interface PlaceholdersData {
-  [name: string]: Array<ComponentRendering | HtmlElementRendering>;
+export type PlaceholdersData<TYPEDNAME extends string = string> = {
+  [P in TYPEDNAME]: Array<ComponentRendering | HtmlElementRendering>;
 }
 
 /**


jantimon:
Wow cool thanks @kamsar 👍

Issue #80, opened 2018-11-23
createDefaultDisconnectedServer expects appRoot to be set to the script folder


jantimon:
## Description

Right now the createDefaultDisconnectedServer is expecting the "root" not to be set to the root of the project but to the scripts directory:

https://github.com/Sitecore/jss/blob/0292a05077c4ef21b668eb9c30412f57620c3779/samples/react/scripts/disconnected-mode-proxy.js#L18-L19

This is also hard coded inside the `createDefaultDisconnectedServer` (Version 9.0.6) code:

`@sitecore-jss/sitecore-jss-dev-tools/dist/disconnected-server/create-default-disconnected-server.js` line 52, 53:

```
app.use('/assets', express_1.default.static(path_1.join(options.appRoot, '../assets')));
app.use('/data/media', express_1.default.static(path_1.join(options.appRoot, '../data/media')));
```

As you can see it has to add `..` to get rid of the scripts folder.

kamsar:
As mentioned in https://github.com/Sitecore/jss/issues/79#issuecomment-441726735 these are default conventions of a JSS app. If you want to override those conventions, expand `createDefaultDisconnectedServer()` and tweak away. The use of `appRoot` is a confusing name here, yes.

jantimon:
Is there any reason why you chose to use the script folder over the actually root directory as appRoot?

I guess this is very unintuitive for new developers but probably easy to work around.

kamsar:
It was a bad naming choice. Originally `createDefaultDisconnectedServer` was expanded in every sample app, and when it was extracted so it could be shared for maintainability that happened.

It may be worth investigating fixing this up after release to be less confusing, assuming we can do it in a non-breaking way.

jantimon:
Cool thanks for explaining the history 👍

kamsar:
This has been clarified and made to work by default using the _actual_ appRoot in #94

PR #94, opened 2018-12-11
Make disconnected appRoot mean appRoot not scriptsRoot

kamsar:
## Description
Per #80 the appRoot of the proxy actually meant "/config", which makes no sense.

This commit causes the behaviour of appRoot to expect the actual app root path, in such a way that existing JSS apps will continue to work unchanged as well (each sample app was tested with old and tweaked configs in disconnected-mode-proxy)

## Motivation
Removes principle of least surprise violation

## How Has This Been Tested?
Ran each sample app with old and new proxy config to verify backwards compatibility.

## Types of changes

### DIFF
--- a/packages/sitecore-jss-dev-tools/src/disconnected-server/create-default-disconnected-server.ts
+++ b/packages/sitecore-jss-dev-tools/src/disconnected-server/create-default-disconnected-server.ts
@@ -10,8 +10,7 @@ export interface DisconnectedServerOptions {
   appName: string;
 
   /**
-   * Root physical path to the app's /scripts folder
-   * (i.e. __dirname from the calling script to start the middleware)
+   * Root physical path to the app (i.e. where your package.json is)
    */
   appRoot: string;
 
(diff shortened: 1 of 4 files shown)


packages/sitecore-jss-dev-tools/src/disconnected-server/create-default-disconnected-server.ts
```diff
@@ -94,6 +94,21 @@ export function createDefaultDisconnectedServer(options: DisconnectedServerOptio
...
+  if(options.appRoot.endsWith('scripts')) {
+    options.appRoot = join(options.appRoot, '..');
+  }
+
+  // further backwards compatibility for #80
+  // allows apps with GA watch path of '../data' (relative to /scripts)
+  // to keep working even with appRoot now relative to the actual app root
+  // We do this by stripping '../' from path leads, making the path './data' instead - theoretically, the chance of
+  // wanting to actually escape from the app root entirely otherwise is awfully low.
+  options.watchPaths = options.watchPaths.map(path => path.startsWith('../') ? path.substring(1) : path);
```
aweber1: Are `watchPaths` always relative to `appRoot`? If so, might be worth adding a note to the `watchPaths` option comment (line 18) that they're expected to be relative.

kamsar: They can be but don't have to be (i.e. `c:\foo` would also be valid) - but if they are, they are app-root-relative, yes.

kamsar: Added the comment. Good suggestion.

PR #95, opened 2018-12-12
Modify SSR handling in React and Vue

### LINKED ISSUE #76: FR: Allow to use ContentSecurityPolicy without 'unsafe-inline'
Hey

this proposal is a little bit ambitious but it comes with a lot of advantages.

## Summary

Right now the `jss` boilerplate uses a script tag (`<script>__JSS_STATE__=...</script>`) to communicate between backend and frontend.

I would propose to change this a little bit to the following structure:

```
<script data-name="JSS_STATE" type="text/json">
{...}
</script>
```

This technique is described in different blog articles - this one is quite nice to read:
https://quipblog.com/efficiently-loading-inlined-json-data-911960b0ac0a

### Disadvantages

- According to [serialize-javascript json](https://www.npmjs.com/package/serialize-javascript#optionsisjson) `state` must not function or regexps values.

### Advantages

- Faster state generation. According to [serialize-javascript json](https://www.npmjs.com/package/serialize-javascript#optionsisjson) serialization is 3 times faster on the node js side
- According to [https://quipblog.com](https://quipblog.com/efficiently-loading-inlined-json-data-911960b0ac0a) even faster in frontend - (however I am not sure if this is still true)
image

- Removes one security vector as the `<script type="text/json">` is not able to execute javascript ([the script tag is ignored](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-type))
- Adds the ability for additional security hardening by using the [CSP](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src) without `'unsafe-inline'`.
- Adds more flexibility as it would allow to have multiple react widgets on one page (E.g. one for a complex navigation and one for the body) 

## How would that work on the frontend side?

Lets assume that the backend would generate the following markup

```html
<script data-name="JSS_STATE" type="text/json">
{...}
</script>
<div data-name="jss-root" data-app="Navigation">server side rendered markup</div>
```

The frontend code could be for the `Navigation` (maybe that part could also be reusable):

```js
const jssRoots = document.querySelectorAll('[data-name="jss-root"][data-app="Navigation"]');
for (let i = 0; i < jssRoots.length; ++i) {
  const jssRoot = jssRoots[i];
  const jssConfigScript = jssRoot.previousElementSibling;
  const jssInitialState = JSON.parse(jssConfigScript.innerHTML);
  // Hydrate the app e.g. with react:
  ReactDOM.hydrate(React.createElement(Navigation, jssInitialState), jssRoot);
}
```

kamsar:
* Fixes #76
* Allows for use of CSP without unsafe-inline
* Faster parse times for the SSR data
* Better safety by not allowing injection of JS functions, etc into the page only data
* This is an optional enhancement to the sample apps, and no library changes were needed
* Both old and new styles of SSR will continue to work fine

## How Has This Been Tested?

e2e tests have passed in integrated and disconnected modes, visually inspected and checked for JS errors

## Types of changes

### DIFF
--- a/samples/react/public/index.html
+++ b/samples/react/public/index.html
@@ -22,9 +22,7 @@
     </noscript>
     <div id="root"></div>
 
-    <script>
-      window.__JSS_STATE__=null;
-    </script>
+    <script type="application/json" id="__JSS_STATE__">null</script>
 
     <!--
       This HTML file is a template.
--- a/samples/react/server/server.js
+++ b/samples/react/server/server.js
@@ -94,9 +94,10 @@ export function renderView(callback, path, data, viewBag) {
         // write the string version of our state
         html = assertReplace(
           html,
-          '__JSS_STATE__=null',
-          `__JSS_STATE__=${serializeJavascript(state)}`
+          '<script type="application/json" id="__JSS_STATE__">null',
+          `<script type="application/json" id="__JSS_STATE__">${serializeJavascript(state)}`
         );
+
         // render <head> contents from react-helmet
         html = assertReplace(
           html,
--- a/samples/react/src/index.js
+++ b/samples/react/src/index.js
@@ -14,16 +14,24 @@ let renderFunction = ReactDOM.render;
 /*
   SSR Data
   If we're running in a server-side rendering scenario,
-  the server will provide the window.__JSS_STATE__ object
+  the server will provide JSON in the #__JSS_STATE__ element
   for us to acquire the initial state to run with on the client.
 
   This enables us to skip a network request to load up the layout data.
+  We are emitting a quiescent script with JSON so that we can take advantage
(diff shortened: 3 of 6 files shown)


jantimon:
Wow that looks amazing! 👍

You should probably use https://github.com/yahoo/serialize-javascript#optionsisjson

E.g.

`<script type="application/json" id="__JSS_STATE__">${serializeJavascript(state), {isJSON: true}}`

instead of 

`<script type="application/json" id="__JSS_STATE__">${serializeJavascript(state)}`

This will be alot faster:

> This option is a signal to serialize() that the object being serialized does not contain any function or regexps values. This enables a hot-path that allows serialization to be over 3x faster. If you're serializing a lot of data, and know its pure JSON, then you can enable this option for a speed-up.

PR #99, opened 2018-12-13
Azure pipelines #2

aweber1:
DevOps is _super_ exciting and not the least bit annoying.


aweber1:
codeship, stop it. just stop.

PR #100, opened 2018-12-13
Azure pipelines #3

aweber1:
will codeship finally be gone?


aweber1:
nope.

aweber1:
But Azure pipelines is winning.

Issue #90, opened 2018-12-11
Angular Scaffolding is broken


kamsar:
## Description
Scaffolding with `jss scaffold` in Angular project is currently non-operational.

Refer to this SO question https://sitecore.stackexchange.com/questions/15464/error-when-trying-to-scaffold-an-angular-component

## Steps To Reproduce

1. `jss create foo angular`
1. `jss scaffold Foo`

## Feedback Wanted

The scaffolding system seems to be broken due to major breaking changes in the Angular Schematics module. JSS Angular scaffolding is based on the `ng generate component` schematic, which is unfortunately not particularly extensible and there's more code copying than I'd like.

There are a couple ways we could go about fixing this, and I'd like to hear from those who care with input on it:

### Stop using schematics

This is not the first time Angular breaking changes have caused problems for JSS. Given the simple nature of JSS scaffolding (and aim to allow users to customize it), we could dump schematics and use pure JS like the React and Vue samples do. This would lose the extra parameters that come with the component scaffold. It would also improve our ability to support multiple Angular versions as we would not be tied to breaking changes in the schematics module.

### Scaffold only the disconnected definition

We could make `jss scaffold` for Angular scaffold only the disconnected definition file. You'd use `ng generate component` directly for the rest of it.

### Some combination of the above

We could look into making `jss scaffold` scaffold the disconnected file and then invoke `ng generate`, noting that there wouldn't be any JSS-specific tweaks to the generated components.

JonK723:
I think the third option of scaffolding the disconnected definition and invoking `ng generate` aligns better with what it looks like the React and Vue scaffold commands do. I personally don't mind doing a little more manual work to connect my component to the rest of my application as long as I'm not tied to a feature that has the possibility of breaking everything.

kamsar:
#106 fixes the scaffolding itself without any semantic changes

x1c0:
Hello,

Not sure if I should be writing on this close issue, but I just fresh install everything following the guide [https://jss.sitecore.com/docs/getting-started/first-component](url) and I have the following error:

![image](https://user-images.githubusercontent.com/2421483/50761443-affdf400-126a-11e9-877b-802ca0ba4ab5.png)

Is this related? should I open a new issue?

x1c0:
ah this fix was only merged to Dev, when are you releasing master to npm?

Thanks in advance!

x1c0:
[...]

PR #106, opened 2018-12-13
Fixes angular scaffolding

kamsar:
Ref #90

Schematics have an issue where depending on the version of schematics in use there are different errors when one attempts to compose a scaffold of multiple template roots - in this case, a template for the manifest and another for the component.

With schematics 0.6 (angular 6) this results in an undefined entries. With schematics 7, it results in an infinite loop.

This commit works around that by not rebasing the manifest template and switching the manifest template to have a full path to the file. Doing this removes the ability to variableize the `manifestPath` but it does allow the scaffold to actually work. This also updates schematics to refer to schematics for angular 7+, which shipped with GA.

## Motivation

Scaffolding in Angular, without `--lazyload`, is currently broken.

## How Has This Been Tested?

Ran scaffolding lazy and not lazy against the angular sample app in the monorepo.

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)


kamsar:
All things are possible with Angular :trollface:

kamsar:
@spike1292 does this seem reasonable to you? Not going to get in your way?

spike1292:
> @spike1292 does this seem reasonable to you? Not going to get in your way?

We store the definitions next to the component, so this is a breaking change for use. This can be solved by moving the manifest manually.

The scaffold is now broken. So choosing between not working and a manual step. I prefer the manual step😉

kamsar:
@spike1292 as a possible workaround you could scaffold with `--noManifest` and then scaffold that on the side with a different script, if you wanted to remove the manual step. React/Vue's scaffold script has stealable code.

PR #116, opened 2019-01-03
Master

Eugenio161288:
Here's an initial pr to add a feature to modify layout service data.

There are 2 new parameters in proxy config. They are "**useProxyUrls**" and "**proxyHost**".
If the **useProxyUrls** is enabled, then all **apiHost** URLs are replaced by **proxyHost**, otherwise everything should work as before

Please review that, will be happy to get any inputs to improve my solution, as it looks quite ugly at the moment


aweber1:
Thanks for the PR! We are working on this feature but it is part of a larger refactoring effort for the node SSR proxy, so the approach will be different from the approach in this PR. As such, I'm closing the PR. Thanks again for submitting.

PR #118, opened 2019-01-04
Add beforeMiddlewareRegistered hook to DisconnectedServerOptions

coreyasmith:
Add `beforeMiddlewareRegistered` hook to `DisconnectedServerOptions` that allows for the registration of middleware that should execute before the disconnected server middleware.

## Motivation
Some middleware needs to be registered before the disconnected server middleware, like `body-parser` and `cookie-parser`. For example, the `customizeRoute` method on `DisconnectedServerOptions` gives you access to the request and response objects, but to access the cookies on the request object, you need to register `cookie-parser` with Express so it executes before the `customizeRoute` method. Currently if you want to do this you have to add the Express package to your project, create an instance of Express, register your middleware, and pass the Express instance to `DisconnectedServerOptions`:

```javascript
// ... omitted for brevity ...
const app = Express();
app.use(cookieParser());

const proxyOptions = {
  // ... omitted for brevity ...
  server = app,
  customizeRoute: (route, rawRoute, currentManifest, request) => {
    // do something with request.cookies
  }
};
// ... omitted for brevity ...
```

It'd be nice to be able to register middleware that executes before the disconnected server middleware without having to include the `express` package in your project and new up your own instance:

```javascript
// ... omitted for brevity ...
const proxyOptions = {
  // ... omitted for brevity ...
  beforeMiddlewareRegistered: (app) => {
    app.use(cookieParser());
  },
  customizeRoute: (route, rawRoute, currentManifest, request) => {
    // do something with request.cookies
  }
};
// ... omitted for brevity ...
```

## How Has This Been Tested?
Updated `disconnected-mode-proxy.js` in the React sample app to reference `'../../../packages/sitecore-jss-dev-tools'` and verified that the code above works.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sitecore-jss-dev-tools/src/disconnected-server/create-default-disconnected-server.ts
+++ b/packages/sitecore-jss-dev-tools/src/disconnected-server/create-default-disconnected-server.ts
@@ -47,6 +47,12 @@ export interface DisconnectedServerOptions {
   sourceFiles?: string[];
 
   /**
+   * Hook function that is called before the disconnected server middleware is registered with the server.
+   * Useful to add your own middleware before the disconnected middleware.
+   */
+  beforeMiddlewareRegistered?: (server: any) => void;
+
+  /**
    * Hook function that is called after the disconnected server middleware is registered with the server,
    * but before the server starts listening. Useful to add your own middleware after the disconnected middleware.
    */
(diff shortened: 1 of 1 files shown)


kamsar:
Given that an existing method, albeit one you're less a fan of, allows this customization I'm inclined to decline this PR as it adds additional complexity to offer a second path to do the same thing (that then must be maintained, tested, documented, and a user must then choose one of the ways). 

FWIW `express` is already transitively in your node_modules because it's used as the default server in the library.

PR #142, opened 2019-01-25
Add new link directive which work with angular route

agalbulev:
Add new link directive (scRouterLink) which work with angular route. This will prevent refresh the whole page and lost of state of app after refresh.

## Description
Extended the LinkDirective class and overwrited only one method renderTemplate(). New directive has access to Router Service and on click make navigation using navigate() method of Route service.

## Motivation
Prevent refresh of whole page and lost of state of app after refresh. 

## How Has This Been Tested?
Can be test in sample app after change one of *scLink with *scRouterLink. Open the DevTool and check the network tab, In the tab need to see, after click over navigation link,  only one call to Layout Service without call for main.js, vendor.js and other and also the loader for browser tab need to missing.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sitecore-jss-angular/package.json
+++ b/packages/sitecore-jss-angular/package.json
@@ -13,6 +13,10 @@
     "name": "Sitecore Corporation",
     "url": "https://jss.sitecore.net"
   },
+  "contributors": [{
+    "name": "Alexandar Galbulev",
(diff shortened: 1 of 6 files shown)


kamsar:
This has been merged with some fixes added (keeps the href attribute, adds doc note, and added a sample to the styleguide). Thanks for the contribution!

PR #146, opened 2019-01-30
Added support for TypeScript when 'jss build'ing

boro2g:
I've not added in 
```
config.module.rules.push({
    test: /\.(ts|tsx)?$/,
    loader: 'ts-loader',
    exclude: /node_modules/,
    options: {
      appendTsSuffixTo: [/\.vue$/],
    }
  });
```
as I figured people probably wanted to opt into that. 

If adding ^ we'd need to update package.json / package-lock.json with **ts-loader** and **typescript** and also add to webpack override: `config.resolve.extensions.push('.ts');`

### DIFF
--- a/samples/vue/server/server.vue.config.js
+++ b/samples/vue/server/server.vue.config.js
@@ -38,7 +38,7 @@ const vueConfig = {
     */
 
     config.module.rules.unshift({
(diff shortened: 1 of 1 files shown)


sc-dawidrutkowski:
@boro2g I've merged your pull request.

PR #155, opened 2019-02-13
Placeholder loaded event

Idicious:
While implementing lazy loading with Sitecore JSS we ran into the issue that there was no way to know when a placeholder was done loading its children. In order to solve this issue we implemented the following solution. 

## Description
This change adds a `loaded` event to the placeholder component. This is emitted once all its child components are loaded.

## Motivation
Our usecase for this was related to analytics. We allow modules to register analytics enhancers which can augment the default analytics data. The basic analytics implementation was based on router events, the problem we ran into when implementing lazy loaded modules was that the router event fired before the lazy module was loaded. 
Because of this, on page load, the event was completed before the lazily loaded enhancer was registered. We use this event to dispatch an action to the store, and zip that action from the action stream together with the router event.

## How Has This Been Tested?
We've tested that our above usecase is solved with the help of this change.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sitecore-jss-angular/src/components/placeholder.component.ts
+++ b/packages/sitecore-jss-angular/src/components/placeholder.component.ts
@@ -4,12 +4,14 @@ import {
   ComponentFactoryResolver,
   ContentChild,
   DoCheck,
+  EventEmitter,
   Inject,
   Input,
   KeyValueDiffer,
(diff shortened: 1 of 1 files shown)


kamsar:
Tested and added a sample to the styleguide. Thanks for the contribution :)

PR #168, opened 2019-04-01
Improve tree shaking for @sitecore-jss/sitecore-jss-react

jantimon:
## Description

This pull request is rather a proposal than a feature complete code change.

Right now `@sitecore-jss/sitecore-jss-react` is compiled to `commonjs`.  
In general that is a really good idea as commonjs is very compatible with javascript bundlers.  

From [https://unpkg.com/@sitecore-jss/sitecore-jss-react@11.0.2/dist/index.js](https://unpkg.com/@sitecore-jss/sitecore-jss-react@11.0.2/dist/index.js):

```js
"use strict";
Object.defineProperty(exports, "__esModule", { value: true });
var sitecore_jss_1 = require("@sitecore-jss/sitecore-jss");
exports.dataApi = sitecore_jss_1.dataApi;
exports.mediaApi = sitecore_jss_1.mediaApi;
...
```

However tree shaking will not work in that case as webpack will not be able to detect if sitecore-jss-react includes any side effect and therefore will keep the entire code of all dependencies to stay safe.

[The author of Rollup](https://github.com/rollup/rollup/wiki/pkg.module) suggested in 2017 to bundle modules twice - once tree shakable using esnext and once using common js.

This pull request only changes the build from commonjs to esnext to keep the change as small as possible. Please challenge this idea and let me know what you think. 

Merging this pull request would change the file format to the following:

```
export { Placeholder } from './components/Placeholder';
export { Image } from './components/Image';
export { RichText } from './components/RichText';
export { Text } from './components/Text';
...
```

Just as a site note: For webpack there might be an [alternative solution]( https://webpack.js.org/guides/tree-shaking/#mark-the-file-as-side-effect-free) however I didn't look into it to much.

## Motivation

**Decrease the size of @sitecore-jss/sitecore-jss-react by 70% in our project**. (8kb instead of 27kb)

## How Has This Been Tested?

Only in our environment

## Types of changes

This will break any non esmodule compatible bundler.  
Unfortunately I just realized that this also broke your mocha setup.

So maybe we should release both versions as [written here](https://github.com/rollup/rollup/wiki/pkg.module#wait-it-just-means-import-and-export--not-other-future-javascript-features).

What do you think?

## Checklist:

### DIFF
--- a/packages/sitecore-jss-react/tsconfig.json
+++ b/packages/sitecore-jss-react/tsconfig.json
@@ -1,7 +1,7 @@
 {
   "compilerOptions": {
     "target": "es5",
-    "module": "commonjs",
+    "module": "ESNext",
     "newLine": "LF",
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,


aweber1:
Thanks for starting this discussion. We've had an item in the backlog to re-examine the "tree shake-ability" of the various JSS component libraries. Just wanted to let you know it's on the radar but due to human resource constraints we simply haven't had the time to look into it. 

We would definitely want to minimize the impact (if any) on existing implementations, so shipping with multiple targets may be appropriate if there isn't a simpler way. Again, it just requires more exploration that would have to include the Vue and Angular libraries as well.

In the meantime, you could _perhaps_ try importing components directly from their package path:
```
import { Placeholder } from '@sitecore-jss/sitecore-jss-react/dist/components/Placeholder';
```

That should help prevent webpack from importing any dead code into your bundle. I fully realize that's not ideal in any way, but might be good enough as an interim solution.

Along those same lines, perhaps creating a thin import wrapper in your app around the various JSS components could ease the refactoring, e.g.
```
// my-import-wrapper.js
export { Placeholder } from '@sitecore-jss/sitecore-jss-react/dist/components/Placeholder';
export { Text } from '@sitecore-jss/sitecore-jss-react/dist/components/Text';

// my-component.js
import { Placeholder, Text } from './my-import-wrapper';
```

Again, I understand it's not ideal, but at the moment I can't commit to a timeline for exploring the issue further (for all 3 libs).

I'll leave the PR open for further discussion and awareness (or if anyone wants to volunteer to take on the effort).

sc-antonkulagin:
Thank you for your suggestion. We are still  investigating it. For now to limit the number of open pull request we had to close it. Feel free to create an issue in the repo, so we can keep track of it better

PR #177, opened 2019-05-01
Feature/connected demo

JohnLynnBeck:
Adding documentation to the docs app for JSS Connected Demo.


docs/package.json
```diff
@@ -68,6 +68,7 @@
     "react-helmet": "^5.2.0",
     "react-router": "^4.3.0",
     "react-router-dom": "^4.3.1",
+    "react-router-hash-link": "^1.2.1",
```
anastasiya29: Do we still need to add this new module? I believe we ended up removing all the hash links.

docs/assets/css/sidenav.css
```diff
@@ -45,4 +45,73 @@
     height: 25px;
     color: black;
     font-weight: bold;
+}
+
```
anastasiya29: Originally the new sidenav had different styling from the docs sidenav, but the end result is that both side navs use the original styling. Can you confirm that this new css is still needed? Or is this leftover and needs to be removed?

docs/package.json
JohnLynnBeck: Ah, after the sidenav reorg, we do not. I'll remove.

docs/assets/css/sidenav.css
JohnLynnBeck: Definitely not needed. Thought I'd deleted it already, as a matter of fact.

aweber1:
I'm going to be super annoying, but can you close this PR and branch off of `master` instead of `dev` for these changes? Then open a PR to merge into `master`.

I think we need to change the process for making non-feature-related docs changes.

aweber1:
Closing in favor of #180

PR #219, opened 2019-07-19
[Angular] Add directive for rich texts which allows routerlinks that can be both internal and external

jlast:
Add directive for rich texts which allows routerlinks that can be both internal and external.
This allows Content Editors to add internal links to other pages, while still being able to maintain external links.

Also add a small change so linkfield classes will be merged instead of replaced. 

## Description
Add directive for rich texts which allows routerlinks that can be both internal and external.
This allows Content Editors to add internal links to other pages, while still being able to maintain external links.

## Motivation

Currently the implementation of a rich text is limited. All links won't use angular routerlinks.. This directive decides whether the link has a protocol and is therefor an external link.

## How Has This Been Tested?
Added unit tests and tested it on our application.

## Types of changes

## Checklist:


packages/sitecore-jss-angular/src/components/link.directive.ts
```diff
@@ -67,7 +71,11 @@ export class LinkDirective implements OnChanges {
       ...this.getElementAttrs(),
       ...this.attrs,
     };
-    Object.keys(attrs).forEach((key) => this.renderer.setAttribute(span, key, attrs[key]));
+    Object.keys(attrs).forEach((key) => {
```
spike1292: could also use Object.entries, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries

tslib is set on es2017 so entries should be available , https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss-angular/tsconfig.json

jlast: Thanks for the suggestion. I consider this change out of scope for the current PR, I will raise an issue for this.

PR #230, opened 2019-08-07
Feature/http render engine

sc-dawidrutkowski:
Merging `http-render-engine` branch to `dev` after fixing outdated tests and making sure that all of them are passing.


jbreuer:
I assume this is the JSS Rendering Host demo that was also given at SUGCON? https://twitter.com/j_breuer/status/1154653858175602690

Any idea when the next version of JSS with this feature will be released?

aweber1:
@jbreuer with Sitecore 9.1.1 and 9.2, the Sitecore server-side bits are ready to be used but client side is missing (as you know). Until the JSS team has more time and is able to finalize the client-side bits, I created a gist to demonstrate a "basic" rendering host setup: 

https://gist.github.com/aweber1/39d1e9e07fb19aac9bc9b94feb87bb0f

PR #269, opened 2019-09-30
Upgrade Vue sample

sc-illiakovalenko:


samples/vue/src/lib/GraphQLClientFactory.js
```diff
@@ -1,5 +1,3 @@
-/* eslint-disable import/first */
```
sc-illiakovalenko: Currently Eslint don't throw errors without this rule

PR #289, opened 2019-11-11
Remove SitecoreForm import from JSON example

erzr:
`SitecoreForm` is not used and the guide does not tell you to install `@sitecore-jss/sitecore-jss-forms` yet so the example will throw a compilation error during `jss start` if you have not done that yet. If `SitecoreForm` just being used for typing (here: /** @type {SitecoreForm} */), maybe move the NPM package section to the beginning of the guide instead of in the middle.


kamsar:
Yep it's being used for JSDoc typing

sc-illiakovalenko:
@erzr I moved `Installation` section above sample implementation section

Issue #299, opened 2019-12-15
SitecoreContextReactContext.Provider is not working properly, because value never changes as it is always same class instance


M1r1k:
## Description
`SitecoreContext` class provide sitecore context via `<SitecoreContextReactContext.Provider value={this.contextFactory}>` but as suggested in samples implementation  is Singleton, default React's Object.is() comparison is not working when `SitecoreContextFactory.setSitecoreContext` is called and Provider's children are not rerendered.

## Expected behavior
All consumers be updated if context value is modified

## Steps To Reproduce

1. Start `./samples/react` and go to http://localhost:3000/styleguide/custom-route-type for example.
2. Add this line to `./src/AppRoot.js` - `window.debug1 = SitecoreContextFactory;` - for debugging purpose later
3. Now on "Custom route type" page we can trigger setSitecoreContext from the console
4. Run this code
```
debug1.context.route.fields.headline.value = "Reproduced"
debug1.setSitecoreContext(debug1.context)
```
This has to update `headline` property's value that should effect our page's tittle. But it does not.
However if you change `<SitecoreContextReactContext.Provider value={this.contextFactory}>` to 
`<SitecoreContextReactContext.Provider value={Object.assign( Object.create( Object.getPrototypeOf(this.contextFactory)), this.contextFactory)}>` it will work fine.

## Possible Fix
We have to use React best practice and pass simple objects to Provider value, like this.contextFactory.context

## Your Environment
* JSS Version: 13.0.0

sc-illiakovalenko:
Fixed in #354

PR #301, opened 2019-12-15
Switch to pure object context value instead of class instance

M1r1k:
See #299 for details


sc-illiakovalenko:
@M1r1k I pulled your commit and tried to reproduce this issue,

1. Added `window.debug1 = SitecoreContextFactory;` to `./src/AppRoot.js`
2. Go to `/styleguide/custom-route-type`
3. Open the console and run: 
```
debug1.context.route.fields.headline.value = "Reproduced"
debug1.setSitecoreContext(debug1.context)
```
Result: Value is changed, but new value is not rendered on the page, because
`Styleguide-CustomRouteType` didn't get new props with `context` that contains new value.

sc-illiakovalenko:
@M1r1k If you will have any updates, please, feel free
to open PR

PR #327, opened 2020-02-19
Update en.md

alvarocjunq:
Changed propose, to understand better what the significate of SSR

## Description
When we look to the letters: SSR
Commonly we think In Server Side Rendering, and the description of that feature really talked about this.
Make sense the title have the same objective

## Types of changes

## Checklist:


sc-illiakovalenko:
@alvarocjunq server side routing and server side rendering are two different functions. one is about how the framework handles changes to the url, and the other is how the framework handles processing the JS & generating the html. In our case this article is describing how framework handles changes to the url

Issue #141, opened 2019-01-24
BUG: Cannot add components to placeholder without saving first


jeppeskovsen:
## Description
In the Experience Editor when I add a component containing a placeholder, and I try to add a component to that, I get this error:
![image](https://user-images.githubusercontent.com/25830173/51677792-02fdd800-1fdb-11e9-8ebc-f0d2d55b5a55.png)

To make it work, I have to add the component and save. Then I can add components to it's placeholder. It's like it doesn't register that the component I added contains a placeholder until I save.
Will also mention that my list of allowed controls does not work either, when this error occurs.

## Your Environment

* Sitecore Version: 9.1
* JSS Version: 11.0.2

sc-illiakovalenko:
@jeppeskovsen Fixed in #343

PR #343, opened 2020-03-12
Fix bug (#141) XP editor for react requires save and refresh before adding new renderings

vsirakova:
Fixes issue #141: when working in XP editor in the react version, after adding a rendering with a placeholder, another rendering cannot be added in the freshly loaded placeholder but save and refresh of the page is required

## Description
The issue is caused by the fact that the very important key attribute of the `<code>` DOM element that represents the placeholder is missing after SSR. Because it is a react special prop it cannot be set directly as a prop and then transferred as an attribute and that's why in the sitecore jss library for react this problem is solved by adding the key attribute by directly modifying the DOM on the client side. But after a rendering is freshly added through SSR, it is not hydrated and the lifecycle methods are not running, so the key attribute is never added. 
The fix replaces phkey attribute with key attribute directly in the html that is about to be sent to XP editor to inject. 

## Motivation
It is quite annoying to have to save your page on each step if you want to add renderings within other renderings. It is much faster and smoother to be able to add multiple renderings in one go.

Fixes issue #141 

## How Has This Been Tested?
Tested manually on my local environment

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/samples/react/server/server.js
+++ b/samples/react/server/server.js
@@ -119,6 +119,10 @@ export function renderView(callback, path, data, viewBag) {
           `<head>${helmet.title.toString()}${helmet.meta.toString()}${helmet.link.toString()}`
         );
 
+        // replace phkey attribute with key attribute so that newly added renderings
+        // show correct placeholders, so save and refresh won't be needed after adding each rendering
+        html = html.replace(new RegExp('phkey', 'g'), 'key');
+
         callback(null, { html });
       })
       .catch((error) => callback(error, null));


sc-illiakovalenko:
@vsirakova Thank you very much, good catch. 👍 
I tested, it's working as expected.

aweber1:
I would suggest noting this issue and resolution in documentation somewhere because the fix is at the app level and not the React package level. Therefore, anyone that is _not_ using the JSS sample app as a basis will still encounter this issue.

sc-illiakovalenko:
@aweber1 I will mention this issue.
Can I add section "Troubleshooting" in
https://jss.sitecore.com/docs/client-frameworks/react/sample-app
at the end? Or is it enough to add it to release notes?

PR #346, opened 2020-03-16
Responsive image srcset fallback with src output

unterweltler:
## Description
Adding `srcSet` to \<Image> Component removes the `src` Attribute.
This PR will always output the `src` Attribute, even when a `srcSet` Object is passed.

This will improve the support for older Browsers, since they might don't support `srcSet` (like IE11). Using the Progressive Enhancement Approach newer Browsers can benefit from the `srcSet` and will ignore the `src` Attribute.

## Motivation
Improve Support for older Browsers without the need for a `srcSet` Polyfill - which is difficult to implement when using Jss <Link> or Router Links.

## How Has This Been Tested?
Tests for Image updated to test for `srcSet` / `sizes` Attributes output.
I used the same values as in the Styleguide Image Example.
Running `npm run test` inside `root` or `/packages/sitecore-jss-react` will run the new test

## Types of changes

## Checklist:

### DIFF
--- a/packages/sitecore-jss-react/src/components/Image.test.tsx
+++ b/packages/sitecore-jss-react/src/components/Image.test.tsx
@@ -42,6 +42,35 @@ describe('<Image />', () => {
     });
   });
 
+  describe('with responsive image object', () => {
+    const props = {
+      media: {
+        src: '/assets/img/test0.png',
+      },
+      srcSet: [{ mw: 100 }, { mw: 300 }],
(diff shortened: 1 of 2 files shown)


sc-illiakovalenko:
@unterweltler Thank you for contributing! We can add it.

sc-illiakovalenko:
@unterweltler Thank you for contributing! It will solve problem in IE11.

petervangennip:
This would be a nice addition for the jss-vue packaged too.

PR #357, opened 2020-04-14
Updated context extension docs around caching

boro2g:
Added suggested documentation tweak around how you need to be careful of the performance of Context Extensions as OTB cache options aren't available.


sc-illiakovalenko:
@boro2g Thank you for contribution.

PR #368, opened 2020-05-05
Allow using SitecoreContext with a hook

jantimon:
## Description

On February 16, 2019, React `16.8` was released by facebook and added `hooks` to make functional components more powerfull.

This pull request allows using the `SitecoreContext` with way less effort than the current High Order Component approach `withSitecoreContext`:

```ts
const EditMode = () => {
   const {pageEditing} = useSitecoreContext();
   return <span>Edit Mode is {pageEditing ? 'active' : 'inactive'}</span>
}
```

## How Has This Been Tested?

This feature has **not** been tested.   
If you think this would be a good addition I will go on with testing.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sitecore-jss-react/src/enhancers/withSitecoreContext.tsx
+++ b/packages/sitecore-jss-react/src/enhancers/withSitecoreContext.tsx
@@ -32,3 +32,22 @@ export function withSitecoreContext(options?: WithSitecoreContextOptions) {
     };
   };
 }
+
+/**
+ * This hook grants acess to the current SiteCore page context
+ * by default JSS includes the following properties in this context:
+ * - pageEditing - Provided by Layout Service, a boolean indicating whether the route is being accessed via the Experience Editor.
+ * - pageState - Like pageEditing, but a string: normal, preview or edit.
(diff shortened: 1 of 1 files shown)


kamsar:
Presumably adding hooks would require dropping support for React < 16.8. IMO that is worth doing but it's @AdamNaj who would make that decision. It could also be done in a separate hooks package perhaps to avoid that necessity.

Hooks make React really, really nice and IMO it's time JSS got them for context and also the placeholder transforms.

jantimon:
Cool thanks for your feedback 👍

Actually this change is not breaking backwards compatibility.
`useSitecoreContext` would only cause an error `React.useContext is not a function` for old versions once it is called.  

As long as this new hook is not being used, everything will work just as it did before.  

But you are right a next step might be to use it internally and then it would definitely require react 16.8.

sc-illiakovalenko:
@jantimon Thank you, I merge it in future JSS 16 release

sc-illiakovalenko:
@jantimon Added new useSitecoreContext hook in #508

jantimon:
Thanks @sc-illiakovalenko 👍

PR #375, opened 2020-05-19
[Angular] Add directive for rich texts which allows routerlinks that can be both internal and external

jlast:
Add directive for rich texts which allows routerlinks that can be both internal and external.
This allows Content Editors to add internal links to other pages, while still being able to maintain external links.

Recreated from https://github.com/Sitecore/jss/pull/219 after repository maintenance.

Also add a small change so linkfield classes will be merged instead of replaced. 

## Description
Add directive for rich texts which allows routerlinks that can be both internal and external.
This allows Content Editors to add internal links to other pages, while still being able to maintain external links.

## Motivation

Currently the implementation of a rich text is limited. All links won't use angular routerlinks.. This directive decides whether the link has a protocol and is therefor an external link.

## How Has This Been Tested?
Added unit tests and tested it on our application.

## Types of changes

## Checklist:


sc-illiakovalenko:
@jlast Can you remove changes related to code formatting? Seems that some formatter is installed in your workspace.
It's hard to analyse your changes.
For example, check all places:
image

jlast:
> @jlast Can you remove changes related to code formatting? Seems that some formatter is installed in your workspace.
> It's hard to analyse your changes.
> For example, check all places:
> image

Thanks for the reply, updated the merge request

sc-illiakovalenko:
@jlast I want to ask you, switch target branch to: release/15.0.0. And pull it

packages/sitecore-jss-angular/src/components/generic-rich-text.directive.ts
```diff
@@ -0,0 +1,73 @@
...
+  private updateView() {
+    const field = this.field;
+    if (!field || (!field.editable && !field.value)) {
+      return;
+    }
+
+    const html = field.editable && this.editable ? field.editable : field.value;
+    this.viewRef.rootNodes.forEach((node) => {
+      node.innerHTML = html;
+      if(node.querySelectorAll != null) {
```
sc-illiakovalenko, line 56: @jlast Will be more beautiful:
```
if(!node.querySelectorAll) {
     return
}

.... // rest code
```

packages/sitecore-jss-angular/src/components/generic-rich-text.directive.ts
```diff
@@ -0,0 +1,73 @@
...
+        const links: NodeListOf<HTMLLinkElement> = node.querySelectorAll('a[href]');
+        const linksArray: Array<HTMLLinkElement> = [].slice.call(links);
+
+        linksArray.forEach((link) => {
+          const href = link.getAttribute('href');
+
+          if (href != null && !isAbsoluteUrl(href)) {
```
sc-illiakovalenko, line 63: @jlast Can be better if:
```
if (!href || isAbsoluteUrl(href)) {
   return;
}

... // rest code
```

packages/sitecore-jss-angular/src/components/link.directive.ts
```diff
@@ -81,6 +81,16 @@ export class LinkDirective implements OnChanges {
     this.inlineRef = span;
   }
 
+  protected updateAttribute(node: any, key: string, prop: any) {
+    if (prop != null && prop !== '') {
```
sc-illiakovalenko, line 85: @jlast Can be better if:
```
if (!prop || prop === '') {
   return;
}

if (key === 'class' && node.className !== '') {
    this.renderer.setAttribute(node, key, `${node.className} ${prop}`);
} else {
    this.renderer.setAttribute(node, key, prop);
}
```
More hard to read nested code.

sc-illiakovalenko:
@jlast I'm not sure that understand correctly, but why can't we extend `rich-text.directive.ts` functionality in `updateView` function? Instead of it we have two directives for RichText, and they have difference only in terms of handling of `routerlinks`.
For example we have `GenericLinkDirective` and it extends `LinkDirective`, but now we have `GenericRichTextDirective` and it's not extending default `RichTextDirective`. What is the difference between Generic and default directives? I'm not in a context a little bit

jlast:
> @jlast I'm not sure that understand correctly, but why can't we extend `rich-text.directive.ts` functionality in `updateView` function? Instead of it we have two directives for RichText, and they have difference only in terms of handling of `routerlinks`.
> For example we have `GenericLinkDirective` and it extends `LinkDirective`, but now we have `GenericRichTextDirective` and it's not extending default `RichTextDirective`. Maybe `GenericRichTextDirective` should extend `RichTextDirective`?
> 
> So scGenericLink - handle both internal and external links, for internal links - it is using Angular routerlinks, for external -
> scLink - handle both internal and external links, for both types it's using default ? Am I right?
> 
> So you want to have same behavior but in rich-text field? Because for now RichTextDirective can't render `routerlinks`

jlast:
I'll update this merge request probably in 1-2 weeks, working on other priorities for our client at the moment.

jpspringall:
This will be awesome when it goes in as I was thinking of doing something similiar

jlast:
> @jlast I'm not sure that understand correctly, but why can't we extend `rich-text.directive.ts` functionality in `updateView` function? Instead of it we have two directives for RichText, and they have difference only in terms of handling of `routerlinks`.
> For example we have `GenericLinkDirective` and it extends `LinkDirective`, but now we have `GenericRichTextDirective` and it's not extending default `RichTextDirective`. Maybe `GenericRichTextDirective` should extend `RichTextDirective`?
> 
> So scGenericLink - handle both internal and external links, for internal links - it is using Angular routerlinks, for external -
> scLink - handle both internal and external links, for both types it's using default ? Am I right?
> 
> So you want to have same behavior but in rich-text field? Because for now RichTextDirective can't render `routerlinks`

@sc-illiakovalenko 
To be honest I was not sure if this is what all consumers expect and use. If you feel like this can replace the current RichText implementation I would gladly replace it. Same can be said for the generic-link directive.

Please advice me how to move forward

illiakovalenko:
@jlast Related PR is merged, I close this one

PR #384, opened 2020-06-11
Changed component factory to match only first occurence of component

MickJar:
## Motivation
The problem it solves is that currently generate component factory does not support components that extend or implement other components.

Thus it now supports the following naming conventions:

1. export class AComponent 
2. export class AComponent extends BComponent
3. export class AComponent implements CComponent

## How Has This Been Tested?
This has been tested locally by running jss start , with an existing JSS solution.

## Types of changes

## Checklist:

- [ x] I have read the Contributing guide.
- [x ] My code/comments/docs fully adhere to the Code of Conduct.
- [ x] My change is a code change and it requires an update to the documentation.

### DIFF
--- a/samples/angular/scripts/generate-component-factory.ts
+++ b/samples/angular/scripts/generate-component-factory.ts
@@ -76,7 +76,7 @@ function generateComponentFactory() {
 
     // ASSUMPTION: your component should export a class directly that follows Angular conventions,
     // i.e. `export class FooComponent` - so we can detect the component's name for auto registration.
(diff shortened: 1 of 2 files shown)


sc-illiakovalenko:
@MickJar It will not work, if I'll test it using:
```
export class StyleguideComponentParamsComponent implements OnInit
```

sc-illiakovalenko:
@MickJar It will work with following regexp: `export class (.+?)Component\b`

sc-illiakovalenko:
@MickJar Merged in #391

MickJar:
Thanks!

PR #391, opened 2020-06-30
[Angular sample app] Component factory match first occurrence of Component name.

sc-illiakovalenko:


sc-illiakovalenko:
Merged with #384

Issue #380, opened 2020-05-27
Prevent the Angular scLink directive from adding an empty href attribute


callmeaponte:
## Is your feature request related to a problem? Please describe.
The `*scLink` directive adds an href attribute, even if the link data (from the JSS JSON) is empty:

```
{
    "fields": {
        "HeadingText": {
            value: "Hello World"
        },

        "HeadingLink": {
            "value": {
                "href": "",
                "text": "",
                "linktype": "",
                "url": "",
                "anchor": "",
                "title": "",
                "target": "",
            }
        }
    }
}
```

```
<!-- input -->
<a *scLink="HeadingLink">
    <h3>{{ HeadingText.value }}</h3>
</a>
```

```
<!-- expected output -->
<a>
    <h3>Hello World</h3>
</a>

<!-- actual output -->
<a href>
    <h3>Hello World</h3>
</a>
```

This causes a few issues:

1) Links with an empty href attribute still display a pointer icon on hover, indicating to users that it is clickable - when it is not.
2) We have to write a bunch of boilerplate code (see below) to accommodate situations where a content author may, or may not, want to make an element linkable in Sitecore.
3) In terms of accessibility, having an empty href is worse than omitting it.

To elaborate on point #2 (boilerplate code), the sample JSS JSON above represents a Sitecore content item that lets users specify a heading, and also to (optionally) wrap that heading in a link if they wanted to. Currently, the markup we need to write to support this on the front-end is as follows:

```
<ng-template [ngIf]="HeadingLink.value.href">
    <a *scLink="HeadingLink">
        <h3>{{ HeadingText.value }}</h3>
    </a>
</ng-template>

<ng-template [ngIf]="!HeadingLink.value.href">
    <h3>{{ HeadingText.value }}</h3>
</ng-template>
```

... In contrast, it would be nice to be able to simply write:

```
<a *scLink="HeadingLink">
    <h3>{{ HeadingText.value }}</h3>
</a>
```

## Describe the solution you'd like
It would be preferable to omit the attribute instead:

```
<!-- input -->
<a *scLink="someData"></a>

<!-- output -->
<a></a>
```

## Describe alternatives you've considered
The alternative is to use the boilerplate code displayed above. However this is not ideal, especially when you have to repeat that code for dozens and dozens of link-able elements across your code base.

## Additional information
If you approve of this change in behavior then I can open up a PR, just let me know.

sc-illiakovalenko:
[...]

PR #406, opened 2020-07-14
[sitecore-jss-angular] Prevent the Angular scLink directive from adding an empty href attribute

sc-illiakovalenko:


sc-illiakovalenko:
Related to #380

Issue #408, opened 2020-07-15
Prevent the Angular scLink directive from adding an href attribute, when href is equal to `http://` or `https://`


sc-illiakovalenko:
Related to #380. Need to extend #406 in order to prevent adding of `href` attribute when `href` is equal to `http://` or `https://`

sc-illiakovalenko:
Fixed in #409 
@callmeaponte As I promised 😉

callmeaponte:
@sc-illiakovalenko thank you! It's much appreciated.

PR #409, opened 2020-07-16
Prevent the Angular scLink directive from adding an href attribute, when href is equal to `http://` or `https://`

sc-illiakovalenko:


sc-illiakovalenko:
Related to #408

PR #421, opened 2020-07-23
Section field type - cssClass Property

bic742:
Section field type needs to inherit from FieldViewModel, so it has access to read the cssClass property on the field rendering.

## Description
I changed the inherited type for `FormFieldSection` to be of type `FormField<FieldViewModel` to expose the `cssClass` property on the model: `field.model.cssClass`

## Motivation
We can get to the property using bracket notation (`this.field.model['cssClass']`). Setting the Css Class for a section is an available option in the Sitecore Forms Builder, and other properties are able to access it via the strong typing: `field.model.cssClass`. It makes sense to be able to do this with a Section, too.

## Types of changes

## Checklist:


sc-illiakovalenko:
@bic742 It's already fixed. https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss-forms/src/FormField.ts#L32
Pull changes from dev

PR #447, opened 2020-09-12
[JSS 13.0.5][React sample] Race condition, under load React app may render HTML from a different route 

sc-illiakovalenko:


samples/react/server/server.js
```diff
@@ -58,15 +56,21 @@ export function renderView(callback, path, data, viewBag) {
...
+        const renderedAppHtml = renderToStringWithData(
+          <AppRoot path={path}
+            Router={StaticRouter}
+            graphQLClient={graphQLClient}
+            ssrState={state}
+          />
         )
-      )
+
+        return renderedAppHtml;
```
nickwesselman, line 72: We could get rid of the `{ }` and `return` statement in this part of the promise chain now that this clause has been simplified.

samples/react/src/AppRoot.js
```diff
@@ -22,21 +21,53 @@ export const routePatterns = [
...
+      ssrRenderComplete
+    })
+  }
+
+  render() {
+    const { path, Router, graphQLClient, ssrState } = this.props;
+
+    if (ssrState && ssrState.sitecore && ssrState.sitecore.route) {
+      // set the initial sitecore context data if we got SSR initial state
+      this.state.contextFactory.setSitecoreContext({
```
jacobrask, line 41: The `render` method should not have side effects, in any applications using `React.StrictMode` the `render` method will even be run twice in development mode to dissuade you from side effects in `render`.

PR #459, opened 2020-09-30
Fix prop-types for missingComponentComponent to allow a functional component

vsirakova:
## Description
* Add a unit test in `Placeholder.test.tsx` that checks the case when an unknown component is requested for rendering -> then the passed `missingComponentComponent` prop of the `Placeholder` should be used.
* Fix prop-types for `missingComponentComponent` prop as it currently allows only objects while it should allow a function

## Motivation
When a custom `MissingComponent` implementation is passed to the `missingComponentComponent` prop there is a runtime error saying `Failed prop type: Invalid prop missingComponentComponent supplied to PlaceholderComponent.`
This is because the prop-types are incorrect and don't allow a function.

## How Has This Been Tested?
I have added a test case that is using the missingComponentComponent prop and when this test is ran with the old prop-types then the runtime error is reproduced.
When prop-types allow for a `func` the runtime error disappears.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sitecore-jss-react/src/components/MissingComponent.tsx
+++ b/packages/sitecore-jss-react/src/components/MissingComponent.tsx
@@ -7,7 +7,7 @@ export interface MissingComponentProps {
   };
 }
 
-export const MissingComponent: React.SFC<MissingComponentProps> = (props) => {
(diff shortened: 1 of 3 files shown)


sc-illiakovalenko (approved):
@vsirakova Thank you for contribution!

PR #460, opened 2020-09-30
Fix return type of ComponentFactory to allow a component definition instead of an instantiated component

vsirakova:
Fix the typing for `ComponentFactory` to support a component type instead of an instantiated component

## Description
Change the return type of the `ComponentFactory` function type to `ComponentType` instead of just `Component` which assumes an instantiated component

## Motivation
This incorrect type is causing typing issues when `componentFactory` prop is passed to `SitecoreContext`. The test that is changed in this PR will result into an error if ran with the old typing. The error would be something like this `Type '(componentName: string) => React.FC<{}>' is not assignable to type 'ComponentFactory'.`

## How Has This Been Tested?
The incorrect type is causing an error in the Placeholder.test.tsx file while with the updated type the error disappears.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sitecore-jss-react/src/components/Placeholder.test.tsx
+++ b/packages/sitecore-jss-react/src/components/Placeholder.test.tsx
@@ -12,10 +12,10 @@ import {
 import { convertedData as eeData } from '../testData/ee-data';
 
 const componentFactory: ComponentFactory = (componentName: string) => {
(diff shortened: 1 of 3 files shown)


sc-illiakovalenko (approved):
@vsirakova Thank you, will add release notes about it

PR #476, opened 2020-10-17
UPDATED | The debug message whether or not a url was ignored was inve…

jpspringall:
The debug message whether or not a url was ignored was inverted

## Description
In the function isUrlIgnored, the message round whether or not the url was ignored was the wrong way round.
It was saying it would be a straight pass through when it wasn't and vice versa

I note it has also updated the formatting on the file.
I note there is a .prettierrc file, as I user prettier, I'm assuming that this formatting changes are valid 

## Motivation
Just to help with debugging

## How Has This Been Tested?
Tested with my own application 
No unit tests as function not exported

## Types of changes

## Checklist:


packages/sitecore-jss-proxy/src/index.ts
```diff
@@ -388,12 +391,12 @@ function isUrlIgnored(originalUrl: string, config: ProxyConfig, noDebug: boolean
     if (!noDebug && config.debug) {
       if (!result) {
```
sc-illiakovalenko, line 392: @jpspringall 
DEFINITION OF pathRewriteExcludeRoutes:
pathRewriteExcludeRoutes: A list of absolute paths that are NOT app routes and should not attempt to render a route using SSR. These route prefixes are directly proxied to the apiHost, allowing the proxy to also proxy GraphQL requests, REST requests, etc. Local static assets, Sitecore API paths, Sitecore asset paths, etc should be listed here. URLs will be encoded, so e.g. for a space use '%20' in the exclude.

In case if - result is False: originalUrl DIDN'T match the proxy exclude list (because we didn't find part of this url in config.pathRewriteExcludeRoutes), and will be treated as layout service to render.
if result is True: originalUrl MATCH the proxy exclude list, and will be served verbatim as received.

It matches to definition of pathRewriteExcludeRoutes.
Can you check again? Seems like everything were defined correctly in console.log

sc-illiakovalenko (changes requested):

packages/sitecore-jss-proxy/src/index.ts
jpspringall: Hi @sc-illiakovalenko This PR can be closed, you fixed the issue here: https://github.com/Sitecore/jss/commit/d68490b52ad3a4b4a6ef0fe0f4a7f1342d5b9b19#diff-ededfb74a5a0276f0c31fe53cf2f41b07474e4db17b92d4cf5243a1d6c49f323

Issue #420, opened 2020-07-23
jss deploy doesn't support custom config file


thany:
## Description
We need a way to tell `jss deploy` which config file to pick up. This is for the purpose of deploying to different environments, e.g. content delivery, content management, and local SSR. We deploy to these environments in parallel, so in order to support `jss deploy` as it is, we *have to* deploy in serial, because the current `jss deploy` [hardcoded requires a static file called `scjssconfig.json` in the project root](https://github.com/Sitecore/jss/blob/938305ee66fc9937b18a6de9e6b68c3f6feda18b/packages/sitecore-jss-dev-tools/src/resolve-scjssconfig.ts#L19). We can prepare one for each environment, but then we couldn't ever deploy in parallel.

It also makes sense for other purposes, like when you want config files to reside in their own directory.

## Expected behavior
Add a CLI argument like `--config` and the path to the config file.

## Steps To Reproduce
(Not applicable)

## Possible Fix
See *expected behaviour*

## Your Environment
* Sitecore Version: 9.2
* JSS Version: 12.0.0
* Browser Name and version: Not applicable
* Operating System and version (desktop or mobile): Windows 10 (desktop)
* Link to your project (if available): Not available (private repo)

sc-illiakovalenko:
@thany Implemented in #485

PR #485, opened 2020-11-05
[JSS CLI] Add --config key in order to specify path to scjssconfig.json file

sc-illiakovalenko:
[JSS CLI] Add --config key in order to specify path to scjssconfig.json file

Related to #420 

## Types of changes

## Checklist:

### DIFF
--- a/packages/sitecore-jss-dev-tools/src/resolve-scjssconfig.ts
+++ b/packages/sitecore-jss-dev-tools/src/resolve-scjssconfig.ts
@@ -14,9 +14,13 @@ export interface ScJssConfig {
   [configName: string]: JssConfiguration;
 }
 
-export function resolveScJssConfig(configName = 'sitecore', assert = true): Promise<ScJssConfig> {
+export function resolveScJssConfig({ 
+  configPath = './scjssconfig.json',
+  configName = 'sitecore',
+  assert = true
+} = {}): Promise<ScJssConfig> {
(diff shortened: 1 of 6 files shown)


sc-illiakovalenko:
@anastasiya29 @nickwesselman Can you review, please? For me seems like here is a breaking change. Maybe somebody can use this `resolveScJssConfig` function. Am I right? https://github.com/Sitecore/jss/pull/485/files#diff-2e90590e1376e0af69c5d8cacf91a2806e27dab16ab19bf7eacac71b79a3e001R17

packages/sitecore-jss-dev-tools/src/resolve-scjssconfig.ts
```diff
@@ -14,9 +14,13 @@ export interface ScJssConfig {
   [configName: string]: JssConfiguration;
 }
 
-export function resolveScJssConfig(configName = 'sitecore', assert = true): Promise<ScJssConfig> {
+export function resolveScJssConfig({
```
ambrauer, line 17: Yeah, this would be a breaking change - but I wonder how many folks are using this function directly? We would be able to provide backwards compatibility by leaving the signature and adding the `configPath` parameter at the end, but this makes things overly verbose (having to specify each parameter in our internal calls). Switching to a named/object parameter certainly seems like the better direction long-term.

ambrauer (approved):

PR #493, opened 2020-11-13
[NextJs] i18n updates

ambrauer:
Updates after further i18n analysis / testing, including:

* Consolidate default language/locale settings (always defer to package.json config language)
* Allow sample to run with Next.js i18n disabled
* Add locale variations of configured rewrites (for Sitecore requests pass-through e.g. /sitecore, /media, /layouts)
* Upgraded next-localization (w/ rosetta) library as dictionary solution of choice. Simplified usage examples. Updated comments in _app.tsx.
* Remove export scripts for now (since it doesn't work with next.js v10 i18n and image optimization), to be addressed in separate task

Known limitations / open issues:

* Root page client-side route for 'da-DK' shows 404
  * Repro: go to http://localhost:3000/da-DK/styleguide, click on logo > get 404
  * Direct request for http://localhost:3000/da-DK works fine (refresh page)
  * RESOLUTION: Known issue with fix in Next.js, **confirmed fixed with latest canary build**
* i18n support is not compatible with next export
* Cannot force sub-path/prefix usage for default locale
  * Currently breaking test for multilingual styleguide component (English href is "/styleguide" instead of expected "/en/styleguide")
  * Would be preferred with "always" languageEmbedding in Sitecore
  * This doesn't mean it won't resolve the route, just that next/link does not add it
  * RESOLUTION:
    * Fix/enhancement is in the works in Next.js
    * In the meantime, workaround would be to add the prefix yourself in all client routing. e.g.
    ```
    <Link href={`${router.locale}/styleguide`} locale={false}>
      <a className="p-2 text-dark">{i18n.t('Styleguide')}</a>
    </Link>
    ```


sc-illiakovalenko (commented):
@ambrauer For me, everything is fine. Will we store somewhere these limitations/issues?

ambrauer:
@sc-illiakovalenko Yep, these limitations are covered in docs and/or backlog items.

PR #494, opened 2020-11-13
Clean up dependency versions

anastasiya29:
## Motivation
Inconsistent jss package versions prevents local symlinking from working properly.

## Description
This updates packages/nextjs to 15.0.1, and sets the dependency from samples/nextjs to * so that we don't have to update it while this is in development.

## How Has This Been Tested?
1. Deleted node folder and lock file from both nextjs packages
2. Ran reset command from repo root and verified that symlinks for all jss dependencies in the nextjs sample (including the unpublished nextjs core package) are getting added
3. Started nextjs sample in disconnected mode 

## Types of changes
Non-breaking fix for new feature
No docs required

### DIFF
--- a/samples/nextjs/package.json
+++ b/samples/nextjs/package.json
@@ -31,7 +31,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@sitecore-jss/sitecore-jss-nextjs": "*",
+    "@sitecore-jss/sitecore-jss-nextjs": "^15.0.1",
     "@sitecore-jss/sitecore-jss-tracking": "^15.0.1",
     "apollo-cache-inmemory": "~1.6.3",
     "apollo-client": "~2.6.4",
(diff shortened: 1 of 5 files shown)


samples/nextjs/package.json
```diff
@@ -31,8 +31,8 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@sitecore-jss/sitecore-jss-nextjs": "^15.0.0",
-    "@sitecore-jss/sitecore-jss-tracking": "^15.0.0",
+    "@sitecore-jss/sitecore-jss-nextjs": "*",
```
ambrauer, line 34: I think we should just keep this in sync with the rest ("^15.0.1"). Now that the nextjs package and sample are merged into dev, they should be kept in sync when we bump version. Also, we don't have to remember to remove the asterisk when we release :)

ambrauer, line 34: I think we should just keep this in sync with the rest ("^15.0.1"). Now that the nextjs package and sample are merged into dev, they should be kept in sync when we bump version. Also, we don't have to remember to remove the asterisk when we release :)

ambrauer (changes requested):

PR #496, opened 2020-11-18
Fix issue with VisitorIdentification control on Next.js

ambrauer:
- Fixes an issue with `VisitorIdentification` React control on Next.js by moving everything (script and meta tag addition) to purely client-side 
- A pageEditing check has also been added within the control which avoids having to check in consuming app

### DIFF
--- a/packages/sitecore-jss-react/src/components/VisitorIdentification.tsx
+++ b/packages/sitecore-jss-react/src/components/VisitorIdentification.tsx
@@ -3,7 +3,6 @@ import { withSitecoreContext } from '../enhancers/withSitecoreContext';
 
 interface VIProps {
   sitecoreContext: {
-    pageEditing?: boolean;
     visitorIdentificationTimestamp?: string;
   };
 }
@@ -13,12 +12,10 @@ const VIComponent: React.FC<VIProps> = ({ sitecoreContext }) => {
   if (
     emittedVI ||
     typeof document === 'undefined' ||
-    sitecoreContext.pageEditing ||
     !sitecoreContext.visitorIdentificationTimestamp
   ) {
     // Don't emit VI script and meta tag if we've already done so,
-    // aren't rendering client-side, are in Experience Editor, or
-    // don't have a VI timestamp.
+    // aren't rendering client-side, or don't have a VI timestamp.
     return null;
   }
   emittedVI = true;
@@ -31,9 +28,9 @@ const VIComponent: React.FC<VIProps> = ({ sitecoreContext }) => {
   meta.name = 'VIcurrentDateTime';
   meta.content = sitecoreContext.visitorIdentificationTimestamp;
 
-  const head = document.getElementsByTagName('head')[0];
-  head.appendChild(script);
-  head.appendChild(meta);
+  const head = document.querySelector('head');
+  head && head.appendChild(script);
+  head && head.appendChild(meta);
 
   return null;
 };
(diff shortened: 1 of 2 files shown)


packages/sitecore-jss-react/src/components/VisitorIdentification.tsx
```diff
@@ -3,27 +3,37 @@ import { withSitecoreContext } from '../enhancers/withSitecoreContext';
...
 let emittedVI = false;
-let VIComponent: React.SFC<VIProps> = ({ sitecoreContext }) => {
+const VIComponent: React.FC<VIProps> = ({ sitecoreContext }) => {
   if (
-    !emittedVI &&
-    typeof document !== 'undefined' &&
-    sitecoreContext.visitorIdentificationTimestamp
+    emittedVI ||
+    typeof document === 'undefined' ||
+    sitecoreContext.pageEditing ||
```
sc-illiakovalenko, line 16: @ambrauer So, in Experience Editor this warning also exists?

sc-illiakovalenko: Nvm, I understood that it shouldn't be invoked in EE

packages/sitecore-jss-react/src/components/VisitorIdentification.tsx
```diff
@@ -3,27 +3,37 @@ import { withSitecoreContext } from '../enhancers/withSitecoreContext';
...
+
+  const script = document.createElement('script');
+  script.src = `/layouts/system/VisitorIdentification.js`;
+  script.type = 'text/javascript';
+
+  const meta = document.createElement('meta');
+  meta.name = 'VIcurrentDateTime';
+  meta.content = sitecoreContext.visitorIdentificationTimestamp;
+
+  const head = document.getElementsByTagName('head')[0];
```
sc-illiakovalenko, line 34: @ambrauer What about `document.querySelector('head')`?

packages/sitecore-jss-react/src/components/VisitorIdentification.tsx
sc-illiakovalenko: Need remember to add in release notes that pageEditing will be checked inside component

sc-illiakovalenko (changes requested):

packages/sitecore-jss-react/src/components/VisitorIdentification.tsx
ambrauer: @sc-illiakovalenko Probably safer, I've switched to use.

packages/sitecore-jss-react/src/components/VisitorIdentification.tsx
ambrauer: @sc-illiakovalenko Actually removed sitecoreContext.pageEditing check after further testing. This is not necessary as sitecoreContext.visitorIdentificationTimestamp is always undefined in editing/preview mode. I believe when I noticed this issue it was during editing, but the main issue was adding the meta tag w/ VI timestamp at render time (which has been resolved).

samples/nextjs/src/components/Layout.tsx
```diff
@@ -72,7 +66,7 @@ const Layout = ({ route, sitecoreContext }: LayoutProps) => {
 
         VI detection only runs once for a given analytics ID, so this is not a recurring operation once cookies are established.
       */}
-      {!sitecoreContext.pageEditing && <VisitorIdentification />}
```
sc-illiakovalenko, line 75: @ambrauer You removed page editing check inside VisitorIdentification component, should you revert changes here?

sc-illiakovalenko (changes requested):

samples/nextjs/src/components/Layout.tsx
ambrauer: @sc-illiakovalenko No, this check was made originally in the POC as a quick way to get past this issue with the VisitorIdentification control. But no longer required now that it's resolved.

sc-illiakovalenko (approved):

PR #499, opened 2020-11-24
[NextJS] Move component-props-service to sitecore-jss-nextjs

sc-illiakovalenko:
## Types of changes

## Checklist:


samples/nextjs/package.json
```diff
@@ -88,6 +89,7 @@
     "yaml-loader": "^0.6.0"
   },
   "scripts": {
+    "prepare": "cd ./node_modules/@sitecore-jss/sitecore-jss-nextjs && npm link ../../react",
```
sc-illiakovalenko, line 92: Called by lerna, when it bootstrap packages/samples. It solves problem of multiple instances of `react` when use `hooks`

sc-illiakovalenko: https://github.com/lerna/lerna/tree/main/commands/bootstrap

packages/sitecore-jss-nextjs/src/services/component-props-service.test.ts
```diff
@@ -0,0 +1,309 @@
...
+  const service = new ComponentPropsService();
+
+  const rendering = (name: string): ComponentRendering => ({
+    uid: name,
+    componentName: name,
+  });
+
+  type CustomContext = { locale: string };
+
+  const placeholders = {
```
ambrauer, line 20: How are we handling the same component/rendering added multiple times (to the same placeholder, or spread among different placeholders)? I think these are valid scenarios - the renderings may differ by `datasource` or `params` - so you'd have multiple renderings with the same componentName, but different uid. Might be good to have a test for this scenario.

packages/sitecore-jss-nextjs/package.json
```diff
@@ -42,12 +41,14 @@
     "enzyme-adapter-react-16": "^1.14.0",
     "jsdom": "^15.1.1",
     "mocha": "^8.1.3",
+    "next": "^10.0.2",
```
ambrauer, line 44: Would you mind updating next version in the sample app as well? Also, 10.0.3 just dropped overnight :)

packages/sitecore-jss-nextjs/src/sharedTypes/component-props.ts
```diff
@@ -0,0 +1,30 @@
+import { GetServerSidePropsContext, GetStaticPropsContext } from "next";
+import {ComponentRendering, LayoutServiceData} from '@sitecore-jss/sitecore-jss';
+
+/**
+ * Shape of component props storage
+ */
+export type ComponentProps = {
+  [componentName: string]: unknown;
```
ambrauer, line 8: Given my comment above about needing to handle multiple components with the same name - this will likely need to change to a uid key.

samples/nextjs/package.json
ambrauer, line 92: Not great that this will be in users apps when they create a nextjs sample. There isn't a way for us to do this in the repo root package.json scripts (even if it is ugly)?

samples/nextjs/src/lib/page-props.ts
```diff
@@ -1,6 +1,9 @@
-import { DictionaryPhrases, LayoutServiceData } from '@sitecore-jss/sitecore-jss-nextjs';
+import {
+  DictionaryPhrases,
+  LayoutServiceData,
+  ComponentProps,
```
ambrauer, line 4: Should we name this type something more specific to its dictionary/collection purpose, like "ComponentPropsCollection"? I could see it being confusing (vs those types in my "component-props.ts"). 

The property name "componentProps" on the SitecorePageProps type is fine I think.

packages/sitecore-jss-nextjs/src/components/ComponentPropsContext.tsx
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Component props context which we are using in order to store data fetched on components level (getStaticProps/getServerSideProps)
+ */
+export const ComponentPropsReactContext = createContext<ComponentProps>({});
+
+/**
+ * Hook in order to get access to props related to specific component. Data comes from ComponentPropsContext.
+ * @see ComponentPropsContext
+ */
+export function useComponentProps<ComponentData>(componentName: string): ComponentData {
```
ambrauer, line 13: Given my comment below about needing to handle multiple components with the same name - this will likely need to change to a componentUid parameter / key.

packages/sitecore-jss-nextjs/src/services/component-props-service.ts
```diff
@@ -0,0 +1,176 @@
...
+  }
+
+  /**
+   * Execute request for component props
+   */
+  async execRequests<NextContext>(requests: ComponentPropsRequest<NextContext>[]) {
+    const componentProps: ComponentProps = {};
+
+    const promises = requests.map((req) =>
+      req.fetch(req.rendering, req.layoutData, req.context).then((result) => {
```
ambrauer, line 142: Should we add error handling for the individual fetch functions? It's my understanding that Promise.all would reject as soon as one promise rejects. So, if one of your components errors, the entire page would fail?

ambrauer (changes requested):

packages/sitecore-jss-nextjs/src/index.ts
```diff
@@ -22,6 +22,27 @@ export {
   ComponentParams,
 } from '@sitecore-jss/sitecore-jss';
 
+export {
+	ComponentProps,
+	GetStaticComponentProps,
+	GetServerSideComponentProps,
+	ComponentPropsFetchFunction,
```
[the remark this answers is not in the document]
sc-illiakovalenko: May be yes, I was thinking about export `FetchComponentPropsArguments` - but I'm not sure that somebody will use it

packages/sitecore-jss-nextjs/package.json
sc-illiakovalenko: Changed

samples/nextjs/package.json
sc-illiakovalenko: Changed

packages/sitecore-jss-nextjs/src/services/component-props-service.test.ts
sc-illiakovalenko: Now I use uid in order to store data, instead of componentName.

sc-illiakovalenko: this `placeholders` - contains MyCustomComponent - several times, and in different placeholders

packages/sitecore-jss-nextjs/src/sharedTypes/component-props.ts
sc-illiakovalenko: Changed

samples/nextjs/src/lib/page-props.ts
sc-illiakovalenko: Agree, it looks more clear, when I use Collection word

packages/sitecore-jss-nextjs/src/components/ComponentPropsContext.tsx
sc-illiakovalenko: Changed

packages/sitecore-jss-nextjs/src/services/component-props-service.ts
sc-illiakovalenko: Added `catch` handler, log error and write it inside component props.

ambrauer (approved):

PR #501, opened 2020-11-25
Add linting steps to CI pipeline

ambrauer:
Note it's **only linting Next.js package and sample for now**. The remaining packages/samples will be addressed and added in future tasks.

Also fixes all package lint scripts to actually return errors.


package.json
```diff
@@ -2,13 +2,13 @@
   "scripts": {
     "reset": "lerna clean --yes && npm run bootstrap && npm run build-packages",
     "bootstrap": "lerna bootstrap && npm run bootstrap-nextjs-sample",
+    "bootstrap:ci": "lerna bootstrap --ci",
```
sc-illiakovalenko, line 5: @ambrauer Do we need to bootstrap packages+samples? Previously we only bootstrapped packages using: `bootstrap-packages:ci`

ambrauer: @sc-illiakovalenko Yeah, the samples need to be bootstrapped as well in order to lint them.

sc-illiakovalenko (approved):

PR #511, opened 2020-12-11
[NextJS] Move LayoutService to sitecore-jss package

sc-illiakovalenko:
## Types of changes

## Checklist:


packages/sitecore-jss/src/layout-service.ts
```diff
@@ -0,0 +1,182 @@
...
+  siteName: string;
+  /**
+   * Enables/disables analytics tracking for the Layout Service invocation (default is true).
+   * More than likely, this would be set to false for SSG/hybrid implementations, and the
+   * JSS tracker would instead be used on the client-side: {@link https://jss.sitecore.com/docs/fundamentals/services/tracking}
+   */
+  tracking?: boolean;
+}
+
+type FetchConfig = {
```
ambrauer, line 20: Reading through everything, I think this `FetchConfig` has conflicting interests, and may not even be necessary. Seems like the properties here could be split up, moved to other locations more suited to their usage:

- `dataFetcher` - move to `LayoutServiceConfig`. There's no reason to be specified/created for each request
- `onReq`, `onReqError`, `onRes`, `onResError` - move to `AxiosDataFetcher` constructor. These are only used here anyway.
- `req`, `res` - move to `fetchLayoutData` as optional parameters (in place of current `FetchConfig`)

ambrauer (changes requested):

packages/sitecore-jss/src/layout-service.ts
```diff
@@ -0,0 +1,182 @@
+import { AxiosRequestConfig, AxiosResponse } from 'axios';
+import { AxiosDataFetcher } from './data-fetcher';
+import { LayoutServiceData } from './dataModels';
+import * as dataApi from './dataApi';
+import { HttpJsonFetcher } from './httpClientInterface';
+import { IncomingMessage, ServerResponse } from 'http';
+
+interface LayoutServiceConfig {
```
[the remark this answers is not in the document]
sc-illiakovalenko: Done

packages/sitecore-jss/src/layout-service.ts
```diff
@@ -0,0 +1,182 @@
...
+}
+
+interface FetchOptions {
+  layoutServiceConfig: {
+    host: string;
+  };
+  querystringParams: FetchParams;
+}
+
+export class LayoutService {
```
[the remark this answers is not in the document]
sc-illiakovalenko: Added '@default' above `tracking?: bolean`

packages/sitecore-jss/src/layout-service.ts
sc-illiakovalenko: Good idea, changed

packages/sitecore-jss/src/layout-service.ts
```diff
@@ -0,0 +1,100 @@
...
+  tracking?: boolean;
+  /**
+   * Custom data fetcher. @see HttpJsonFetcher<T> type for implementation details/notes.
+   * By default used @see AxiosDataFetcher
+   * @param {string} url The URL to request; may include query string
+   * @param {any} [data] Data to POST with the request.
+   * @param {IncomingMessage} [req] Request instance
+   * @param {ServerResponse} [res] Response instance
+   */
+  dataFetcher?: HttpJsonFetcher<LayoutServiceData>;
```
ambrauer, line 26: The `HttpJsonFetcher` doesn't accept the `req`/`res` parameters directly, but I think it is important we support a custom dataFetcher that accepts them. Probably need to switch this up to more of a factory pattern (like the `getDefaultFetcher` below) - which accepts `req`/`res` and returns `HttpJsonFetcher`.

packages/sitecore-jss/src/layout-service.ts
```diff
@@ -0,0 +1,100 @@
...
+   */
+  fetchLayoutData(
+    itemPath: string,
+    language?: string,
+    req?: IncomingMessage,
+    res?: ServerResponse
+  ): Promise<LayoutServiceData> {
+    const fetchOptions = this.getFetchOptions(language);
+
+    const fetcher = this.serviceConfig.dataFetcher || this.getDefaultFetcher(req, res);
```
ambrauer, line 96: Pass along req/res to custom dataFetcher here (see comments above).

ambrauer (changes requested):

packages/sitecore-jss/src/layout-service.ts
```diff
@@ -1,12 +1,12 @@
 import { AxiosDataFetcher } from './data-fetcher';
-import { LayoutServiceData } from './dataModels';
-import * as dataApi from './dataApi';
+import { LayoutServiceData, PlaceholderData } from './dataModels';
+import { fetchPlaceholderData, fetchRouteData, LayoutServiceConfig } from './dataApi';
 import { HttpJsonFetcher } from './httpClientInterface';
 import { IncomingMessage, ServerResponse } from 'http';
 
-export type DataFetcherResolver = (req?: IncomingMessage, res?: ServerResponse) => HttpJsonFetcher<LayoutServiceData>;
+export type DataFetcherResolver<T> = (req?: IncomingMessage, res?: ServerResponse) => HttpJsonFetcher<T>;
```
ambrauer, line 7: Think we could make an exception and use `any` here to avoid having 2 DataFetcherResolvers on the `LayoutServiceInstanceConfig`.
```suggestion
export type DataFetcherResolver = (req?: IncomingMessage, res?: ServerResponse) => HttpJsonFetcher<any>;
```

ambrauer (changes requested):

packages/sitecore-jss/src/layout-service.ts
```diff
@@ -0,0 +1,135 @@
...
+   * a specific route item. Allows you to retrieve rendered data for individual placeholders instead of entire routes.
+   * @param {string} placeholderName
+   * @param {string} itemPath
+   * @param {IncomingMessage} [req] Request instance
+   * @param {ServerResponse} [res] Response instance
+   */
+  fetchPlaceholderData(
+    placeholderName: string,
+    itemPath: string,
+    req?: IncomingMessage,
```
ambrauer, line 124: We should add `language?: string,` as 3rd parameter and pass it to `getFetchOptions()`.

sc-illiakovalenko: Good call, will add

ambrauer (approved):

PR #516, opened 2020-12-17
[NextJs] Move SitecorePageProps creation out of page routes

ambrauer:
## Description
Moves the work of SitecorePageProps creation out of page routes into separate SitecorePagePropsFactory.

## Motivation
Keeps page routes concise and removes duplicate code between SSG/SSR/Editing. This also sets us up for upcoming integration to GraphQL/Experience Edge.

## Types of changes

## Checklist:


samples/nextjs/src/lib/page-props-factory.ts
```diff
@@ -0,0 +1,149 @@
...
+    // Get normalized Sitecore item path
+    const path = extractPath(context.params);
+    // Use context locale if Next.js i18n is configured, otherwise use language defined in package.json
+    const locale = context.locale ?? packageConfig.language;
+
+    let notFound = false;
+
+    // Retrieve layoutData from Layout Service
+    const layoutData = await this.layoutService
+      .fetchLayoutData(path, locale)
```
sc-illiakovalenko, line 79: @ambrauer We don't have an option to execute `fetchLayoutData` using `req`, `res`, as it was previously, so it's only working for SSG

samples/nextjs/src/components/GraphQL-ConnectedDemo/index.tsx
```diff
@@ -98,9 +99,9 @@ const GraphQLConnectedDemo = (props: StyleguideComponentProps): JSX.Element => {
 
 /**
  * Will be called during SSG
- * @param rendering
- * @param layoutData
- * @param context
+ * @param rendering {ComponentRendering}
+ * @param layoutData {LayoutServiceData | null}
```
sc-illiakovalenko, line 103: @ambrauer Probably, it can't be null? Should we change in `ComponentPropsFetchFunction` (in sitecore-jss-nextjs/component-props.ts) layoutData to be only LayoutServiceData? I see that you removed null option

samples/nextjs/src/components/GraphQL-ConnectedDemo/index.tsx
```diff
@@ -122,9 +123,9 @@ export const getStaticProps: GetStaticComponentProps = async (rendering, layoutD
 
 /**
  * Will be called during SSR
- * @param rendering
- * @param layoutData
- * @param context
+ * @param rendering {ComponentRendering}
+ * @param layoutData {LayoutServiceData | null}
```
sc-illiakovalenko, line 127: @ambrauer Same as above

samples/nextjs/src/components/GraphQL-ConnectedDemo/index.tsx
```diff
@@ -144,4 +145,24 @@ export const getServerSideProps: GetServerSideComponentProps = async (rendering,
   return result.data;
 };
 
+/**
+ * Will be called during editing
+ * @param rendering {ComponentRendering}
+ * @param layoutData {LayoutServiceData | null}
```
sc-illiakovalenko, line 151: @ambrauer Same as above

samples/nextjs/src/lib/page-props-factory.ts
```diff
@@ -0,0 +1,149 @@
...
+    // Stubbed out as getter for potential logic here (e.g. based on constructor props)...
+    return dictionaryService;
+  }
+
+  /**
+   * Create SitecorePageProps in an SSG/SSR context (within getStaticProps/getServerSideProps function).
+   * @param context {GetStaticPropsContext|GetServerSidePropsContext}
+   * @see SitecorePageProps
+   */
+  public async create<T extends GetStaticPropsContext | GetServerSidePropsContext>(
```
sc-illiakovalenko, line 67: @ambrauer May be it's better to create separate a separate func: `createForSSR`, `createForSSG`?

samples/nextjs/src/lib/page-props-factory.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @param context {GetStaticPropsContext|GetServerSidePropsContext}
+ */
+const isGetServerSidePropsContext = function (
+  context: GetServerSidePropsContext | GetStaticPropsContext
+): context is GetServerSidePropsContext {
+  return 'req' in context;
+};
+
+export class SitecorePagePropsFactory {
+  private get componentPropsService(): ComponentPropsService {
```
sc-illiakovalenko, line 44: @ambrauer Do we need to create new instance of `ComponentPropsService` every time?

samples/nextjs/src/pages/_edit.tsx
```diff
@@ -23,46 +18,19 @@ const SitecorePage = ({ layoutData }: SitecorePageProps): JSX.Element => {
...
-  // if (props.layoutData) {
-  //   // Retrieve component props using side-effects defined on components level
-  //   props.componentProps = await componentPropsService.fetchInitialComponentProps({
-  //     layoutData: props.layoutData,
-  //     context,
-  //     componentModule,
-  //   });
-  // }
-
+  const propsFactory = new SitecorePagePropsFactory();
```
sc-illiakovalenko, line 32: @ambrauer For me it makes sense to create a `lib/page-props-factory.ts` and define it there, because this class doesn't contain arguments, and we don't need to create separate in every nextjs `page`

sc-illiakovalenko (changes requested):

samples/nextjs/src/lib/page-props-factory.ts
ambrauer: @sc-illiakovalenko Great catch, those got lost in the shuffle. Fixed.

samples/nextjs/src/components/GraphQL-ConnectedDemo/index.tsx
ambrauer: @sc-illiakovalenko You're right, that's no longer nullable. Fixed.

samples/nextjs/src/lib/page-props-factory.ts
ambrauer: @sc-illiakovalenko I went back and forth on this, but reading through this after a weekend away, I'm with you. Separate would be clearer, especially since there's a dedicated `createForEditing` already.

samples/nextjs/src/lib/page-props-factory.ts
ambrauer: @sc-illiakovalenko Moved to private var / constructor init.

samples/nextjs/src/pages/_edit.tsx
ambrauer: @sc-illiakovalenko I agree, though eventually this may have constructor parms, it makes sense to export an instance and use that.

sc-illiakovalenko (approved):

samples/nextjs/src/components/GraphQL-ConnectedDemo/index.tsx
```diff
@@ -144,4 +145,24 @@ export const getServerSideProps: GetServerSideComponentProps = async (rendering,
   return result.data;
 };
 
+/**
+ * Will be called during editing
+ * @param rendering {ComponentRendering}
+ * @param layoutData {LayoutServiceData}
+ * @param context {NextPageContext}
```
sc-illiakovalenko, line 152: @ambrauer Probably, need to place `type` before param name

ambrauer: Ah yes, fixed

PR #523, opened 2020-12-29
Feature/439767 - Configure e2e tests for next.js.

sc-olexandrtroyanovskyy:
## Types of changes

## Checklist:


e2e/appmodes.js
```diff
@@ -19,6 +19,13 @@ if (!verb || (verb !== 'connected' && verb !== 'disconnected')) {
 }
 
 const config = [
+  // Due to current inablitiy to set\change a port for the next.js app
```
ambrauer, line 22: The Next.js CLI [supports a port parameter](https://nextjs.org/docs/api-reference/cli#development) for the `dev` and `start` commands, which should be able to read an env variable. What if you change the Next.js sample app's "next:dev" to:

```
"next:dev": "cross-env NODE_OPTIONS='--inspect' next dev --port $PORT",
```

It seems to use 3000 by default if `PORT` isn't defined (which is what we'd want), but should pick up our e2e setting I think?

ambrauer (changes requested):

e2e/appmodes.js
sc-illiakovalenko: @ambrauer If you use `cross-env` it will not pick up PORT env variable, I tried to do it (if you use `appmodes.js` script).
In case If I use `cross-env-shell NODE_OPTIONS='--inspect' next dev --port $PORT` it will work only if $PORT is defined, if this variable is not provided, it will throw error: `UnhandledPromiseRejectionWarning: Error: Option requires argument: --port`

samples/nextjs/src/components/Styleguide-Multilingual/index.tsx
```diff
@@ -33,7 +33,7 @@ const StyleguideMultilingual = (props: StyleguideMultilingualProps): JSX.Element
       </p>
 
       <p>
-        <Link href={`${publicUrl}/styleguide`} locale="en">
+        <Link href={`${publicUrl}/en/styleguide`} locale={false}>
```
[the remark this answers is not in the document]
sc-olexandrtroyanovskyy: Comment has been added to bring in an explanation of using both techniques, thank you.

e2e/cypress/package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "edit": "cypress open",
     "test": "cypress run",
-    "test:integrated": "npm run test:integrated:vue && npm run test:integrated:react && npm run test:integrated:angular",
+    "test:integrated": "npm run test:integrated:vue && npm run test:integrated:react && npm run test:integrated:angular && npm run test:integrated",
```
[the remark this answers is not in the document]
sc-olexandrtroyanovskyy: That is correct, the initial intention was to add && npm run test:integrated:nextjs and then later it was not properly removed from the file, thank you for this catch, Adam.

e2e/appmodes.js
ambrauer: @sc-illiakovalenko That's a bummer. I attempted to use a fork of `cross-env` which [adds default value functionality](https://www.npmjs.com/package/@naholyr/cross-env#default-value). That solved the "option requires argument" error, however didn't pick up the port supplied by `appmodes.js`. 

I suppose the other option would be to rewrite `appmodes.js` to fire up and test each app sequentially - avoiding any port conflicts - but that's beyond the scope of this task.

ambrauer (approved):

PR #526, opened 2021-01-05
[NextJS] Base Dictionary Service

sc-illiakovalenko:
## Description
Similar to layout-service, create base service for dictionary
Move and productize the dictionary-service.ts that's in the Next.js sample to sitecore-jss package
Add basic caching mechanism with short TTL (disabled by default) to reduce identical requests, similar to headless-ssr-proxy

## Types of changes

## Checklist:


packages/sitecore-jss/package.json
```diff
@@ -46,6 +47,7 @@
   },
   "dependencies": {
     "lodash.unescape": "^4.0.1",
+    "memory-cache": "^0.2.0",
```
ambrauer, line 50: Is there a reason [memory-cache](https://www.npmjs.com/package/memory-cache) was used over [node-cache](https://www.npmjs.com/package/node-cache)? The node-headless-ssr-proxy sample uses `node-cache` for dictionary caching, `memory-cache` for general request middleware (not sure why the 2 caching packages are necessary). Seems like `node-cache` is more popular/active?

packages/sitecore-jss/src/dictionary-service.ts
```diff
@@ -0,0 +1,85 @@
+import { AxiosResponse } from 'axios';
+import mcache from 'memory-cache';
+import { AxiosDataFetcher } from './data-fetcher';
+import { DictionaryPhrases, DictionaryServiceData } from './dataModels';
+import { HttpJsonFetcher } from './httpClientInterface';
+
+export type DictionaryServiceConfig = {
+  apiHost: string;
```
ambrauer, line 8: Should add JSDoc for these 3 params (looks like layout-service is missing these JSDoc comments as well).

packages/sitecore-jss/src/dictionary-service.ts
```diff
@@ -0,0 +1,85 @@
...
+  apiKey: string;
+  jssAppName: string;
+  /**
+   * Custom data fetcher
+   * @see HttpJsonFetcher<T>
+   */
+  dataFetcher?: HttpJsonFetcher<DictionaryServiceData>;
+  /**
+   * Enable/disable caching mechanism
+   * @default false
```
ambrauer, line 18: I think we should enable by default, and bump the value to 60000. This matches the node-headless-ssr-proxy.

packages/sitecore-jss/src/dictionary-service.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    if (cacheEnabled) {
+      const cachedBody = mcache.get(dictionaryServiceUrl);
+
+      if (cachedBody) {
+        return cachedBody;
+      }
+    }
+
+    const response = await fetcher(dictionaryServiceUrl);
```
ambrauer, line 54: I think we should expose and make use of the `fetchData` function that is part of `dataApi.ts`. Once the `fetchRouteData` and `fetchPlaceholderData` functions are fully removed, our layout-service code will need to be refactored to use this as well. This sets us up to consistently use the error checking and query parameter construction that it uses (to feed the dataFetcher / HttpJsonFetcher<T>).

packages/sitecore-jss/src/dictionary-service.ts
```diff
@@ -0,0 +1,85 @@
...
+
+export type DictionaryServiceConfig = {
+  apiHost: string;
+  apiKey: string;
+  jssAppName: string;
+  /**
+   * Custom data fetcher
+   * @see HttpJsonFetcher<T>
+   */
+  dataFetcher?: HttpJsonFetcher<DictionaryServiceData>;
```
ambrauer, line 15: Do we want to follow the `dataFetcherResolver` pattern like layout-service for consistency?

ambrauer (changes requested):
See suggestions

packages/sitecore-jss/package.json
sc-illiakovalenko: @ambrauer Agree, changed to `node-cache`

packages/sitecore-jss/src/dictionary-service.ts
sc-illiakovalenko: In this case - yes

packages/sitecore-jss/src/dictionary-service.ts
sc-illiakovalenko: Added

packages/sitecore-jss/src/dictionary-service.ts
sc-illiakovalenko: I'm not sure that we need it, I remember that we used this approach, because we had a different `return` types for dataFetcher: `layoutServiceData`, `placeholderData`. And now we have only `DictionaryPhrases`. Are you agree?

packages/sitecore-jss/src/dictionary-service.ts
sc-illiakovalenko: Good idea, I used `fetchData` in `dictionaryService`

sc-illiakovalenko:
@ambrauer Added doc update

packages/sitecore-jss/src/dictionary-service.ts
ambrauer: @sc-illiakovalenko I think it was more about needing to pass along the req and res parameters (i.e. for header access/manipulation). Not sure we'd ever need to do that for dictionary, but I'm just thinking about consistency in our APIs.

sc-illiakovalenko: @ambrauer Agree, but seems like we didn't make manipulations with headers when request dictionary

ambrauer: @sc-illiakovalenko Correct, we don't - and the req/res parameters are optional with the `DataFetcherResolver` anyway. Guess I'm just thinking we could use the same config mechanism. i.e.:
1. move `DataFetcherResolver` from `layout-service.ts` to `data-fetcher.ts`
2. switch dictionary-service to use same `dataFetcherResolver?: DataFetcherResolver;` for config

packages/sitecore-jss/src/dictionary-service.ts
```diff
@@ -0,0 +1,102 @@
...
+
+    if (cacheEnabled) {
+      const cachedBody = this.dictionaryCache.get<DictionaryPhrases>(dictionaryServiceUrl);
+
+      if (cachedBody) {
+        return cachedBody;
+      }
+    }
+
+    const response = await fetchData<DictionaryServiceData>(dictionaryServiceUrl, fetcher);
```
ambrauer, line 71: FYI, could use optional "params" here for query string portion of url (sc_apikey).

ambrauer (changes requested):

packages/sitecore-jss/package.json
```diff
@@ -27,6 +27,7 @@
     "@types/chai-spies": "^1.0.3",
     "@types/chai-string": "^1.4.2",
     "@types/lodash.unescape": "^4.0.6",
+    "@types/memory-cache": "^0.2.1",
```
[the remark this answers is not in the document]
sc-illiakovalenko: Removed

packages/sitecore-jss/src/dictionary-service.ts
sc-illiakovalenko: Good idea, changed

ambrauer (approved):

PR #530, opened 2021-01-15
[NextJS] GraphQL Request Client + GraphQL Sitemap Service

sc-illiakovalenko:
## Description
Generates a sitemap using Graphql queries

## Types of changes

## Checklist:


samples/nextjs/src/pages/[[...path]].tsx
```diff
@@ -36,14 +39,28 @@ const SitecorePage = ({ notFound, layoutData, componentProps }: SitecorePageProp
...
   // See https://nextjs.org/docs/basic-features/data-fetching#incremental-static-regeneration
   //
-  // Ultimately, this is where we'll also be able to request a "sitemap" from Sitecore.
+
+  if (process.env.BUILD_MODE === 'export') {
+    const ROOT_ITEM = `/sitecore/content/${config.jssAppName}/home`;
+    const paths = await graphQLSitemapService.fetchSitemap(
+      context.locales || [],
+      ROOT_ITEM,
+      nextConfig.i18n.defaultLocale
```
Comment (ambrauer) [inline samples/nextjs/src/pages/[[...path]].tsx:55]: Should be able to use `context.defaultLocale` here

samples/nextjs/src/pages/[[...path]].tsx
```diff
@@ -36,14 +39,28 @@ const SitecorePage = ({ notFound, layoutData, componentProps }: SitecorePageProp
...
+export const getStaticPaths: GetStaticPaths = async (context) => {
   // Fallback, along with revalidate in getStaticProps (below),
   // enables Incremental Static Regeneration. This allows us to
   // leave certain (or all) paths empty if desired and static pages
   // will be generated on request.
   // See https://nextjs.org/docs/basic-features/data-fetching#incremental-static-regeneration
   //
-  // Ultimately, this is where we'll also be able to request a "sitemap" from Sitecore.
+
+  if (process.env.BUILD_MODE === 'export') {
```
Comment (ambrauer) [inline samples/nextjs/src/pages/[[...path]].tsx:50]: As discussed, we don't have to worry about export in our sample, but we should definitely have a condition here to enable/disable the sitemap fetch, especially in development since it [runs on every request](https://nextjs.org/docs/basic-features/data-fetching#runs-on-every-request-in-development-1). 

Could keep it simple and just use `NODE_ENV` for now:
 ```
  if (process.env.NODE_ENV !== 'development') {
```
Could also introduce a new .env variable, but would need to have comments there to document usage (not needed for SSR, etc).

**Review by ambrauer — CHANGES_REQUESTED** [review]

**Comment by nickwesselman** [comment]
Guessing it's a separate backlog item but will there be an ability to swap out the sitemap service for Disconnected mode, e.g. to use the manifest to get route paths?

File: samples/nextjs/src/pages/[[...path]].tsx (hunk 2 above)
Comment (nickwesselman) [reply samples/nextjs/src/pages/[[...path]].tsx:50]: The locale is still needed/desired for Incremental SSG right?

File: packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts [hunk 3]
```diff
@@ -0,0 +1,155 @@
...
+
+    return data?.item.id;
+  }
+
+  /**
+   * Graphql query in order to fetch sitemap
+   * @param {string} rootItemId
+   * @param {locale} locale
+   */
+  getSitemapQuery(rootItemId: string, locale: string): string {
```
nickwesselman, line 66: Is there a way to override this query, or make it more flexible? What if I want to limit the paths that are generated at build time, because I am using `fallback: true` ? e.g. What if I wanted to exclude my large product catalog? Would be nice if a custom query could be provided w/out needing to recreate all the other logic/invocation around it.

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -0,0 +1,155 @@
+import { GraphQLRequestClient } from '@sitecore-jss/sitecore-jss';
+
+export type GraphQLSitemapServiceConfig = {
+  endpoint: string;
```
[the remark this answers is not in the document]
sc-illiakovalenko: Added

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -0,0 +1,155 @@
...
+};
+
+type SearchResult = {
+  search: {
+    results: { url: { path: string } }[];
+  };
+};
+
+export class GraphQLSitemapService {
+  constructor(private config: GraphQLSitemapServiceConfig) {}
```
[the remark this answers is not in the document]
sc-illiakovalenko: Added

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * Returns new graphql client instance
+   */
+  createClient(): GraphQLRequestClient {
```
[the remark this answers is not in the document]
sc-illiakovalenko: Agree, changed

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -0,0 +1,155 @@
...
+    defaultLocale: string
+  ): Promise<StaticPath[]> {
+    if (!locales.length) {
+      return [];
+    }
+
+    const rootItemId = await this.getRootItemId(rootItemPath);
+
+    if (!rootItemId) {
+      console.error('Сan not fetch sitemap: root item id is not provided');
```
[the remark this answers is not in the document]
sc-illiakovalenko: It's looking better

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -0,0 +1,155 @@
...
+      return [];
+    }
+
+    const getStaticPaths = async (locale: string): Promise<StaticPath[]> => {
+      const query = this.getSitemapQuery(rootItemId, locale);
+
+      const data = await this.createClient()
+        .request<SearchResult>(query)
+        .catch((error) => {
+          console.error('Error occured during fetching sitemap:', error.response);
```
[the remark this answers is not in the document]
sc-illiakovalenko: Changed

packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -0,0 +1,30 @@
+import { GraphQLClient } from 'graphql-request';
+
+export class GraphQLRequestClient {
+  constructor(private endpoint: string) {}
```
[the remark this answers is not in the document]
sc-illiakovalenko: Added

samples/nextjs/src/pages/[[...path]].tsx
Comment (sc-illiakovalenko) [reply samples/nextjs/src/pages/[[...path]].tsx:55]: Don't need `defaultLocale` - we decided to use single language for Export. And we will not have language embedding

samples/nextjs/src/pages/[[...path]].tsx
Comment (sc-illiakovalenko) [reply samples/nextjs/src/pages/[[...path]].tsx:50]: @ambrauer I will use `NODE_ENV` it doesn't make sense to define separate variable. If somebody will need it - they can define it.

@nickwesselman Yes, I'm using `locale` property.

sc-illiakovalenko:
> Guessing it's a separate backlog item but will there be an ability to swap out the sitemap service for Disconnected mode, e.g. to use the manifest to get route paths?

I'm not sure how it will be implemented for disconnected mode, but may be we can define separate method for disconnected mode, or create a separate service, it can be replace in `getStaticPaths`

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
sc-illiakovalenko: @nickwesselman Added ability to customize search query part. But I don't provide way to customize return type

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -0,0 +1,252 @@
...
+export class GraphQLSitemapService {
+  /**
+   * Provides ability to fetch sitemap from graphql endpoint.
+   * Sitemap can be used for SSG/Export
+   * @param {GraphQLSitemapServiceConfig} config graphql sitemap service config
+   */
+  constructor(private config: GraphQLSitemapServiceConfig) {}
+
+  /**
+   * Fetch sitemap which could be used for generation of static pages using Export mode
```
ambrauer, line 32: We might want to call out the difference w/r/t locale handling between these two. e.g.
"Fetch sitemap which could be used for generation of static pages during 'next export'. The 'locale' parameter will be used in the item query, but since i18n is not supported, the output paths will not include a 'locale' property."

samples/nextjs/src/pages/[[...path]].tsx
```diff
@@ -36,14 +38,27 @@ const SitecorePage = ({ notFound, layoutData, componentProps }: SitecorePageProp
 
 // This function gets called at build and export time to determine
 // pages for SSG ("paths", as tokenized array).
-export const getStaticPaths: GetStaticPaths = async () => {
+export const getStaticPaths: GetStaticPaths = async (context) => {
   // Fallback, along with revalidate in getStaticProps (below),
   // enables Incremental Static Regeneration. This allows us to
   // leave certain (or all) paths empty if desired and static pages
   // will be generated on request.
```
Comment (ambrauer) [inline samples/nextjs/src/pages/[[...path]].tsx:45]: We should tie in the sitemap generation to this comment. Something like

```suggestion
  // Fallback, along with revalidate in getStaticProps (below),
  // enables Incremental Static Regeneration. This allows us to
  // leave certain (or all) paths empty if desired and static pages
  // will be generated on request (development mode in this example).
  // Alternatively, the entire sitemap could be pre-rendered 
  // ahead of time (non-development mode in this example).
```

samples/nextjs/src/pages/[[...path]].tsx
```diff
@@ -36,14 +38,27 @@ const SitecorePage = ({ notFound, layoutData, componentProps }: SitecorePageProp
...
   // See https://nextjs.org/docs/basic-features/data-fetching#incremental-static-regeneration
-  //
-  // Ultimately, this is where we'll also be able to request a "sitemap" from Sitecore.
+
+  // use graphQLSitemapService.fetchExportSitemap(context.defaultLocale, ROOT_ITEM)
```
Comment (ambrauer) [inline samples/nextjs/src/pages/[[...path]].tsx:48]: Should this note be further down next to the `fetchSSGSitemap` call? Part of me wants to just leave this export comment out and leave it up to the export documentation...

ambrauer (approved):
Approved, with some suggested doc/comment changes

samples/nextjs/src/pages/[[...path]].tsx
Comment (sc-illiakovalenko) [reply samples/nextjs/src/pages/[[...path]].tsx:48]: Agree, can remove - will add it to doc for export

samples/nextjs/src/pages/[[...path]].tsx
Comment (sc-illiakovalenko) [reply samples/nextjs/src/pages/[[...path]].tsx:45]: Looks good, changed

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
sc-illiakovalenko: Thank you, changed

PR #532, opened 2021-01-18
[NextJS] Fix dev mode fast refresh in Experience Editor

sc-illiakovalenko:
## Description

When Next.js is run in dev mode, the fast refresh feature breaks the Experience Editor chrome. We should force the EE to rebind the chromes and/or reload on Fast Refresh.

Added `getPublicUrl` from https://github.com/Sitecore/jss/pull/529 PR

## Types of changes

## Checklist:


samples/nextjs/src/pages/[[...path]].tsx
```diff
@@ -1,13 +1,22 @@
...
+  ComponentPropsContext,
+  subscribeOnRebuildAppEvent,
+} from '@sitecore-jss/sitecore-jss-nextjs';
 import { StyleguideSitecoreContextValue } from 'lib/component-props';
 import { SitecorePageProps } from 'lib/page-props';
 import { sitecorePagePropsFactory } from 'lib/page-props-factory';
 import { componentFactory } from 'temp/componentFactory';
 
 const SitecorePage = ({ notFound, layoutData, componentProps }: SitecorePageProps): JSX.Element => {
+  useEffect(() => {
```
Comment (nickwesselman) [inline samples/nextjs/src/pages/[[...path]].tsx:16]: Can we add a comment to the sample here so devs know why it's there?

packages/sitecore-jss-nextjs/src/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+  return url.toString().replace(/\/$/, '');
+};
+
+/**
+ * Since Experience Editor does not support Fast Refresh:
+ * 1. Subscribe on events provided by webpack.
+ * 2. Reset experience editor chromes when build is finished
+ * @param {Function} onBuildEnd custom logic to be executed after build is finished
+ */
+export const subscribeOnRebuildAppEvent = (onBuildEnd?: (payload: unknown) => void): void => {
```
ambrauer, line 30: The logic here is specific to Experience Editor, I think the function name should reflect that better (maybe helps a bit with Nick's comment as well). "handleEditingFastRefresh"?

packages/sitecore-jss-nextjs/src/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+  if (isExperienceEditorActive()) {
+    const eventSource = new window.EventSource(`${getPublicUrl()}/_next/webpack-hmr`);
+
+    window.addEventListener('beforeunload', () => eventSource.close());
+
+    eventSource.onopen = () => console.log('[Experience Editor HMR Listener] Online');
```
ambrauer, line 36: Should maybe use Next.js's term "[Fast Refresh](https://nextjs.org/docs/basic-features/fast-refresh)" instead of "HMR" in our logs.

packages/sitecore-jss-nextjs/src/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+    eventSource.onmessage = (event) => {
+      if (event.data.indexOf('{') === -1) return; // heartbeat
+
+      const payload = JSON.parse(event.data);
+
+      console.debug(`[Experience Editor HMR Listener] Saw event: ${JSON.stringify(payload)}`);
+
+      if (payload.action !== 'built') return;
+
+      if (onBuildEnd) {
```
ambrauer, line 47: Instead of the `onBuildEnd` parameter, should we just make the implementation support both alternatives (`resetExperienceEditorChromes()` and `window.location.reload()`) and make the parameter a boolean allowing selection? Maybe `fullReload`, which is false by default (use resetExperienceEditorChromes() by default).

Feel like this might make the `onBuildEnd` intent more obvious.

ambrauer (changes requested):

packages/sitecore-jss-nextjs/src/utils.ts
sc-illiakovalenko: Will rename it to `handleExperienceEditorFastRefresh`

packages/sitecore-jss-nextjs/src/utils.ts
sc-illiakovalenko: Yes, it's more related to Fast Refresh than to Hot module replacement term

packages/sitecore-jss-nextjs/src/utils.ts
sc-illiakovalenko: Agree, I didn't know how to exactly name this hook - and `onBuildEnd` is not so obvious.
Will rename it to `forceReload`

samples/nextjs/src/pages/[[...path]].tsx
Comment (sc-illiakovalenko) [reply samples/nextjs/src/pages/[[...path]].tsx:16]: Added

nickwesselman (approved):

ambrauer (approved):
Looks good!

PR #535, opened 2021-01-20
[Next.js] [EE] Finalize editing secret following JSS.Server implementation

ambrauer:
This is a follow-up to #529, which finalizes the editing secret now that JSS.Server implementation is complete. 
Updated names (SITECORE_SECURITY_TOKEN > JSS_EDITING_SECRET and SecurityToken > JssEditingSecret) and added doc comments.

### DIFF
--- a/samples/nextjs/sitecore/config/JssNextWeb.config
+++ b/samples/nextjs/sitecore/config/JssNextWeb.config
@@ -23,10 +23,10 @@
         JSS EDITING SECRET
         To secure the Experience Editor endpoint exposed by your Next.js app (see `serverSideRenderingEngineEndpointUrl` below),
         a secret token is used. This is taken from an env variable by default, but could be patched and set directly by uncommenting.
-        This (server-side) value must match your client-side value, which is configured by the JSS_EDITING_SECRET env variable (see .env file).
+        This (server-side) value must match your client-side value, which is configured by the JSS_EDITING_SECRET env variable (see the Next.js .env file).
         We recommend an alphanumeric value of at least 16 characters.
 
-        <setting name="JavaScriptServices.ViewEngine.Http.JssEditingSecret" value="$(env:SITECORE_JSS_EDITING_SECRET)" />
+        <setting name="JavaScriptServices.ViewEngine.Http.JssEditingSecret" value="" />
      -->
     </settings>
     <sites>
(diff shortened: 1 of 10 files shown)


ambrauer:
@nickwesselman Could you please take a look at the `.env` and `JssNextWeb.config` changes and confirm these make sense (from product standpoint / container setup / secret complexity)? You can ignore the rest of the updates as these are just renames.

samples/nextjs/sitecore/config/JssNextWeb.config
```diff
@@ -18,6 +18,16 @@
         thus making analytics track the correct original client IP address.
       -->
       <setting name="Analytics.ForwardedRequestHttpHeader" set:value="X-Forwarded-For" />
+
+      <!--
+        JSS EDITING SECRET
+        To secure the Experience Editor endpoint exposed by your Next.js app (see `serverSideRenderingEngineEndpointUrl` below),
+        a secret token is used. This is taken from an env variable by default, but could be patched and set directly by uncommenting.
+        This (server-side) value must match your client-side value, which is configured by the JSS_EDITING_SECRET env variable (see .env file).
```
nickwesselman, line 26: > see .env file

This could be confused for the docker-compose `.env`

nickwesselman (commented):
@ambrauer I'm a little confused about the difference in names between the two env vars, I guess there are different conventions in platform vs JSS though? Suggested the `SITECORE_` prefix for the server since all the other env vars there use it. Better to be consistent between the two environments, or with other env vars used in JSS?

ambrauer:
> @ambrauer I'm a little confused about the difference in names between the two env vars, I guess there are different conventions in platform vs JSS though? Suggested the `SITECORE_` prefix for the server since all the other env vars there use it. Better to be consistent between the two environments, or with other env vars used in JSS?

Yeah, I was struggling with that. I'm not sure there are any (as of yet) shared between server and client. Personally I think it'd be just as confusing if they _did_ match (since they really are independent, I could see cases of "I SET THIS ALREADY!"). I lean toward consistency with other env vars in Node / JSS (we've got things like `JSS_APP_NAME` so far). And I thought this lined up nicely for non-container environments who are just setting the <setting "JavaScriptServices.ViewEngine.Http.**JssEditingSecret**"> directly.

samples/nextjs/sitecore/config/JssNextWeb.config
ambrauer: Added clarification "see the Next.js .env file"

nickwesselman (approved):

PR #536, opened 2021-01-21
Specify buildArtifactsPath in samples/nextjs/package.json

anastasiya29:
This var is needed for Sitecore import

## Description
Next.js doesn't use a 'dist' folder for build artifacts, so we need to specify the build output path. Otherwise, `jss deploy` fails.

## How Has This Been Tested?
Locally tested deploying next.js app to Sitecore with and without this variable.

### DIFF
--- a/samples/nextjs/package.json
+++ b/samples/nextjs/package.json
@@ -11,7 +11,8 @@
     "sitecoreDistPath": "/dist/JssNextWeb",
     "sitecoreConfigPath": "/App_Config/Include/zzz",
     "graphqlEndpointPath": "/api/jssnextweb",
-    "language": "en"
+    "language": "en",
+    "buildArtifactsPath": "./.next"
   },
   "engines": {
     "node": ">=8.1",


sc-illiakovalenko:
@anastasiya29 Why do we need to deploy `.next` to sitecore instance? How this build folder will be used?
Dev's should not use `jss deploy files` in next.js sample app

sc-illiakovalenko (changes requested):
@anastasiya29 Why do we need to deploy `.next` to sitecore instance? How this build folder will be used?
Dev's should not use `jss deploy files` in next.js sample app

ambrauer:
Agree, since there is no integrated mode for Next.js, the `jss deploy files` isn't applicable - just `config` and `items`. We've got a task in the backlog to clarify/document this (439766).

anastasiya29:
> Agree, since there is no integrated mode for Next.js, the `jss deploy files` isn't applicable - just `config` and `items`. We've got a task in the backlog to clarify/document this (439766).

Are you saying that using the Nextjs server is mandatory for all cases? ie the "nodejs" rendering engine cannot be used even if someone wants to do it that way for some reason?

ambrauer:
> Are you saying that using the Nextjs server is mandatory for all cases? ie the "nodejs" rendering engine cannot be used even if someone wants to do it that way for some reason?

Correct, essentially the "traditional" JSS node rendering engine is incompatible with Next.js.

PR #537, opened 2021-01-21
[NextJS] Document Next.js Static HTML Export

sc-illiakovalenko:
## Description

Maintaining static html export capability is becoming increasingly complicated with new releases of Next.js. In version 10, for example, both i18n and image optimization are not compatible with next export.
We should define documentation to help customer enable next export

## Types of changes

## Checklist:

### DIFF
--- a/docs/data/routes/docs/client-frameworks/nextjs/nextjs-static-export/en.md
+++ b/docs/data/routes/docs/client-frameworks/nextjs/nextjs-static-export/en.md
@@ -8,11 +8,10 @@ title: Next.js static export
 
 Using JSS Next.js sample app you can run `jss export` that allows you to export your app to static HTML, which can be run standalone without the need of a Node.js server. Read more about Next.js [Static HTML Export](https://nextjs.org/docs/advanced-features/static-html-export).
 
-> Currently here are limitations for usage of Static Export: 
+> Currently, these are the limitations of the Next.js sample app for usage of Static HTML Export:
 > 1. i18n not supported, so multilingual apps can't be exported.
 > 1. Visitor Identification not available.
 > 1. Request rewrites not supported.
-> 1. API Routes are not supported because they can't be prerendered to HTML.
 
 These instructions you should apply in order to run `next export`:
 1. Remove `i18n`:
@@ -24,11 +23,11 @@ These instructions you should apply in order to run `next export`:
 	import { GetStaticPaths } from 'next';
 	import { graphQLSitemapService } from 'lib/graphql-sitemap-service';
 	import config from 'temp/config';
-	import packageConfig from '../../package.json';
+	import { config as packageConfig } from '../../package.json';
 
 	export const getStaticPaths: GetStaticPaths = async (context) => {
 	  const ROOT_ITEM = `/sitecore/content/${config.jssAppName}/home`;
-	  const paths = await graphQLSitemapService.fetchExportSitemap(packageConfig.config.language, ROOT_ITEM);
+	  const paths = await graphQLSitemapService.fetchExportSitemap(packageConfig.language, ROOT_ITEM);
 
 	  return {
     	paths,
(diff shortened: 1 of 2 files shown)


nickwesselman:
@sc-illiakovalenko Removing the API routes breaks Experience Editor right? I would expect that customers would want to still run a Next.js instance for EE support, while using `next export`. Is there an alternative to removing them entirely?

sc-illiakovalenko:
@nickwesselman It's not required to remove api/routes, user will see warning during export. I can leave api routes. Should I leave them?

docs/data/routes/docs/client-frameworks/nextjs/nextjs-static-export/en.md
```diff
@@ -0,0 +1,42 @@
...
+name: nextjs-static-export
+routeTemplate: ./data/component-templates/article.yml
+title: Next.js static export
+---
+
+# Next.js Static HTML Export
+
+Using JSS Next.js sample app you can run `jss export` that allows you to export your app to static HTML, which can be run standalone without the need of a Node.js server. Read more about Next.js [Static HTML Export](https://nextjs.org/docs/advanced-features/static-html-export).
+
+> Currently here are limitations for usage of Static Export:
```
ambrauer, line 11: 1. Might be good to callout whether each limitation is Next.js' or Sitecore's.
2. Add Next.js Image Optimization (next/image) limitation
3. Should we remove the API Routes one?

docs/data/routes/docs/client-frameworks/nextjs/nextjs-static-export/en.md
```diff
@@ -0,0 +1,42 @@
...
+	  return {
+    	paths,
+    	fallback: false,
+	  };
+	}
+	```
+	> The `fallback: true` mode of `getStaticPaths` is not supported when using next export.
+1. Remove usage of `<VisitorIdentification />` component in `src/Layout.tsx`, since Visitor Identification not available.
+1. Define `PUBLIC_URL` in `.env`.
+1. Add script `next build && next export` in package.json.
```
ambrauer, line 42: `bootstrap` should probably be part of this as well. Maybe we provide an entire script example, i.e.:
```
"export": "npm-run-all --serial bootstrap next:build next:export",
```

docs/data/routes/docs/client-frameworks/nextjs/nextjs-static-export/en.md
```diff
@@ -0,0 +1,42 @@
...
+1. Remove `i18n`:
+	1. Delete i18n configuration in `next.config.js`.
+	1. Define language in `package.json` using `config.language` which will be used during export.
+1. Remove `rewrites` in `next.config.js`.
+1. Define `getStaticPaths` in `[[...path]].tsx` and use:
+	```js
+	import { GetStaticPaths } from 'next';
+	import { graphQLSitemapService } from 'lib/graphql-sitemap-service';
+	import config from 'temp/config';
+	import packageConfig from '../../package.json';
```
ambrauer, line 27: Could do `
import { config as packageConfig } from '../../package.json';` here, then use `packageConfig.language` below.

ambrauer (changes requested):
Just a few comments

docs/data/routes/docs/client-frameworks/nextjs/nextjs-static-export/en.md
sc-illiakovalenko: @ambrauer As Nick mentioned, we will not remove api routes, as I understood
And we are not using Image Optimization even in SSG, should we mention it?

ambrauer: @sc-illiakovalenko I was referring to the limitation bullet "API Routes are not supported because they can't be prerendered to HTML.". Should that be mentioned still?

True, maybe it's best that we don't mention Image Optimization since we no longer include it. I think we should be clear then that this list is only the limitations of our sample app. e.g. "Currently, these are the limitations of the Next.js sample app for usage of Static HTML Export:"

sc-illiakovalenko: @ambrauer Renamed `restrictions` - first line

docs/data/routes/docs/client-frameworks/nextjs/nextjs-static-export/en.md
sc-illiakovalenko: @ambrauer Rename, it's looking better

docs/data/routes/docs/client-frameworks/nextjs/nextjs-static-export/en.md
sc-illiakovalenko: @ambrauer Like it :)

ambrauer (approved):

PR #538, opened 2021-01-26
Fix/458679

anastasiya29:
Fix prop-types for errorComponent (this is a Placeholder prop in the react package)

## Motivation
This fixes a bug where passing a custom error component via the errorComponent prop (as described in [the docs](https://jss.sitecore.com/docs/client-frameworks/react/react-placeholders#error-components)) caused an "Invalid prop" error.

## How Has This Been Tested?
Unit tests added.
Manual tests on react sample app. 
1.  Create a new CustomError component
2.  Update Layout.js to pass a react component to the errorComponent prop to Placeholder `<Placeholder errorComponent={CustomError} />`
3.  Verify that Home page renders without errors in disconnected and in EE
4.  Update ContentBlock component to throw an error
5.  Verify that CustomError is rendered on Home page instead of ContentBlock
6.  Update Layout.js to pass a renderProp function to the errorComponent prop to Placeholder `<Placeholder errorComponent={() => <div>Error</div>} />`
7.  Verify that "Error" is rendered on Home page instead of ContentBlock

## Types of changes

## Checklist:

- [ x] I have read the Contributing guide.
- [ x] My code/comments/docs fully adhere to the Code of Conduct.

### DIFF
--- a/packages/sitecore-jss-react/package.json
+++ b/packages/sitecore-jss-react/package.json
@@ -8,7 +8,7 @@
     "lint": "eslint ./src/**/*.tsx ./src/**/*.ts",
     "test": "mocha --require ts-node/register --require ./src/tests/shim.ts ./src/tests/jsdom-setup.ts ./src/tests/enzyme-setup.ts \"./src/**/*.test.ts\" \"./src/**/*.test.tsx\"",
     "prepublishOnly": "npm run build",
-    "coverage": "nyc --require ts-node/register npm test"
+    "coverage": "nyc npm test"
   },
   "engines": {
     "node": ">=8.1"
@@ -43,6 +43,7 @@
     "eslint-plugin-react": "^7.21.5",
     "jsdom": "^15.1.1",
     "mocha": "^8.1.3",
+    "nyc": "^15.1.0",
     "react": "^17.0.1",
     "react-dom": "^17.0.1",
     "react-test-renderer": "^16.9.0",
--- a/packages/sitecore-jss-react/package-lock.json
+++ b/packages/sitecore-jss-react/package-lock.json
@@ -13,11 +13,166 @@
         "@babel/highlight": "^7.10.4"
       }
     },
+    "@babel/core": {
+      "version": "7.12.10",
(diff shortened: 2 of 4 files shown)


packages/sitecore-jss-react/package.json
```diff
@@ -8,7 +8,7 @@
     "lint": "eslint ./src/**/*.tsx ./src/**/*.ts",
     "test": "mocha --require ts-node/register --require ./src/tests/shim.ts ./src/tests/jsdom-setup.ts ./src/tests/enzyme-setup.ts \"./src/**/*.test.ts\" \"./src/**/*.test.tsx\"",
     "prepublishOnly": "npm run build",
-    "coverage": "nyc --require ts-node/register npm test"
```
ambrauer, line 11: Removing this means the coverage only includes things that have unit tests. So things like `VisitorIdentification.tsx`, which have 0 tests will not show up in the report. 

I remember fighting with this on the nextjs package ... Looks like in this case, you might just have to add the following to the top of the `utils.ts` to fix the 'style-attr' type definitions error:
```
// eslint-disable-next-line spaced-comment
/// <reference types="../global" />
```

ambrauer (changes requested):
Just the one comment re: coverage, and appears there may be linting errors based on the failed checks? Otherwise looks good to me.

ambrauer:
Also, should this fix be mentioned in doc release notes?

packages/sitecore-jss-react/package.json
anastasiya29: @ambrauer - got it. This package didn't even have nyc installed so we were not running a coverage report for it. After adding 'nyc' I made the script match what I saw us using in other packages. Specifically, the nextjs package also uses `"coverage": "nyc npm test"`. Seems like that one needs to be updated as well, or is there a reason why that one is different?

anastasiya29: As discussed with @ambrauer, will create a separate ticket for looking into the coverage script of react and nextjs packages

PR #539, opened 2021-01-26
Feature/458360

anastasiya29:
Refactor and simplify the component structure of the Next.js sample app to be in line with Next.js docs and other Next.js production projects.

The previous structure required a directory for every component, but it didn't provide a way to use directories to group related components. As a result, route names were included in component names to show grouping, which is not good practice.
For example, we had something like:
```
/src
  /components
    /Styleguide-SomeComponent
      /index.tsx
    /Styleguide-AnotherComponent
      /index.tsx
    /GraphQL-SomeComponent
      /index.tsx
    /GraphQL-AnotherComponent
      /index.tsx
/sitecore
  /definitions
    /components
      /Styleguide-SomeComponent.sitecore.tsx
      /Styleguide-AnotherComponent.sitecore.tsx
      /GraphQL-SomeComponent.sitecore.tsx
      /GraphQL-AnotherComponent.sitecore.tsx
```

With the updated `scaffold-component` and `generate-component-factory` scripts, we now support a cleaner structure with less nesting.
For example, we can have this:

```
/src
  /components
    /styleguide
      /SomeComponent.tsx
      /AnotherComponent.tsx
    /graphql
      /SomeComponent.tsx
      /AnotherComponent.tsx
/sitecore
  /definitions
    /components
      /styleguide
        /SomeComponent.sitecore.tsx
        /AnotherComponent.sitecore.tsx
      /graphql
        /SomeComponent.sitecore.tsx
        /AnotherComponent.sitecore.tsx
```

## Changes to `generate-component-factory`
Instead of looking for child subdirectories that contain `index.tsx`, we now recursively loop through all directories and process any `.tsx` files that we find, with the exception of type definition files (`.d.tsx`)

See comments in script for more details

## Changes to `scaffold-component`
- Before, the scaffold script expected a component name that started with a capital letter, and included only letters, numbers, or dashes (`jss scaffold ComponentName` )
- Now we can also have underscores in the ComponentName
- Now we can provide a component name OR a relative path to the script (relative to componentRootPath). Ex. `jss scaffold some/new/path/ComponentName` will generate the files `src/components/some/new/path/ComponentName.tsx` and `sitecore/definitions/components/some/new/path/ComponentName.sitecore.tsx`. This makes the scaffold command useful for real projects, which likely organize components into folders.
- Boilerplate text for scaffolding files has been pulled out into separate files to improve readability and make customization easier.
See comments in script for more details

## How Has This Been Tested?
The components in the Next.js sample app have been reorganized to mirror this new structure. `npm run build` was used to test component factory generation.

`jss scaffold` was tested manually with valid and invalid inputs.

## Types of changes
This is a change to the Next.js sample app, which is not released yet, so no breaking changes.

## Checklist:
I have added details to the comments in the two scripts. Once these updates are reviewed, I will copy the changes into /docs

- [x] My change is a code change and it requires an update to the documentation.


samples/nextjs/scripts/templates/component-factory.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @returns component factory file contents
+ */
+function generateComponentFactory(imports: string[], registrations: string[]): string {
+  return `/* eslint-disable */
+// Do not edit this file, it is auto-generated at build time!
+// See scripts/generate-component-factory.ts to modify the generation of this file.
+${imports.join('\n')}
+
+const components = new Map();
+${registrations.join('\n')}
```
ambrauer, line 14: There's a template dependency baked into the `registrations` array: the **components** in `components.set(...);`. 
Would it be cleaner to pass this `generateComponentFactory` the component list (`ComponentFile[]`) instead and let it be responsible for building those import/registrations internally? That way the generated code, including the "component", all live together in this file.

ambrauer (changes requested):
I like it! Just the one comment regarding the component-factory template dependency. Also, looks like there may be some linting errors to resolve based on the CI checks.

samples/nextjs/scripts/templates/component-factory.ts
anastasiya29: good idea

ambrauer (approved):

PR #542, opened 2021-01-27
[Feature/459416] [sitecore-jss-cli] implemented .env handling for CLI

nickwesselman:
## Description
* primarily to support NODE_EXTRA_CA_CERTS
* with behavior and capabilities close to those of Next.js
* includes hidden `jss environment --name FOO` command for debugging

## Motivation
Primarily to support NODE_EXTRA_CA_CERTS. In the Next.js getting started template, I want to populate this in `.env.local` with the mkcert root.

## How Has This Been Tested?
* Unit tests
* Newly added (hidden) CLI command

## Types of changes

## Checklist:

- [X] My code/comments/docs fully adhere to the Code of Conduct.
- [X] My change is a code change and it requires an update to the documentation.

### DIFF
--- a/packages/sitecore-jss-cli/src/process-env.ts
+++ b/packages/sitecore-jss-cli/src/process-env.ts
@@ -1,0 +1,26 @@
+import path from 'path';
+import * as dotenv from 'dotenv';
+import dotenvExpand from 'dotenv-expand';
+
+/**
+ * @param {string} dir The directory containing the .env files to load.
+ */
+export default function processEnv(dir: string) {
+  // replicate Next.js handling/behavior
+  // https://github.com/vercel/next.js/blob/v10.0.5/packages/next-env/index.ts#L80-L90
+  const mode = process.env.NODE_ENV;
+  const dotenvFiles = [
+    `.env.${mode}.local`,
+    // Don't include `.env.local` for `test` environment
+    // since normally you expect tests to produce the same
+    // results for everyone
+    mode !== 'test' && '.env.local',
+    `.env.${mode}`,
+    '.env',
+  ].filter(Boolean) as string[];
+
+  // inspired by https://github.com/entropitor/dotenv-cli/blob/v4.0.0/cli.js#L53-L55
+  dotenvFiles.forEach(function(env) {
+    dotenvExpand(dotenv.config({ path: path.resolve(dir, env) }));
+  });
+}
--- a/packages/sitecore-jss-cli/src/process-env.ts
+++ b/packages/sitecore-jss-cli/src/process-env.ts
@@ -6,16 +6,16 @@ import dotenvExpand from 'dotenv-expand';
  * @param {string} dir The directory containing the .env files to load.
  */
 export default function processEnv(dir: string) {
-  // replicate Next.js handling/behavior
+  // replicate Next.js handling/behavior, but without a default NODE_ENV
   // https://github.com/vercel/next.js/blob/v10.0.5/packages/next-env/index.ts#L80-L90
-  const mode = process.env.NODE_ENV ?? 'development';
+  const mode = process.env.NODE_ENV;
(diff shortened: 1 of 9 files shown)


packages/sitecore-jss-cli/src/process-env.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * @param {string} dir The directory containing the .env files to load.
+ */
+export default function processEnv(dir: string) {
+  // replicate Next.js handling/behavior
+  // https://github.com/vercel/next.js/blob/v10.0.5/packages/next-env/index.ts#L80-L90
+  const mode = process.env.NODE_ENV;
+  const dotenvFiles = [
+    `.env.${mode}.local`,
```
anastasiya29, line 13: should this one also be excluded for test mode?

packages/sitecore-jss-cli/src/process-env.ts
```diff
@@ -0,0 +1,26 @@
...
+import * as dotenv from 'dotenv';
+import dotenvExpand from 'dotenv-expand';
+
+/**
+ * @param {string} dir The directory containing the .env files to load.
+ */
+export default function processEnv(dir: string) {
+  // replicate Next.js handling/behavior
+  // https://github.com/vercel/next.js/blob/v10.0.5/packages/next-env/index.ts#L80-L90
+  const mode = process.env.NODE_ENV;
```
anastasiya29, line 11: Is the NODE_ENV prop always guaranteed to have a value or do we need to check that before interpolating it?

packages/sitecore-jss-cli/src/process-env.ts
```diff
@@ -0,0 +1,26 @@
...
+  const dotenvFiles = [
+    `.env.${mode}.local`,
+    // Don't include `.env.local` for `test` environment
+    // since normally you expect tests to produce the same
+    // results for everyone
+    mode !== 'test' && '.env.local',
+    `.env.${mode}`,
+    '.env',
+  ].filter(Boolean) as string[];
```
anastasiya29, line 20: I know this code snippet is from the nextjs repo, but I'm wondering to myself why they took the extra step of casting to a string[]. Running the original string array through `Boolean` should just remove empty strings from the array, not change the type. Do you see what edge cases this is covering?

anastasiya29 (approved):

packages/sitecore-jss-cli/src/process-env.ts
nickwesselman: Next.js doesn't, I took that code pretty much verbatim from them. Not sure the reasoning, but my main intention was to have similar behavior.

packages/sitecore-jss-cli/src/process-env.ts
nickwesselman: Good call will fix

packages/sitecore-jss-cli/src/process-env.ts
nickwesselman: I assumed it was for this:

`mode !== 'test' && '.env.local'`

I'm not a typescript expert but I assume that the `Boolean` filter would remove the `false` when `mode == 'test'`, and the cast is needed because otherwise it's a `string | Boolean` array?

packages/sitecore-jss-cli/src/process-env.ts
nickwesselman: done

ambrauer (approved):
Looks great, nice addition to CLI

packages/sitecore-jss-cli/src/process-env.ts
nickwesselman: Based on follow up discussion, now excluding mode .env's if there is not a NODE_ENV.

anastasiya29 (approved):

PR #547, opened 2021-02-02
[NextJs] [EE] Root server URL detection/handling for internal editing API calls (take 2)

ambrauer:
Root server URL detection/handling for internal editing API calls. Server URL is determined based on original `/api/editing/render` request host header, and presence of `VERCEL` for https protocol usage (http otherwise). This is then used for all subsequent page and data API calls.

## Motivation
The previous 2 attempts (using PUBLIC_URL and then relative requests) failed to provide stability across all necessary environments, configurations, and platforms by default. This aims to cover our "planned" scenarios, with the ability to override if necessary.

## Types of changes

## Checklist:

### DIFF
--- a/packages/sitecore-jss-nextjs/src/middleware/editing-render-middleware.ts
+++ b/packages/sitecore-jss-nextjs/src/middleware/editing-render-middleware.ts
@@ -32,9 +32,9 @@ export interface EditingRenderMiddlewareConfig {
   resolvePageUrl?: (serverUrl: string, itemPath: string) => string;
   /**
    * Function used to determine the root server URL. This is used for the route/page and subsequent data API requests.
-   * By default, the host header is used, unless the VERCEL_URL environment variable is present (for Vercel hosting).
+   * By default, the host header is used, with https protocol on Vercel (due to serverless function architecture) and http protocol elsewhere.
    * @param {NextApiRequest} req The current request.
-   * @default process.env.VERCEL_URL ? `https://${process.env.VERCEL_URL}` : `http://${req.headers.host}`;
+   * @default `${process.env.VERCEL ? 'https' : 'http'}://${req.headers.host}`;
    * @see resolvePageUrl
    */
   resolveServerUrl?: (req: NextApiRequest) => string;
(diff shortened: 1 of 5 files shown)


nickwesselman (approved):

packages/sitecore-jss-nextjs/src/middleware/editing-render-middleware.ts
```diff
@@ -21,13 +21,23 @@ export interface EditingRenderMiddlewareConfig {
...
+   * @default `${serverUrl}${itemPath}`
+   * @see resolveServerUrl
    */
-  resolvePageRoute?: (itemPath: string) => string;
+  resolvePageUrl?: (serverUrl: string, itemPath: string) => string;
+  /**
+   * Function used to determine the root server URL. This is used for the route/page and subsequent data API requests.
+   * By default, the host header is used, unless the VERCEL_URL environment variable is present (for Vercel hosting).
+   * @param {NextApiRequest} req The current request.
+   * @default process.env.VERCEL_URL ? `https://${process.env.VERCEL_URL}` : `http://${req.headers.host}`;
```
anastasiya29, line 37: Can we determine the protocol being used? Hardcoding it seems fishy.

packages/sitecore-jss-nextjs/src/middleware/editing-render-middleware.ts
```diff
@@ -107,14 +122,28 @@ export class EditingRenderMiddleware {
...
 
-  private defaultResolvePageRoute = (itemPath: string) => {
-    return itemPath;
+  private defaultResolvePageUrl = (serverUrl: string, itemPath: string) => {
+    return `${serverUrl}${itemPath}`;
+  };
+
+  private defaultResolveServerUrl = (req: NextApiRequest) => {
+    return process.env.VERCEL_URL
+      ? `https://${process.env.VERCEL_URL}`
```
anastasiya29, line 145: Same comment about hardcoding protocols applies here

anastasiya29 (commented):
VERCEL_URL is supposed to be the system variable that Vercel provides during, build right? So you are using this var to get the url of the Next.js server, is that right?

According to https://github.com/vercel/vercel/discussions/5230, VERCEL_URL is the unique deployment URL, not the domain. So will this still work when an app is promoted to production and using a custom domain?

I find `serverUrl` to be a bit of a confusing variable name, since there could be multiple servers at play (Nextjs server, Sitecore server). I recommend making this more specific, or at least adding more details to the comments of the functions that use it.

packages/sitecore-jss-nextjs/src/middleware/editing-render-middleware.ts
ambrauer: @anastasiya29 Agree, we needed some context/explanation here. Per our discussion, I've switched to using `VERCEL` environment variable to determine protocol and `req.headers.host` for consistency / clarity. Updated / added comments throughout to explain.

PR #566, opened 2021-02-12
Nextjs cleanup

anastasiya29:
Small refactoring improvements in Next.js sample
- remove unused 'sitecoreDistPath' variable
- add comments for SITECORE_API_KEY env variable
- exclude 'Layout' and 'NotFound' component from component factory generation

## How Has This Been Tested?
- local and Vercel build
- linting

## Types of changes

## Checklist:


samples/nextjs/.env
```diff
@@ -16,3 +16,10 @@ PUBLIC_URL=http://localhost:3000
 # value must match your server-side value (see \sitecore\config\JssNextWeb.config).
 # We recommend an alphanumeric value of at least 16 characters.
 JSS_EDITING_SECRET=
+
+# Your Sitecore API key is needed to build the app. Typically, the API key is
+# defined in `scjssconfig.json`. This file may not exist when building locally
+# (if you've never run `jss setup`), or when building in a higher environment
+# (since `scjssconfig.json` is ignored from source control). In this case, use
+# this environment variable to provide the value at build time.
+SITECORE_API_KEY=
```
ambrauer, line 25: @anastasiya29 We should also add `SITECORE_API_HOST` w/ comment, since that'd be necessary as well without the `scjssconfig.json`?

samples/nextjs/src/pages/404.tsx
```diff
@@ -1,4 +1,4 @@
-import NotFound from 'components/NotFound';
```
ambrauer, line 1: @anastasiya29 I feel like this would go against Next.js / React conventions (where all React components are under /components folder, for [example here](https://github.com/vercel/next.js/tree/canary/examples/blog-starter-typescript)) and those components would get lost along with all the other files at the /src root. 

Is there another way we could exclude these from componentFactory? Or, I guess, is there any harm having them there (for now)? We could always add a backlog item to improve this for next release.

ambrauer (changes requested):

samples/nextjs/src/pages/404.tsx
nickwesselman: Hmm, relevant:

https://sitecore.stackexchange.com/questions/27161/generated-componentfactory-js-doesnt-cater-for-named-exports/27610

anastasiya29: @ambrauer - I don't think this goes against conventions, as typically there are multiple folders containing components - shared components, util components, layout components. Ex. https://github.com/UnlyEd/next-right-now/tree/v2-mst-aptd-at-lcz-sty/src 

Given that we are using 'src/components' specifically for component factory generation, I think it makes sense to separate components that should be part of the factory and those that should not. 

The harm is leaving those components where they are is that it would interfere with someone trying to create new components/renderings called 'Layout' or 'NotFound'.

@nickwesselman - supporting named exports seems like a separate issue, and that would be considered a new feature imo

samples/nextjs/.env
anastasiya29: Done

ambrauer (approved):

samples/nextjs/src/pages/404.tsx
ambrauer: @anastasiya29 I'm good with this. For whatever reason I was thinking src/ was the sample root (with package.json, etc) :D

nickwesselman: It wasn't the named exports so much as that he had non-Sitecore components in that path which were bombing out b/c they were not Sitecore components.

anastasiya29: @nickwesselman - ah gotchya. 
I tried to capture the best practices related to separating "components that are also sitecore renderings" from "components that are just for JS" in guides; it's definitely worth revisiting - https://jss-docs-preview3.herokuapp.com/guides/thinking-in-jss/project-structure#component-granularity

PR #585, opened 2021-02-22
[doc] [#458739] refactored doc navigation

nickwesselman:
* removed antd
* implemented nav based on bootstrap
* moved nav state into a React context

I'm sure this could be further refactored / cleaned up, but it's temporary.


anastasiya29 (approved):
Only issue I found in testing is that we have some duplicate names in the nav (ie "Getting Started") which causes multiple menus to expand when clicked, but that shouldn't hold up this PR.

Also, can you please add docs/package-lock.json to the repo - I'm not sure why it's ignored

ambrauer:
@nickwesselman  Looks like I'm late to the party :) Only thing I noticed is some flickering of the up/down icons when expanding menus, and the click area for the menu items now has some dead space which is a bit of a usability degradation. Neither of which are terribly urgent.

nickwesselman:
@ambrauer Thanks, these are fixed (or at least improved) in #590

PR #590, opened 2021-02-22
[doc] some UX improvements

nickwesselman:
* ensure markdown sublists indent
* remove extra top/bottom padding on sublists
* expand clickable sidenav area to full sidenav width
* hide arrows while nav animates to avoid jumping


ambrauer (changes requested):
These fixes look good to me. One issue I noticed (not sure if it's this change or last): looks like we lost mobile rendering of the nav.

ambrauer (approved):

PR #618, opened 2021-03-08
[NextJs] Sample app configuration options for 'jss create'

ambrauer:
## Description
Provides additional options for Next.js sample app to use on `jss create` [via `jss-create.js`](https://jss.sitecore.com/docs/techniques/custom-create-templates).
*  `--fetchMode {REST|GraphQL}` : Specifies how Sitecore data (layout, dictionary) is fetched. Default is REST.
*  `--renderMode {SSG|SSR}` : Specifies the Next.js pre-rendering mode of the optional catch-all route. Default is SSG.

Files are renamed or deleted based on selections.

**NOTE**
* the `GraphQLDictionary/LayoutService` implementations are forthcoming in separate PRs, those will be added to `dictionary/layout-service-factory.graphql.ts` when available
* Documentation updates will be a separate PR

## Motivation
Since we're introducing GraphQL-based layout and dictionary fetch options, we need some way to allow selection of REST (current/default) vs GraphQL on create of a Next.js sample app. This same approach works well for SSG (default) vs SSR.

## How Has This Been Tested?
Locally, generating new Next.js apps with parameters. e.g.
```
jss create test-graphql-ssg nextjs --source C:\projects\jss\samples --fetchMode GraphQL --renderMode SSG
```

## Types of changes

## Checklist:


samples/nextjs/jss-create.js
```diff
@@ -17,5 +20,83 @@ module.exports = function createJssProject(argv, nextSteps) {
...
+ * @param {string} [fetchMode] {REST|GraphQL} Default is REST. 
+ */
+function setFetchMode(fetchMode) {
+  const defaultDsfFile = path.join(__dirname, 'src/lib/dictionary-service-factory.ts');
+  const graphqlDsfFile = path.join(__dirname, 'src/lib/dictionary-service-factory.graphql.ts');
+  const defaultLsfFile = path.join(__dirname, 'src/lib/layout-service-factory.ts');
+  const graphqlLsfFile = path.join(__dirname, 'src/lib/layout-service-factory.graphql.ts');
+  let mode = fetchMode ? fetchMode.toLowerCase() : 'rest';
+
+  if (mode !== 'rest' && mode !== 'graphql') {
```
anastasiya29, line 48: Perhaps a good place to use enums?

samples/nextjs/jss-create.js
```diff
@@ -17,5 +20,83 @@ module.exports = function createJssProject(argv, nextSteps) {
 
   applyNameToProject(__dirname, argv.name, argv.hostName, 'JssNextWeb');
 
+  if (!argv.fetchMode || !argv.renderMode) {
```
anastasiya29, line 23: Since we want to get away from all the "modes" in general, maybe these variables should be named based on what they change?
For example, if the default render mode is SSG, then the param could be a boolean 'useSSR'

anastasiya29 (changes requested):

samples/nextjs/jss-create.js
ambrauer: Right... the "modes" avoidance... I had considered a flag, but we lose some flexibility making these binary options. We've discussed adding a (3rd) hybrid render option in the future for example. 
We can rename these using something other than "mode", just can't think of a good replacement.

samples/nextjs/jss-create.js
```diff
@@ -17,5 +20,91 @@ module.exports = function createJssProject(argv, nextSteps) {
...
+/**
+ * Sets how Sitecore data (layout, dictionary) is fetched.
+ * @param {string} [fetchWith] {REST|GraphQL} Default is REST. 
+ */
+function setFetchWith(fetchWith) {
+  const defaultDsfFile = path.join(__dirname, 'src/lib/dictionary-service-factory.ts');
+  const graphqlDsfFile = path.join(__dirname, 'src/lib/dictionary-service-factory.graphql.ts');
+  const defaultLsfFile = path.join(__dirname, 'src/lib/layout-service-factory.ts');
+  const graphqlLsfFile = path.join(__dirname, 'src/lib/layout-service-factory.graphql.ts');
+  const FetchWith = {
```
anastasiya29, line 46: Any particular reason not to use actual enums? (https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) And I was thinking it would be exported in case people needed to do more customizations based on this value. But maybe I'm overthinking this whole scenario.

samples/nextjs/jss-create.js
```diff
@@ -17,5 +20,91 @@ module.exports = function createJssProject(argv, nextSteps) {
...
+    REST: 'rest',
+    GRAPHQL: 'graphql',
+  };
+  let value = fetchWith ? fetchWith.toLowerCase() : FetchWith.REST;
+
+  if (value !== FetchWith.REST && value !== FetchWith.GRAPHQL) {
```
anastasiya29, line 52: since you're using a switch below, would it be more concise to move this error handling to the "default" switch handler.
switch (value) {
    case FetchWith.REST:
     ...
    case FetchWith.GRAPHQL:
     ...
    default:
     ...error handling here...

anastasiya29 (approved):
Unrelated to your PR, but I noticed the create script is .js, not .ts - do you know why?
Also I'm assuming docs will follow in a separate PR.

ambrauer:
See other comment about why .js vs .ts. And correct, docs will be in separate PR.

samples/nextjs/jss-create.js
ambrauer: Possible, but then you'd be duplicating your fallback's logic (SSG in this case) in the default handler as well.

PR #629, opened 2021-03-18
Fix a memory leak in the SSR Proxy

C-Sharper:
## Description

If ws: true, the details of the Layout service call will be saved to memory, this object is never cleared. By setting ws: false (as the layout service call is not a websocket call this is the correct config) the details of the call will not be saved to memory

## Motivation

See before and after fix memory usage
image

## How Has This Been Tested?

Load testing on Azure, manual testing of the proxy. We have this fix in production.

## Types of changes

## Checklist:

- [X] My code/comments/docs fully adhere to the Code of Conduct.


anastasiya29:
Thank you for submitting this PR. We need to check our tests and try to reproduce the issue. I will update this PR with relevant info.

anastasiya29:
@C-Sharper 
Abby, we haven't been able to reproduce the memory peaks on our side. Can you please provide us with more information about your scenario? Sitecore/Headless Services/JSS versions, test configuration, whether the test included route changes (or focused purely focused on page load), any noteworthy complexities about the pages being tested (# of components, use of personalization/caching), etc. Anything that would help us reproduce the issue and confirm the fix.

PR #643, opened 2021-04-02
470753 Fix type errors related to Route | null type in the Next.js sample app

anastasiya29:
The type of `route` was recently changed from `RouteData` to `RouteData | null`

Even though we never tried to render <Layout> when route is null, the possibility of null made TypeScript complain. My workaround is to rely on [[...path]] to build the `context` object since it is aware of whether `route` is null or not


illiakovalenko (approved):
Nice 👍

PR #645, opened 2021-04-06
[JSS] [NextJs] Debug Logging

ambrauer:
## Description
Debug logging for SDKs (focused on Next.js) using the [debug](https://www.npmjs.com/package/debug) module.

## Types of changes

## Checklist:

### DIFF
--- a/docs/data/routes/docs/nextjs/troubleshooting/en.md
+++ b/docs/data/routes/docs/nextjs/troubleshooting/en.md
@@ -6,12 +6,23 @@ title: Documentation
 
 # Troubleshooting
 
+- [Debug logging](#debug-logging)
 - [General setup checklist](#general-setup-checklist)
 - [Server-side JavaScript errors](#server-side-javascript-errors)
 - [Errors regarding SSL certificates](#errors-regarding-ssl-certificates)
 - [Errors deploying a JSS app locally](#errors-deploying-a-jss-app-locally)
 - [Data-fetching issues](#data-fetching-issues)
 - [Errors in GraphiQL](#errors-in-graphiql)
+
+## Debug logging
+
+You can enable debug logging of the Sitecore JSS SDKs by setting the `DEBUG` environment variable. In your `.env` (or `.env.local` for local environments), add the following line:
+
+```
+DEBUG=sitecore-jss:*
+```
+
+This will output _all_ debug logs. However, you may wish to be more selective. Refer to [Debug logging](/docs/fundamentals/troubleshooting/debug-logging) for additional details.
 
 ## General setup checklist
 
--- a/docs/data/routes/docs/nextjs/troubleshooting/en.md
+++ b/docs/data/routes/docs/nextjs/troubleshooting/en.md
@@ -16,7 +16,7 @@ title: Documentation
 
 ## Debug logging
 
-You can enable debug logging of the Sitecore JSS SDKs by setting the `DEBUG` environment variable. In your `.env` (or `.env.local` for local environments), add the following line:
+You can enable debug logging of the Sitecore JSS npm packages by setting the `DEBUG` environment variable. In your `.env` (or `.env.local` for local environments), add the following line:
 
 ```
 DEBUG=sitecore-jss:*
--- a/packages/sitecore-jss/src/graphql-request-client.ts
+++ b/packages/sitecore-jss/src/graphql-request-client.ts
@@ -1,21 +1,34 @@
 import { GraphQLClient, ClientError } from 'graphql-request';
 import { DocumentNode } from 'graphql';
-import debug from './debug';
+import debuggers, { Debugger } from './debug';
+
+export type GraphQLRequestClientConfig = {
+  /**
+   * The API key to use for authentication. This will be added as an 'sc_apikey' header.
+   */
+  apiKey?: string;
+  /**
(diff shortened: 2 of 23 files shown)


packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -19,24 +27,20 @@ export class GraphQLRequestClient {
...
-          Variables: ${JSON.stringify(variables, null, 2)}
-          Error: ${JSON.stringify(error, null, 2)}
-        `)
-      );
-      throw error;
-    };
-
-    // todo: we should also print the query if we are in debug mode (Anastasiya, March 2021)
-    return await client.request(query, variables).catch(onError);
+    return new Promise((resolve, reject) => {
```
anastasiya29, line 30: It seems odd to create a new promise when we already have a promise returned from client.request. Why is this needed?

packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -1,14 +1,22 @@
...
    * Provides ability to execute graphql query using given `endpoint`
    * @param {string} endpoint The Graphql endpoint
    * @param {string} [apiKey] The API key to use for authentication. This will be added as an 'sc_apikey' header.
    */
-  constructor(private endpoint: string, private apiKey?: string) {}
+  constructor(private endpoint: string, apiKey?: string) {
+    if (apiKey) {
+      this.headers.sc_apikey = apiKey;
+    }
+    this.client = new GraphQLClient(endpoint, { headers: this.headers });
```
anastasiya29, line 18: More of a note... I remember asking why we're initializing a new GraphQLClient for every request and whether it's safe to keep reusing the same instance across multiple requests. The answer was that this was indeed intentional based on documentation, and investigating this further is mentioned in 469713.

packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -1,14 +1,22 @@
...
+  private client: GraphQLClient;
+  private headers: Record<string, string> = {};
+
   /**
    * Provides ability to execute graphql query using given `endpoint`
    * @param {string} endpoint The Graphql endpoint
    * @param {string} [apiKey] The API key to use for authentication. This will be added as an 'sc_apikey' header.
    */
-  constructor(private endpoint: string, private apiKey?: string) {}
+  constructor(private endpoint: string, apiKey?: string) {
```
anastasiya29, line 14: The api key needs to be exposed publicly?

anastasiya29 (changes requested):
This debug lib that you found is really cool and overall I think this is a great feature.

I have 2 suggestions.

1. Right now the 3rd part lib is added to both, sitecore-jss and sitecore-jss-nextjs. That means we need to maintain the dependency (and how we use it) in 2 places. I think it would be cleaner if we encapsulated the dependency in a thin wrapper in sitecore-jss.

There's already the start of a wrapper in `sitecore-jss/src/debug.ts`, so maybe this module could just be expanded?

2. It looks like most of the debug calls that are being added are related to an async api call, which makes perfect sense. Since we've already gone through the trouble of standardizing how we call our apis (using either DataFetcher or GraphQLClient), I wonder if the debug logging should be part of that standardization.

For example, graphql-sitemap-service calls debug right before calling fetch. And fetch calls GraphQLClient.request, which also calls debug. The number of calls to debug could be reduced if we just let GraphQLClient be responsible for logging.

docs/data/routes/docs/nextjs/troubleshooting/en.md
```diff
@@ -6,13 +6,24 @@ title: Documentation
...
 - [General setup checklist](#general-setup-checklist)
 - [Server-side JavaScript errors](#server-side-javascript-errors)
 - [Errors regarding SSL certificates](#errors-regarding-ssl-certificates)
 - [Errors deploying a JSS app locally](#errors-deploying-a-jss-app-locally)
 - [Data-fetching issues](#data-fetching-issues)
 - [Errors in GraphiQL](#errors-in-graphiql)
 
+## Debug logging
+
+You can enable debug logging of the Sitecore JSS SDKs by setting the `DEBUG` environment variable. In your `.env` (or `.env.local` for local environments), add the following line:
```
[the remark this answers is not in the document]
ambrauer: Fixed

packages/sitecore-jss/src/graphql-request-client.ts
ambrauer: It's no longer a property at all - instead used to populate the private `headers` property in the constructor body.

packages/sitecore-jss/src/graphql-request-client.ts
ambrauer: I think it still makes sense to create a new one here, but definitely a good question for users of this client.

packages/sitecore-jss/src/graphql-request-client.ts
ambrauer: This is necessary for us to be able to add our debug log hooks to the response (within `then`/`catch`).

ambrauer:
@anastasiya29 Both great suggestions, I've made updates to incorporate these. Please take another look.

anastasiya29 (approved):

PR #646, opened 2021-04-06
[Docs][468636] Redirect based on field value in headless ssr

ancaemcken:
## Types of changes

## Checklist:

- [x] My change is a documentation change and it requires an update to the navigation.


docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
```diff
@@ -0,0 +1,134 @@
...
+2. To apply the template to existing items, [edit or change the template that an item is based on](https://doc.sitecore.com/users/101/sitecore-experience-platform/en/edit-or-change-the-template-that-an-item-is-based-on.html).
+
+## Customize the `node-headless-ssr-proxy` sample to handle redirects
+
+> This walkthrough assumes you have followed the ["Headless SSR via `sitecore-jss-proxy`" documentation](/docs/techniques/ssr/headless-mode-ssr) to set up the headless server-side rendering of your application.
+
+1. In the  `/.env` file, define environment variables: 
+
+   ```
+   SITECORE_JSS_APP_NAME="JssReactWeb"
```
anastasiya29, line 28: The env values in this example are specific to my local environment. Also, using `dotenv` together with an `.env` file to specify env variables is one way of doing it, but not the only way. So I think we should define which variables are needed, and then specify that this is an example. Would it be more clear?

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
```diff
@@ -0,0 +1,134 @@
...
+
+      ```javascript
+      const graphQLEndpoint = process.env.SITECORE_GRAPHQL_ENDPOINT || serverBundle.graphQLEndpoint || `${apiHost}/api/${appName}`;
+      
+      const defaultLanguage = process.env.DEFAULT_LANGUAGE || serverBundle.defaultLanguage || 'en';
+      
+      const hostname = process.env.SITECORE_HOSTNAME || serverBundle.hostname || apiHost;
+      ```
+
+   3. For the `config` module, define a new member function `getRedirects`. This example uses a GraphQL query:
```
anastasiya29, line 60: Link to config module: https://github.com/Sitecore/jss/blob/master/samples/node-headless-ssr-proxy/config.js

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
```diff
@@ -0,0 +1,134 @@
...
+
+      ```javascript
+      getRedirects: () => {
+          return fetch(graphQLEndpoint, {
+            method: 'POST',
+            headers: { 'Content-Type': 'application/json' },
+            body: JSON.stringify({
+              query: `
+              query {
+                search(fieldsEqual: [{ name: "_templatename", value: "Redirect Route" }]) {
```
anastasiya29, line 70: It is obvious what we're doing here? Here is some additional info that could be helpful:

The "Redirect Route" string passed to the `_templatename` arg correlates to the name of the template that was created above. So, basically we are searching the content tree for instances of Redirect Route items.

Also, this query can be improved. For example, right now it's searching the entire content tree, but in a real application, the scope of the query should be limited to the current site by specifying the search root item. (This detail is probably outside the scope of this page, but I'll let you decide)

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
```diff
@@ -0,0 +1,134 @@
...
+                }
+      
+                console.log(JSON.stringify(redirects, null, 4));
+                return redirects;
+              }, {});
+            });
+        }
+      ```
+
+4. In `/index.js`, leverage the new `getRedirects` method:
```
anastasiya29, line 111: Link to index.js: https://github.com/Sitecore/jss/blob/master/samples/node-headless-ssr-proxy/index.js

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
```diff
@@ -0,0 +1,134 @@
...
+
+4. In `/index.js`, leverage the new `getRedirects` method: 
+
+   ```javascript
+   let redirectPromise;
+   
+   server.use('*', (req, res, next) => {
+     if (redirectPromise === undefined) {
+       // cache list of redirects in memory
+       redirectPromise = config.getRedirects();
```
anastasiya29, line 119: This line is caching the results of the query so that the query is executed only once. This means that the headless ssr proxy would need to be restarted in order to fetch new values (in case content authors made changes). This example could be improved by adding cache invalidation logic.

anastasiya29 (changes requested):
The Headless SSR Proxy can only be used by React/Vue/React (not Next). Should we specify?

anastasiya29:
@nickwesselman - check it out, this is how we are handling providing code examples for commonly asked questions. what do you think?

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
AncaIO: > So I think we should define which variables are needed, and then specify that this is an example. Would it be more clear?

Yes, definitely.

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
AncaIO: At this step, we assume they already have a local copy of the file (see note at line 23). Because of that, I wouldn't link to a web page, just tell them which file they are working on inside the directory they created while setting up headless server-side rendering.

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
AncaIO: It's a big chunk of code and I agree that it could use a bit more information - either as sentences or code comments. I will see what works best.

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
AncaIO: Same as above. They should be working in their own directory and therefore have a local copy of the file.

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
AncaIO: Great addition!

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
AncaIO: @anastasiya29 Will this do? 
```
 SITECORE_HOSTNAME="<YOUR_JSS_APP_HOSTNAME>"
   SITECORE_GRAPHQL_ENDPOINT="https://<YOUR_CM_HOSTNAME>/api/<YOUR_JSS_APP_NAME>?sc_apikey={<YOUR_API_KEY>}"
   DEFAULT_LANGUAGE="<YOUR_DEFAULT_LANGUAGE>"
```

AncaIO:
@anastasiya29  Please re-review. Note that I split the query declaration & assignment to make the code chunk more manageable in terms of adding some of the clarifications & extra information you provided. Let me know if I've broken things :D

docs/data/routes/docs/techniques/redirection/headless-ssr/en.md
anastasiya29: That's clear to me

anastasiya29 (approved):

nickwesselman:
See related internal comms on this, please hold on merging pending our meeting next week.

anastasiya29:
@nickwesselman - this is moving to a "samples" repo, right? Should I get that going or are you working on it?

anastasiya29:
We will move this to the examples repo.

PR #692, opened 2021-05-13
Feature/467450 2

anastasiya29:
Update GraphQL API per latest edge schema
- All queries need to receive a language (updated connected/integrated GQL components)
- IDs passed passed to queries no longer need to be lowercased, all formats are now accepted
- No longer need to specify sort order for search, as there is a default sort order on the backend

Improvements
- `getAppRootId` helper is now used by both, sitemap service and gql dictionary service
- Functionality for executing search query that was repeated in multiple places (pagination, validating params) has been pulled out into `SearchService`, which can be used by other services
- New interface for GraphQLClient so that services can chose to use a different graphQL library if they want (no longer coupled to graphql-request)
- Apply similar patter to CacheClient: enable services to chose a different cache library
- Make it easier to customize our services by changing access level of private methods to protected

Testing Details
Unit tests updated for API changes, and more tests added

Documentation
Docs for new services coming in another PR


packages/sitecore-jss/src/i18n/graphql-dictionary-service.ts
```diff
@@ -1,39 +1,39 @@
...
+    $templates: String!
     $pageSize: Int = 10
     $after: String
   ) {
     search(
       where: {
         AND: [
-          { name: "_path", value: $rootItemId, operator: CONTAINS }
-          { name: "_templates", value: $dictionaryEntryTemplateId }
+          { name: "_path", value: $rootItemId }
```
OlegJytnik, line 23: Please add explicit CONTAINS operator

packages/sitecore-jss/src/i18n/graphql-dictionary-service.ts
```diff
@@ -1,39 +1,39 @@
...
           { name: "_language", value: $language }
+          { name: "_templates", value: $templates }
```
OlegJytnik, line 25: Please add explicit CONTAINS operator

OlegJytnik (changes requested):

packages/sitecore-jss/src/i18n/graphql-dictionary-service.ts
anastasiya29: Done

packages/sitecore-jss/src/i18n/graphql-dictionary-service.ts
anastasiya29: Done

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -61,225 +37,116 @@ export interface GraphQLSitemapServiceConfig {
    * The API key to use for authentication.
    */
   apiKey: string;
```
ambrauer, line 39: Feel like there's an opportunity to create a base config that includes 'endpoint' and 'apiKey' as well. `GraphQLServiceConfig`?

packages/sitecore-jss/src/graphql/page-list-service.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * The schema of data returned in response to a page list query request
+ */
+export type PageListQueryResult = { url: { path: string } };
+
+/**
+ * Service that fetches page list data using Sitecore's GraphQL API.
+ * @augments SearchQueryService<PageListQueryResult>
+ */
+export class PageListService extends SearchQueryService<PageListQueryResult> {
```
ambrauer, line 47: This `PageListService` seems fairly specific to the single usage in sitemap service. Given that we need to customize the query in sitemap service (see other comment), not sure this abstraction is providing much?

packages/sitecore-jss/src/graphql/search-service.ts
```diff
@@ -0,0 +1,121 @@
...
+};
+
+/**
+ * Configuration options for service classes that extend @see SearchQueryService.
+ * This extends @see SearchQueryVariables because properties that can be passed to the search query
+ * as predicates should be configurable. 'language' is excluded because, normally, all properties
+ * except 'language' are consistent across languages so they are passed to constructors, and
+ * 'language' can vary so it is passed to methods.
+ */
+export interface SearchServiceConfig extends Omit<SearchQueryVariables, 'language'> {
```
ambrauer, line 63: The comment helps, but I still found this a little confusing (interaction between `SearchQueryVariables` and `SearchServiceConfig` and their usages).

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
ambrauer: And I suppose we've got the same with potential for `RestServiceConfig` (apiHost / apiKey / etc)

packages/sitecore-jss/src/i18n/graphql-dictionary-service.ts
```diff
@@ -158,53 +108,42 @@ export class GraphQLDictionaryService extends DictionaryServiceBase {
...
+  }
 
-    return results;
+  /**
+   * Gets a GraphQL client that can make requests to the API. Uses graphql-request as the default
+   * library for fetching graphql data (@see GraphQLRequestClient). Override this method if you
+   * want to use something else.
+   * @returns {GraphQLClient} implementation
+   */
+  protected getGraphQLClient(): GraphQLClient {
```
ambrauer, line 143: `GraphQLLayoutService` needs the new signature / implementation for `getGraphQLClient()` applied:
https://github.com/Sitecore/jss/blob/feature/467450-2/packages/sitecore-jss/src/layout/graphql-layout-service.ts#L72 

It's still private and returning `GraphQLRequestClient` instead of `GraphQLClient`

ambrauer (changes requested):
See comments. Overall huge improvements.

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
anastasiya29: > Feel like there's an opportunity to create a base config that includes 'endpoint' and 'apiKey' as well. `GraphQLServiceConfig`?

I wanted to do this, but the issue is that GraphQLRequestClient's signature wants to receive endpoint and apikey separately, so it didnt seem right to define a { apiKey, endpoint } config in the client. But I'm certainly open to consider something like that. Let's save for another PR tho.

packages/sitecore-jss/src/graphql/page-list-service.ts
anastasiya29: What about generating sitemap (like, the sitemap that crawlers use). Or navigation.

OlegJytnik (approved):

packages/sitecore-jss/src/graphql/search-service.ts
```diff
@@ -0,0 +1,121 @@
...
+   * 2. Executes search query with pagination
+   * 3. Aggregates pagination results into a single result-set.
+   * @template T The type of objects being requested.
+   * @param {SearchQueryVariables} query the search query.
+   * @param {SearchQueryVariables} args search query arguments.
+   * @returns {T[]} array of result objects.
+   * @throws {RangeError} if a valid root item ID is not provided.
+   * @throws {RangeError} if the provided language(s) is(are) not valid.
+   */
+  async fetch(query: string, args: SearchQueryVariables): Promise<T[]> {
```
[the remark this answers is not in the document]
anastasiya29: fixed

packages/sitecore-jss/src/graphql/page-list-service.ts
anastasiya29: As agreed upon, the extra PageListService abstraction has been removed

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -61,225 +37,116 @@ export interface GraphQLSitemapServiceConfig {
...
+   * @param {string[]} languages Fetch pages that have versions in this language(s).
+   * @param {Function} formatStaticPath Function for transforming the raw search results into (@see StaticPath) types.
+   * @returns list of pages
+   * @throws {RangeError} if the list of languages is empty.
+   * @throws {Error} if the app root was not found for the specified site and language.
    */
   async fetchSitemap(
     languages: string[],
-    rootItemPath: string,
     formatStaticPath: (path: string[], language: string) => StaticPath
```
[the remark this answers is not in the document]
anastasiya29: Per our discussion, our decision is to add support for 'excludes' paths since this is a common use cases. For other customization, the query needs to be overwritten by extending the service class.

packages/sitecore-jss/src/i18n/graphql-dictionary-service.ts
anastasiya29: Done. Good catch.

ambrauer (approved):

PR #699, opened 2021-05-21
abort controller

maximvsidorenko:
## Description / Motivation
Timeout for GraphQL request client by AbortController. 

## Types of changes

### DIFF
--- a/packages/sitecore-jss/src/graphql-request-client.ts
+++ b/packages/sitecore-jss/src/graphql-request-client.ts
@@ -1,6 +1,8 @@
 import { GraphQLClient, ClientError } from 'graphql-request';
+import { RequestInit } from 'graphql-request/dist/types.dom';
 import { DocumentNode } from 'graphql';
 import debuggers, { Debugger } from './debug';
+import AbortController from "abort-controller";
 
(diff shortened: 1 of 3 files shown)


packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -46,6 +46,14 @@ export class GraphQLRequestClient {
     query: string | DocumentNode,
     variables?: { [key: string]: unknown }
   ): Promise<T> {
+    const abortController = new AbortController();
```
SergeyZubatkin, line 49: Do we need to create and set abort controller if timeout not specified?

alinabilaya: What about having a reasonable default timeout (if nothing is specified)? Or just providing by default timeouts set for both endpoints?

maximvsidorenko: TBD with @anastasiya29 and the team: seems the current sample application does not support the concept of the default timeout for requests. Adding it will result in to numerous changes in other APIs, so it need to be planned appropriately,

PR #722, opened 2021-06-11
Update Next.js sample config - layoutServiceConfiguration="default"

anastasiya29:
layoutServiceConfiguration should be set to "default" when using Delivery Edge for API endpoints. When using integrated GraphQL with Delivery Edge, a $language value is injected since language is required in all Edge queries. XP does not do this (which is backwards compatible with JSS versions < 18.0.0). Since GraphQL does not play nice with passing inunused parameters to queries, we need to tell Sitecore type of GraphqL endpoint we are using.

### DIFF
--- a/samples/nextjs/sitecore/config/JssNextWeb.config
+++ b/samples/nextjs/sitecore/config/JssNextWeb.config
@@ -81,8 +81,15 @@
           Defaults are defined in `/App_Config/Sitecore/JavaScriptServices/Sitecore.JavaScriptServices.Apps.config`
 
           NOTE: graphQLEndpoint enables _Integrated GraphQL_. If not using integrated GraphQL, it can be removed.
+
+          NOTE: layoutServiceConfiguration should be set to "default" when using Delivery Edge for
+          API endpoints. When using integrated GraphQL with Delivery Edge, a $language value is injected
+          since language is required in all Edge queries. XP does not do this (which is backwards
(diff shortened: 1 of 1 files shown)


samples/nextjs/sitecore/config/JssNextWeb.config
```diff
@@ -81,8 +81,15 @@
           Defaults are defined in `/App_Config/Sitecore/JavaScriptServices/Sitecore.JavaScriptServices.Apps.config`
 
           NOTE: graphQLEndpoint enables _Integrated GraphQL_. If not using integrated GraphQL, it can be removed.
+
+          NOTE: layoutServiceConfiguration should be set to "default" when using Delivery Edge for
+          API endpoints. When using integrated GraphQL with Delivery Edge, a $language value is injected
+          since language is required in all Edge queries. XP does not do this (which is backwards
+          compatible with JSS versions < 18.0.0). Since GraphQL does not play nice with passing in
+          unused parameters to queries, we need to tell Sitecore type of GraphqL endpoint we are using.
```
ambrauer, line 89:
```suggestion
          NOTE: layoutServiceConfiguration should be set to "default" when using GraphQL Edge schema.
          When using integrated GraphQL with Edge schema, a $language value is injected
          since language is required in all Edge queries. "jss" configuration does not do this (which is backwards
          compatible with JSS versions < 18.0.0).
```

ambrauer (approved):
Looks good, just some suggestions in the comment

PR #728, opened 2021-06-21
[Docs] Update cli docs 479957

ancaemcken:
## Types of changes

### DIFF
--- a/docs/data/routes/docs/fundamentals/cli/en.md
+++ b/docs/data/routes/docs/fundamentals/cli/en.md
@@ -5,28 +5,54 @@ title: JSS CLI
 ---
 # JSS CLI
 
+## JSS commands
+
 During JSS development, you will be commonly using scripts provided by the _JSS CLI_ (command line interface). These scripts can be run from within a JSS application at your command line of choice (bash, PowerShell, cmd, etc).
 
-## Frequently used scripts
+All scripts you need to run during development are defined either via the `jss` CLI command (use `--help` for options) or as package scripts in `package.json`.
 
-These are the main scripts you will be using frequently during JSS development.
-These were partially covered in [Quick start](/docs/client-frameworks/getting-started/quick-start).
+Some commands have application specific arguments. Run `jss --help` and `jss <command> --help` for application specific arguments.
+
+### Quick links
+
+- [`jss build`](#jss-build)
+- [`jss clean`](#jss-clean)
+- [`jss create`](#jss-create)
+- [`jss deploy app`](#jss-deploy-app)
+- [`jss deploy component`](#jss-deploy-component)
+- [`jss deploy config`](#jss-deploy-config)
+- [`jss deploy app`](#jss-deploy-app)
+- [`jss build`](#jss-build)
+- [`jss build`](#jss-build)
+
+### `jss build`
+
+Creates a production build for deploying to Sitecore. Result - a JS bundle and other static assets under `/dist` folder.
+
+### `jss clean`
+
+Cleans the contents of a directory.
+
+|Parameter |Description| Value type|
+| --- | --- | --- |
+|`path`| The path to clean. Defaults to `buildArtifactsPath` config in `package.json`.|`String`|
 
 ### `jss create`
-Creates a JSS application. You must provide an application name and a framework. 
 
-```
+Creates a JSS application. You must provide an application name and a framework.
+
+```bash
 jss create <your-app-name> <app-template-name>
 ```
 
-**Required parameters**
+#### Required parameters
 
-|Parameter |Description| Value type| 
-| --- | --- | --- | 
+|Parameter |Description| Value type|
+| --- | --- | --- |
 |name| The name of the app to create.|`String`|
 |template| The template to create the app from; corresponds to folders in the [Official JSS repository](https://github.com/Sitecore/jss/tree/master/samples)|`String`|
 
-**Optional parameters**
+#### Optional parameters
 
 |Parameter| Description| Options | Default |Template|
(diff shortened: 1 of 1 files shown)


docs/data/routes/docs/fundamentals/cli/en.md
```diff
@@ -5,28 +5,85 @@ title: JSS CLI
...
+# becomes 'adminb'
+EXPANDED=admin$PASSWORD
+
+# becomes 'admin$PASSWORD'
+ESCAPED=admin\$PASSWORD
+```
+
+## JSS commands
+
+During JSS development, you will commonly use scripts provided by the _JSS CLI_ (command-line interface). All scripts you need to run during development are defined either via the `jss` CLI command (use `--help` for options) or package scripts in `package.json`.
```
anastasiya29, line 33: We have commands defined in the sitecore-jss-cli base package (create, setup, deploy), and then there are client-controlled commands like `build` and `scaffold`, which can look like jss scripts because `jss` in the command line acts as a wrapper for `npm run`, but they are totally up to the client.

In this section, these 2 types of scripts are being discussed together, and I'm worried that this may be misleading. I think we should document commands from the base package separately, and then have a section for commands defined in the examples. The tricky part is that *some* commands defined in the samples differ by framework... it might be best to save this update for the new docs? Then each framework could have it's own CLI reference and we could utilize that doc snippet functionality to repeat shared docs.

docs/data/routes/docs/fundamentals/cli/en.md
```diff
@@ -52,68 +110,130 @@ jss create my-jss-app nextjs --prerender SSG --fetchWith GraphQL
 jss create my-jss-app nextjs --prerender SSR --fetchWith REST
```
anastasiya29, line 110: Is it obvious that `prerender` should not be used with RAV frameworks?

docs/data/routes/docs/fundamentals/cli/en.md
```diff
@@ -52,68 +110,130 @@ jss create my-jss-app nextjs --prerender SSG --fetchWith GraphQL
...
 ```
 
-### `jss setup`
+### `jss deploy app`
```
anastasiya29, line 113: Not sure if we have this captured anywhere else - `deploy app/files` are *not* valid for Next.js (because Next.js does not support Integrated mode, so we never need to deploy the JS files to Sitecore).

docs/data/routes/docs/fundamentals/cli/en.md
```diff
@@ -52,68 +110,130 @@ jss create my-jss-app nextjs --prerender SSG --fetchWith GraphQL
...
+> `jss deploy template` has options that enable modifying how the template is scaffolded, such as specifying template fields with `--fields` or setting a display name with `--displayName`. Run `jss deploy template --help` to see all the options with descriptions.
 
-> `jss deploy component` has many options that enable modifying how the component is scaffolded, such as specifying datasource template fields with `--fields` or setting a display name with `--displayName`. Run `jss deploy component --help` to see all the options with descriptions.
+### `jss deploy`
 
-## Other scripts
+Deploys the JSS app artifacts to a Sitecore instance. Use `jss deploy --help` for subcommands.
 
-All scripts you need to run during development are defined either via the `jss` CLI command (use `--help` for options) or as package scripts in `package.json`.
+### `jss environment`
```
anastasiya29, line 210: Huh. I didn't know this was a thing. Does it take into account overrides to scjssconfig coming from .env file? It would probably be helpful to add this info to the doc.

anastasiya29 (changes requested):

docs/data/routes/docs/fundamentals/cli/en.md
```diff
@@ -5,28 +5,85 @@ title: JSS CLI
...
-## Frequently used scripts
+The CLI supports loading environment variables from `.env` files. Use a `.env` file to populate environment variables such as `NODE_EXTRA_CA_CERTS` and `NODE_TLS_REJECT_UNAUTHORIZED`, which you may have already configured in a `.env` file for communication with your Sitecore instance.
 
-These are the main scripts you will be using frequently during JSS development.
-These were partially covered in [Quick start](/docs/client-frameworks/getting-started/quick-start).
+Variables will be loaded from `.env` files with the following priority:
 
-### `jss create`
-Creates a JSS application. You must provide an application name and a framework. 
+* `.env.${NODE_ENV}.local`
```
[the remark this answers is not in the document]
ancaemcken: Ah good to know.

docs/data/routes/docs/fundamentals/cli/en.md
ancaemcken: In the table it is marked as Next.js.

docs/data/routes/docs/fundamentals/cli/en.md
ancaemcken: I think you are right about saving the update for the new docs. We won't have time to add and verify all framework-specific differences on env, and client commands.

docs/data/routes/docs/fundamentals/cli/en.md
ancaemcken: And we come again to the framework-specific docs.

docs/data/routes/docs/fundamentals/cli/en.md
ancaemcken: I'll test this and see.

ancaemcken:
CLI docs will be re-visited for 19.0 as part of the documentation reorg. Closing this PR.

PR #736, opened 2021-06-28
[Next.js] Prevent extraneous router.replace in Experience Editor when using SSG

ambrauer:
Prevent extraneous router.replace (and resulting 404 in certain route configurations) in Experience Editor when using SSG.

## Description / Motivation
When SSG, Next will [attempt to perform a router.replace on the client-side](https://github.com/vercel/next.js/blob/v10.0.3/packages/next/client/index.tsx#L169) to inject the query string parms to the router state. However, this doesn't really work since at this point we're in the editor and the location.search has nothing to do with the Next route/page we've rendered. Beyond the extraneous request, this can result in a 404 with certain route configurations (e.g. multiple catch-all routes).

This fix swaps out injected SSG property w/ SSR to trick it into thinking we're SSR, thus avoiding any router.replace.

## Testing Details
Tested locally in Experience Editor and Horizon.

## Types of changes

### DIFF
--- a/docs/data/routes/docs/nextjs/graphql/sample-app/en.md
+++ b/docs/data/routes/docs/nextjs/graphql/sample-app/en.md
@@ -4,3 +4,78 @@ routeTemplate: ./data/component-templates/article.yml
 title: Sitecore GraphQL in the sample app
 ---
 # Sitecore GraphQL in the sample app
+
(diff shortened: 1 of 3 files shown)


illiakovalenko (approved):
Great work!

PR #784, opened 2021-08-17
493396 [maintenance] [react] Remove obsolete poc code from sample

anastasiya29:
Anything referencing "holygrail" was part of a POC demo that was built for a 2019 SUGCON presentation. This was productized as "rendering host", so that old code should be deleted


CobyPear (approved):
Delete all the code!

PR #800, opened 2021-09-03
[Nextjs] Use the app name as the prefix value for templates

CobyPear:
When developing locally, multiple templates with the same name will conflict causing errors when using strongly typed GraphQL queries. 
This feature adds a prefix to some files so that when `jss create` is run, the prefix is replaced with the appname, which will give each template a unique name in Sitecore when using `jss deploy`
## Description / Motivation

- Added a prefix to pertinent files and filenames in the Nextjs sample app
- Added a function to the create script in `sitecore-jss-cli` that replaces the prefix with the appname

Faster internal development

## Testing Details

- GraphQL components and queries require Capitalization format. A function was written to preform this  task and a unit test was added.
- Used jss create to scaffold a new Nextjs app and deployed it to Sitecore along with the sample app - no conflicting template names = no errors! 🥳

## Types of changes


ambrauer (changes requested):

packages/sitecore-jss-cli/src/create/index.ts
```diff
@@ -105,4 +105,77 @@ export function applyNameToProject(
...
+}
+
+/**
+ * Called during jss create, this function replaces the sample's prefix with the app's name on Sitecore templates
+ * @param {string} projectFolder Project folder
+ * @param {string} name Name value to replace
+ * @param {string} prefix Prefix of the sample app's template - should match Jss[RAV|Next]Web
+ */
+export function replacePrefix(projectFolder: string, name: string, prefix: string) {
+  glob.sync(path.join(projectFolder, '**/*.*')).forEach((filePath: string) => {
```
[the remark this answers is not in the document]
CobyPear: Yes this makes sense, I couldn't tell if a glob like that would work. Will update this.

packages/sitecore-jss-cli/src/create/index.ts
```diff
@@ -105,4 +105,77 @@ export function applyNameToProject(
...
+        fileContents = applyNameReplacement(fileContents, prefix, name);
+        fs.writeFileSync(filePath, fileContents);
+        const newPath: string = applyNameReplacement(filePath, prefix, name);
+        fs.renameSync(filePath, newPath);
+        return;
+      }
+
+      fileContents = applyNameReplacement(fileContents, prefix, name);
+      fs.writeFileSync(filePath, fileContents);
+    }
```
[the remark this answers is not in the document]
CobyPear: In that case we should probably just ditch the lowercase stuff all together and make it all Pascal case?

packages/sitecore-jss-cli/src/create/index.ts
```diff
@@ -105,4 +105,77 @@ export function applyNameToProject(
...
+        if (filePath.match(/GraphQl/)) {
+          fileContents = applyNameReplacement(fileContents, prefix, getGqlFormattedName(name));
+          fs.writeFileSync(filePath, fileContents);
+        }
+        return;
+      }
+
+      if (filePath.includes(`${prefix}-`)) {
+        fileContents = applyNameReplacement(fileContents, prefix, name);
+        fs.writeFileSync(filePath, fileContents);
```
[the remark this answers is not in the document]
CobyPear: I thought so too, but for some reason without returning here I was getting 2 copies of the file, one with a new file name but wrong contents, and the old file remained but with correct contents.

packages/sitecore-jss-cli/src/create/index.ts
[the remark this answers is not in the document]
ambrauer: Totally, we probably could. I don't think there are any requirements to be all lowercase in those cases (manifest 'id').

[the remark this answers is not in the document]
CobyPear: This made things a lot cleaner 😊

packages/sitecore-jss-cli/src/create/index.ts
```diff
@@ -105,4 +105,40 @@ export function applyNameToProject(
...
+}
+
+/**
+ * Called during jss create, this function replaces the sample's prefix with the app's name on Sitecore templates
+ * @param {string} projectFolder Project folder
+ * @param {string} name Name value to replace
+ * @param {string} prefix Prefix of the sample app's template - should match Jss[RAV|Next]Web
+ */
+export function replacePrefix(projectFolder: string, name: string, prefix: string) {
+  console.log(chalk.cyan(`Replacing template prefix with ${getPascalCaseName(name)}`));
```
ambrauer, line 130: Add '...' for consistency with other logs
```suggestion
  console.log(chalk.cyan(`Replacing template prefix with ${getPascalCaseName(name)}...`));
```

ambrauer (commented):
This is working brilliantly. Not sure if you had this planned for "phase 2" (apply to RAV), but we discussed adding a flag to `jss create` to make the prefix opt-in (for backwards compatibility). I think this is still valuable, and I don't think it'd be that much extra work?

packages/sitecore-jss-cli/src/create/index.ts
CobyPear: > 
> 
> This is working brilliantly. Not sure if you had this planned for "phase 2" (apply to RAV), but we discussed adding a flag to `jss create` to make the prefix opt-in (for backwards compatibility). I think this is still valuable, and I don't think it'd be that much extra work?

Yeah I'll try to add that functionality here. Shouldn't be too bad!

CobyPear: https://github.com/Sitecore/jss/pull/800/commits/8bd96971e4647ba33069cbfe5df907c30d916890 should do the trick.

packages/sitecore-jss-cli/src/create/index.ts
```diff
@@ -65,12 +65,14 @@ export function applyHostNameToSitecoreConfig(configXml: string, hostName: strin
...
  * @param {string} replaceName Name value to replace
+ * @param {string} withPrefix Option to pass to replacePrefix - defaults to false
  */
 export function applyNameToProject(
   projectFolder: string,
   name: string,
   hostName: string,
-  replaceName: string
+  replaceName: string,
+  withPrefix?: string
```
ambrauer, line 75: Should this be a boolean?

packages/sitecore-jss-cli/src/create/index.ts
```diff
@@ -125,9 +127,41 @@ export function getPascalCaseName(name: string): string {
...
+ * @param {string} withPrefix if true, replaces prefix with app name in PascalCase,
+ * otherwise strip the prefix
  */
-export function replacePrefix(projectFolder: string, name: string, prefix: string) {
-  console.log(chalk.cyan(`Replacing template prefix with ${getPascalCaseName(name)}`));
+export function replacePrefix(
+  projectFolder: string,
+  name: string,
+  prefix: string,
+  withPrefix?: string
```
ambrauer, line 137: Same here - boolean?

packages/sitecore-jss-cli/src/create/index.ts
```diff
@@ -125,9 +127,41 @@ export function getPascalCaseName(name: string): string {
...
+) {
+  const value: boolean = withPrefix?.toLowerCase() === 'true' ? true : false;
+
+  if (!value) {
+    console.log(chalk.cyan('Removing template prefix...'));
+    name = '';
```
ambrauer, line 143: I'd avoid repurposing the `name` param in this case (usually bad practice to modify incoming parameters), and just use empty string directly below.

ambrauer (changes requested):
Just a couple minor things, otherwise looks good!

packages/sitecore-jss-cli/src/create/index.ts
CobyPear: I had it as a boolean originally but it was still coming in as a string?

ambrauer (approved):

PR #810, opened 2021-09-10
[sitecore-jss-manifest][Nextjs] Allow component manifest definitions to define a template name

ambrauer:
## Description / Motivation
When deploying manifest items to Sitecore (via `jss deploy`), the resulting component and datasource template use the same name (implicit dependency between them). This update adds an optional `templateName` which can be used for component definitions, breaking that dependency. This value will be used when creating the datasource template item in Sitecore (instead of the component `name`).

This allows us to prefix templates **only** (separately from components).

Also included:

- Updates to the Next.js sample app with this new feature in mind (revert component names)
- Removed filename processing from `jss create` prefix logic (since component files are no longer prefixed)

## Types of changes


CobyPear (approved):
Looks like there was some extra work to do inside of generateItems etc, but why did it seem to work yesterday?
Looks good other than that question.

ambrauer:
@CobyPear We only made it through the renderings, I ran into additional errors when it got to the shared content examples.

PR #815, opened 2021-09-16
[Angular] *scRouterLink breaks link generation

CobyPear:
Currently with SXA, this Angular component will add querystring params to the end of the `href` of the link, for example `/test` becomes `/test?sc_site=this-sitename`. This PR adds a tested function that strips the querystring params off of the intended path, resulting in correct client side routing.
Limitations of this component include being unable to access an external link, for such scenarios please use *scLink
## Description / Motivation

Hotfix request

## Testing Details

Tested on SC 10.2 with JSS 20 and SC 10.0 with JSS 16 + SXA

## Types of changes


packages/sitecore-jss-angular/src/components/router-link.directive.spec.ts
```diff
@@ -164,6 +164,19 @@ describe('<a *scRouterLink />', () => {
...
+    const field = {
+      href: '/lorem?sc_site=lorem_ipsum',
+      text: 'ipsum',
+    };
+    comp.field = field;
+    field.href = stripQueryParams(field.href);
+    fixture.detectChanges();
+
+    const rendered = de.query(By.css('a'));
+    expect(rendered.nativeElement.href).toBe('http://localhost:9876/lorem');
```
ambrauer, line 178: Why is this an absolute URL with 'http://localhost:9876'?

Also, I was just looking through the tests we've got, and there really aren't any assertions on the href attribute ... Wonder if we should at least one for a normal case (without query parm)? The "should render all value attributes" test seems like a good candidate.

ambrauer (approved):
Approved with a question / suggestion on tests.

packages/sitecore-jss-angular/src/components/router-link.directive.spec.ts
CobyPear: > Why is this an absolute URL with 'http://localhost:9876'?

That's what was coming back from the test when it was failing, I thought it was strange too.

Yeah that seems worth adding.

ambrauer (approved):

PR #820, opened 2021-09-26
One valtech

victora2:
## Description / Motivation

Replace default routes/pages
so that the template generates just one Company page

Replace logo.


victora2:
should not deploy to azure

PR #821, opened 2021-09-28
[Focused improvement] Add a linting gate for git push actions

CobyPear:
Add a script that 'installs' a pre-commit hook which will lint packages and apps on ~~commit~~ push. If you'd like to skip this hook, add the `--no-verify` flag to your ~~commit~~ push i.e. `git push origin head --no-verify`
## Description / Motivation

Instead of waiting for CI to potentially fail due to unlinted apps/packages, we want to catch those errors locally.

## Types of changes

### DIFF
--- a/scripts/installHooks.js
+++ b/scripts/installHooks.js
@@ -21,7 +21,7 @@ npm run lint-apps;`;
 
   // Write the hook to the local .git folder. Using writeFile in order to catch any errors
   /* eslint-disable no-unused-vars */
-  fs.writeFile('./.git/hooks/pre-commit.local-linter', data, 'utf8', (err, _) => {
+  fs.writeFile('./.git/hooks/pre-commit', data, 'utf8', (err, _) => {
     if (err) {
       console.log('\x1b[31m%o\x1b[0m', err);
     }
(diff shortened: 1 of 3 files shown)


scripts/installHooks.js
```diff
@@ -0,0 +1,32 @@
+const fs = require('fs');
```
ambrauer, line 1: Use our newer file naming convention here "install-hooks.js"?

ambrauer (approved):
Just the one minor suggestion, otherwise this looks good to me. Certain git GUI may not easily facilitate the --no-verify flag, so I like that this is opt-in.

scripts/installHooks.js
CobyPear: Good call!

ambrauer (commented):
One other thought. Can/should we make this a pre-push hook instead of a pre-commit? Would still solve our main goal of preventing linting issues from making it to origin, but less obtrusively to dev workflow.

CobyPear:
> One other thought. Can/should we make this a pre-push hook instead of a pre-commit? Would still solve our main goal of preventing linting issues from making it to origin, but less obtrusively to dev workflow.

Yeah absolutely. Will change!

PR #824, opened 2021-10-04
[Focused Improvement] Consolidate packages

sc-addypathania:
Consolidate manifest and pipeline packages into dev-tools

## Description / Motivation

To be able to maintain all the packages in a easier way in future. 

## Types of changes


illiakovalenko (changes requested):
Well done!
I would propose you to introduce `path aliases` using `tsconfig`
Currently, we reference one package from another, and in `sitecore-jss-dev-tools` you can see that you added a lot of long paths like `../../../../index` etc.
https://www.typescriptlang.org/docs/handbook/module-resolution.html#path-mapping
https://dev.to/larswaechter/path-aliases-with-typescript-in-nodejs-4353

PR #848, opened 2021-11-03
Update contributing guidelines

CobyPear:
Contributing guidelines are outdated, update them to reflect current development workflow and stylistic guidelines.
## Types of changes

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,8 +73,14 @@ npm run test
 
 Problem: Build of package or sample fails with the following error: `cannot find module ...`
 
-Solution: If dependencies in a package change, you may need to reset the repo to have the latest changes. This is often needed after pulling changes that include a dependency change.
+Solution: If dependencies in a package change, you may need to install said dependencies.
 From the root of the monorepo:
+
+```shell
+npm run bootstrap
+```
+
+`npm run bootstrap` will solve the problem in most cases, however in nuclear scenarios you may run the following to completely reset the monorepo:
 
 ```shell
 npm run reset
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,46 +7,71 @@ Want to contribute to Sitecore JavaScript Services? There are a few things you n
 - `node.js` ([Active LTS](https://nodejs.org/en/about/releases/) version) installed (cmd `node -v` to test).
 - `npm` (6.x or later) installed (cmd `npm -v` to test).
 
+## Developing
+
+See Branching overview below - We use `dev` for our current development.
+
+1. [Fork](https://help.github.com/articles/fork-a-repo/) this repository to your own GitHub account and then [clone](https://help.github.com/articles/cloning-a-repository/) it to your local device.
+2. Create a new branch i.e. `git switch -c feature/my-jss-feature`
+3. When you're happy with your changes, open a Pull Request targeting the `dev` branch of the `sitecore/jss` repository. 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -7,46 +7,82 @@ Want to contribute to Sitecore JavaScript Services? There are a few things you n
...
-- From a terminal in the repo root: `cd docs`
-- Create a feature branch from `master` for your changes. e.g. `git checkout -b my-feature-branch`
-  > Note: although `dev` is the default branch for the JSS repo, documentation changes that pertain to the current release of JSS should be branched from `master`. The `master` branch contains the published / active documentation content. Changes related to the next release should be added to the `dev` branch.
-- `npm install`
-- `jss start` (to preview your changes locally)
-- Make your changes
-- Commit, push to your remote fork of the JSS repo, then open a pull request (PR) to the `master` branch of the JSS repo.
+* `master `- latest released version
+* `dev `- latest changes for the next release
+* `release `branches - created in order to make changes/updates to past major version releases
```
ambrauer, line 87: These are also used for sample sources (e.g. `jss create my-app nextjs --branch release/16.0.0`). Not sure if we need to include that necessarily, but maybe we can simplify to something like:
```suggestion
* `release `branches - all major released versions
```

ambrauer (approved):
Nice work! This was overdue... Just some minor suggestions here and there.

CONTRIBUTING.md
CobyPear: Good call, changed.

CONTRIBUTING.md
```diff
@@ -5,48 +5,84 @@ Want to contribute to Sitecore JavaScript Services? There are a few things you n
...
+- `npm` (6.x) installed (cmd `npm -v` to test).
+
+## Developing
+
+See Branching overview below - We use `dev` for our current development.
+
+1. [Fork](https://help.github.com/articles/fork-a-repo/) this repository to your own GitHub account and then [clone](https://help.github.com/articles/cloning-a-repository/) it to your local device.
+2. Create a new branch e.g. `git switch -c feature/my-jss-feature`
+3. When you're happy with your changes, open a Pull Request targeting the `dev` branch of the `sitecore/jss` repository. 
+4. Note: CI will run lint for all packages and samples, as well as tests for all packages. Please make sure these pass or your PR can not be merged.
```
sc-addypathania, line 17: We can add in another point that they must add the dev reviewers group to the reviewer's list after making changes to the PR.

CobyPear: I have this in there, maybe it's not pushed yet. Stand by :)

CONTRIBUTING.md
```diff
@@ -5,48 +5,84 @@ Want to contribute to Sitecore JavaScript Services? There are a few things you n
...
+
+```shell
+npm run test
+```
+
+# Troubleshooting
+
+Problem: Build of package or sample fails with the following error: `cannot find module ...`
+
+Solution: If dependencies in a package change, you may need to reset the repo to have the latest changes. This is often needed after pulling changes that include a dependency change.
```
sc-addypathania, line 76: instead of always running a reset, shouldn't they do a bootstrap ?
and do a reset only if they want to clean the cache in node_modules.
Just saying because Lerna reset is time-consuming.

CobyPear: True. My thought was that locally it would be rare to experience the 'cannot find module' error unless a reset was needed, but for us we see it all the time since we change branches often. I'll add that as the first option tho.

sc-addypathania (approved):

PR #853, opened 2021-11-09
506966 Fetch With using env variables

CobyPear:
Set fetchWith using env variables in preparation for nextjs Yeoman generator

## Types of changes


samples/nextjs/src/lib/dictionary-service-factory.ts
```diff
@@ -1,20 +1,44 @@
-import { DictionaryService, GraphQLDictionaryService } from '@sitecore-jss/sitecore-jss-nextjs';
+import {
+  DictionaryService,
+  RestDictionaryService,
+  GraphQLDictionaryService,
+} from '@sitecore-jss/sitecore-jss-nextjs';
 import config from 'temp/config';
 
-export class DictionaryServiceFactory {
```
ambrauer, line 4: I think we should stick to the single factory `DictionaryServiceFactory` (following the factory pattern, it's a factory that creates a `DictionaryService` instance). The check for environment variable / returning appropriate implementation can happen inside the `create()`.

ambrauer (changes requested):

samples/nextjs/src/lib/dictionary-service-factory.ts
CobyPear: That makes a lot more sense! Will change.

samples/nextjs/.env
```diff
@@ -36,6 +36,9 @@ SITECORE_API_HOST=
 # the resolved Sitecore API hostname + the `graphQLEndpointPath` defined in your `package.json`.
 GRAPH_QL_ENDPOINT=
 
+# Hardcoded FETCH_WITH env variable to prepare for yeoman generators
```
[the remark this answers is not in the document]
CobyPear: Updated

samples/nextjs/src/lib/dictionary-service-factory.ts
```diff
@@ -1,19 +1,29 @@
...
-      apiKey: config.sitecoreApiKey,
-      siteName: config.jssAppName,
-      /*
+    return process.env.FETCH_WITH === 'graphql'
+      ? new GraphQLDictionaryService({
+          endpoint: config.graphQLEndpoint,
+          apiKey: config.sitecoreApiKey,
+          siteName: config.jssAppName,
+          /*
       The Dictionary Service needs a root item ID in order to fetch dictionary phrases for the current
```
ambrauer, line 16: Indentation looks off with the comment here. Otherwise, looks good!

ambrauer (approved):

PR #901, opened 2022-01-05
Initializer fixes (round 2)

ambrauer:
## Description / Motivation
Another round of fixes addressing the following:

- 514353 Fix inconsistency with default values applied in `--yes` mode across initializers. Responsibility is now owned by the individual prompts themselves (using inquirer "when").
- 514355 Handle multiple templates on positional parameter as well

## Types of changes


illiakovalenko (approved):
Looks great 👍 
It's nice that we have defaults in one place, instead of how it was before

PR #914, opened 2022-01-25
[DOCS] Add getting started and working with older versions info to the README

ancaemcken:
## Types of changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,22 +1,88 @@
 # Sitecore JavaScript Rendering SDK (JSS)
 
-[![Build Status](https://dev.azure.com/sitecorejss/sitecore-jss-ci/_apis/build/status/Sitecore.jss?branchName=master)](https://dev.azure.com/sitecorejss/sitecore-jss-ci/_build/latest?definitionId=1?branchName=master) [![PRs Welcome](https://img.shields.io/badge/PRs-welcome-brightgreen.svg)](contributing.md#pull-requests)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,79 @@
...
+2. In a terminal, run the following command to create the JSS application:
+	```
+	jss create <your-app-name> <app-template-name>
+	```	
+	
+	Replace `<your-app-name>` with your desired application name, and `<app-template-name>` with the name of the template. 
+	
+	For example: `jss create my-first-jss-app nextjs`.
+	
+	>This command scaffolds a JSS application based on the ***latest version of JSS templates*** regardless of the JSS CLI version, because it pulls the template code from the `master` branch of the JSS repository.
```
ambrauer, line 41: Steps 1&2 are combined now. e.g.:
```
npm init sitecore-jss nextjs
```
or
```
npx create-sitecore-jss nextjs
```

Can probably get rid of the disclaimer (about branch usage)? Since templates are packaged in `create-sitecore-jss`, it's just a matter of using the proper version tag of the npm package.

ambrauer (approved):
Looks great! Approving, but please see comment about updating to use new create method.

ancaemcken:
@ambrauer I used the CLI getting started (instead of initializers) to have a version of this information for the current publicly released version (19.0.0). Your comments are strictly for release 20.0.0.  I think those should be incorporated closer to the release time. What do you think?

ambrauer:
> @ambrauer I used the CLI getting started (instead of initializers) to have a version of this information for the current publicly released version (19.0.0). Your comments are strictly for release 20.0.0. I think those should be incorporated closer to the release time. What do you think?

@ancaemcken Got it. The intent then would be to cherry pick this commit into `master` and `release/19.0.0` (minus the "TODO" comment for 20.0.0)? In that case, this line should be revised/removed as it doesn't make sense in that context:
> If you need to use an older version of JSS, you must install the JSS CLI version 19.x or older.

ancaemcken:
@ambrauer  I removed the todo comments and the potentially confusing paragraph about the CLI version. The doc is good for cherry-picking to master & release/19.0.0. We can then modify the readme for initializers in the dev branch.

PR #927, opened 2022-02-07
[Headless SXA] Nextjs-SXA empty add-on

sc-addypathania:
Added a `nextjs-sxa` empty add-on to the `create-sitecore-jss` `templates`.

## Types of changes


packages/create-sitecore-jss/src/templates/nextjs-sxa/package.json
```diff
@@ -0,0 +1,15 @@
+﻿{
```
ambrauer, line 1: I think we can get rid of this package.json example. It's referencing things that aren't used or don't exist.

packages/create-sitecore-jss/src/templates/nextjs-sxa/src/components/ContentBlock.tsx
```diff
@@ -0,0 +1,24 @@
+﻿import { Text, RichText, Field, withDatasourceCheck } from '@sitecore-jss/sitecore-jss-nextjs';
```
ambrauer, line 1: Looks like this and `lib/component-props` are just duplicates of the base nextjs template? Is there another basic component we can add here (hello world) instead of these files? Don't spend too much time as this is just a placeholder :)

ambrauer (changes requested):
Looks good, couple minor things with the placeholder files. Also FYI, we shouldn't merge this until release/20 branch has been created.

PR #934, opened 2022-02-23
[Nextjs] NextImage component (with more updates)

sc-addypathania:
New Next Image component.
What changed since the last merge:
- Removed the `blurDataURL`. It's up to the user if they want to pass it.
- Added the `priority` prop, if added the image will be considered high priority and will preload.
- Now user can pass their own loader function to the NextImage component.
- Fixed failing tests.

## Types of changes


ambrauer (changes requested):

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+      const rendered = mount(<NextImage loader={mockLoader} {...props} />);
+      expect(rendered.find('Image').prop('src')).to.equal('/~/jssmedia/img/test0.png?foo=bar');
+      rendered.setProps({
+        ...props,
+        field: { src: '/-assets/img/test0.png' },
+        width: 8,
+        height: 10,
+      });
+      expect(rendered.find('Image').prop('src')).to.equal('/-/jssmedia/img/test0.png?foo=bar');
+    });
```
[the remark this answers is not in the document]
sc-addypathania: Do you mean like if the custom loader is passed as a prop?

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+      expect(result).to.be.a('string');
+      expect(result).to.equal(`${params.config.path}/assets/img/test0.png?mw=100`);
+    });
+
+    it('should not append configPath when src is absolute', () => {
+      const params: ImageLoaderProps = {
+        config: {
+          deviceSizes: [],
+          imageSizes: [],
+          path: null,
```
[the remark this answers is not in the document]
sc-addypathania: Ohh right, that got overwritten while copy-pasting.

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
```diff
@@ -0,0 +1,239 @@
...
+      },
+      width: 8,
+      height: 10,
+    };
+
+    const rendered = mount(<NextImage loader={mockLoader} {...props} />).find('Image');
+
+    it('should render image with url', () => {
+      expect(rendered).to.have.lengthOf(1);
+      expect(rendered.prop('src')).to.equal(props.field.src);
```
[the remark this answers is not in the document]
sc-addypathania: So this only testing is the src width and height are passed as a prop correctly or not.
Without mocking the loader the actual function in the component throws errors.

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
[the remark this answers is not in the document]
ambrauer: Exactly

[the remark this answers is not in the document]
sc-addypathania: Added

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+        src: '/assets/img/test0.png',
+        width: 100,
+      };
+
+      const result = loader(params);
+      expect(result).to.be.a('string');
+      expect(result).to.equal(`${params.config.path}/assets/img/test0.png?mw=100`);
+    });
+
+    it('should not require config path when src is absolute', () => {
```
ambrauer, line 41: This test still doesn't look quite right. Let's create 2 tests for absolute:

1. 'should not require config path when src is absolute' -> with the config.path = undefined
2. 'should not append config path when src is absolute' -> with the config.path = 'https://cm.jss.localhost/'

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
```diff
@@ -0,0 +1,262 @@
...
+    };
+
+    const rendered = mount(<NextImage {...props} />).find('Image');
+
+    it('should render image with url', () => {
+      expect(rendered).to.have.lengthOf(1);
+      expect(rendered.prop('src')).to.equal(props.field.src);
+      expect(rendered.prop('width')).to.equal(props.width);
+      expect(rendered.prop('height')).to.equal(props.height);
+      expect(rendered.prop('loader')).to.equal(props.loader);
```
ambrauer, line 136: Would be good if we could also assert the mock loader was called

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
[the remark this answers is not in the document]
ambrauer: Ah right. Would be good if we can also perform some assertions on the mock loader (called and params).

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
```diff
@@ -0,0 +1,323 @@
...
+    const rendered = mount(<NextImage loader={mockLoader} {...props} />).find('Image');
+    it('should render image with url', () => {
+      rendered.render();
+      expect(rendered).to.have.lengthOf(1);
+      expect(rendered.prop('src')).to.equal(props.field.src);
+      expect(rendered.prop('width')).to.equal(props.width);
+      expect(rendered.prop('height')).to.equal(props.height);
+
+      expect(mockLoader.called).to.be.true;
+      expect(mockLoader).to.have.been.calledWith(loaderParams);
```
ambrauer, line 162: Having the entire `loaderParams` seems unnecessary (and brittle - these defaults may change with Next.js version updates). Ideally we can get rid of `loaderParams` and assert only the properties we're interested in (i.e. src, width). We should be able to do this [using sinon.match](https://stackoverflow.com/a/30875810/233024).

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
```diff
@@ -0,0 +1,323 @@
...
+  const rendered = mount(<NextImage {...props} />).find('Image');
+
+  it('should render image with url', () => {
+    expect(rendered).to.have.lengthOf(1);
+    expect(rendered.prop('src')).to.equal(props.field.src);
+    expect(rendered.prop('width')).to.equal(props.width);
+    expect(rendered.prop('height')).to.equal(props.height);
+    expect(rendered.prop('loader')).to.equal(props.loader);
+    expect(mockLoader.called).to.be.true;
+    expect(mockLoader).to.have.been.calledWith(loaderParams);
```
ambrauer, line 188: This is using the outer-scoped `mockLoader` - should be using `userMockLoader`.  Guessing this is working only b/c other tests are running previously. We should add an `afterEach` and call `mockLoader.reset()` to make sure there is no history leaking between tests.

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
```diff
@@ -0,0 +1,323 @@
...
+          allSizes: [],
+          loader: 'default',
+          domains: [],
+          disableStaticImages: false,
+          minimumCacheTTL: 1,
+          formats: [],
+          dangerouslyAllowSVG: false,
+          contentSecurityPolicy: 'test',
+        },
+        src: 'https://cm.jss.localhost/assets/img/test0.png?mw=100',
```
ambrauer, line 103: Drop the query string (wouldn't be here)
```suggestion
        src: 'https://cm.jss.localhost/assets/img/test0.png',
```

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
```diff
@@ -0,0 +1,323 @@
...
```
ambrauer, line 80: Drop the query string (wouldn't be here)
```suggestion
        src: 'https://cm.jss.localhost/assets/img/test0.png',
```

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
sc-addypathania: The loader function returns the url with the query string.

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
```diff
@@ -0,0 +1,318 @@
...
+const expect = chai.use(chaiString).expect;
+
+type MockLoaderType = ImageLoader extends SinonSpy ? SinonSpy : SinonSpy<any, any>;
+const customLoader = ({ src }) => new URL(`https://cm.jss.localhost${src}`).href;
+const mockLoader = (spy(customLoader) as unknown) as MockLoaderType;
+
+describe('<NextImage />', () => {
+  afterEach(() => {
+    mockLoader.resetHistory();
+  });
```
ambrauer, line 23: This `afterEach` is only effective for tests within the `describe` block. I'd recommend we move the `mockLoader` (lines 16-18) inside this describe block, and then also move inside of here every test that uses it (including the wrapping `describe` blocks). Which would be everything other than the 'Next loader function' ones. That gives us 2 outer blocks: 1) '<NextImage />' - tests the render / mount, and 2) 'Next loader function' - tests the loader function

sc-addypathania: Ohh but I tested it. afterEach was running for all the describes nested inside.
BUt what you suggest sounds better.

ambrauer (approved):
Looks good!

PR #946, opened 2022-03-11
#487176 added robots.txt implementation

matkovskyi:
Added robots.txt implementation, services for sitercore-jss/sitecore-jss-nextjs, sample for create-sitecore-jss

## Types of changes


packages/sitecore-jss-nextjs/src/services/graphql-robots-service.ts
```diff
@@ -0,0 +1,89 @@
+import {
```
ambrauer, line 1: Please move this service to `packages/sitecore-jss`. There's nothing nextjs-specific about this service (could technically be used in any framework). 

We have things organized by sub-modules (e.g. layout, i18n), and we should probably create one for this and upcoming SXA Headless work (redirects, sitemap.xml). 
Maybe call it `site` (since this is all based on `site` query / site settings)? @illiakovalenko @sc-addypathania what do you think?

You can then add pass-through exports to the framework SDKs via the root [`index.ts`](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss-nextjs/src/index.ts).

ambrauer (changes requested):
Overall looking good, but see comments

packages/sitecore-jss-nextjs/src/services/graphql-robots-service.ts
sc-addypathania: @ambrauer I agree. We should move this to its own submodule and directly import it from there.

packages/sitecore-jss-nextjs/src/services/graphql-robots-service.ts
```diff
@@ -0,0 +1,89 @@
...
+
+  /**
+   * The API key to use for authentication.
+   */
+  apiKey: string;
+}
+
+/**
+ * Service that fetch the robots.txt data using Sitecore's GraphQL API.
+ * @mixes Promise<String>
```
illiakovalenko, line 41: What is the purpose of `mixes` here? This class contains different methods

illiakovalenko (changes requested):
Looks good, see some comments

packages/sitecore-jss-nextjs/src/services/graphql-robots-service.test.ts
```diff
@@ -0,0 +1,86 @@
...
+              data: robotsQueryResultNull,
+            }
+      );
+  };
+
+  describe('Fetch robots.txt', () => {
+    it('should work if robots.txt has empty string', async () => {
+      mockPathsRequest();
+
+      const service = new GraphQLRobotsService({ endpoint, apiKey, siteName: '' });
```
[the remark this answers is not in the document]
matkovskyi: I added a new test if siteName is empty(changed logic of service)

.gitignore
```diff
@@ -22,4 +22,7 @@ lerna-debug.log
 coverage/
 /badges
 packages/*/*.tgz
-samples/
\ No newline at end of file
+samples/
+
+.idea
```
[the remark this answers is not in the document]
matkovskyi: I use sometimes WebStorm and this IDE create specials folder of settings. I added '.idea' folder in the ignore list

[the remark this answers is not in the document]
ambrauer: Got it. .idea is fine, but .next shouldn't be necessary.

packages/create-sitecore-jss/src/templates/nextjs-sxa/src/lib/next-config/plugins/robots.js
```diff
@@ -0,0 +1,16 @@
+const robotsPlugin = (nextConfig = {}) => {
+  return Object.assign({}, nextConfig, {
+    async rewrites() {
+      // When disconnected we proxy to the local faux layout service host, see scripts/disconnected-mode-proxy.ts
+      return [
```
ambrauer, line 5: This would override the rewrites defined in base plugins. The code here should check if `nextConfig` param has `rewrites` defined and append to those if so.

Also, the comment here looks to be a c/p artifact and should be updated or removed.

packages/sitecore-jss/src/site/graphql-robots-service.ts
```diff
@@ -0,0 +1,95 @@
...
+    site {
+      siteInfo(site: $siteName) {
+        robots
+      }
+    }
+  }
+`;
+
+/** @private */
+export const siteNameError = 'The jssAppName cannot be empty';
```
ambrauer, line 19:
```suggestion
export const siteNameError = 'The siteName cannot be empty';
```

ambrauer (changes requested):
Just a couple minor things. Looking really good!

ambrauer:
@matkovskyi You'll need to add `site.js` and `site.d.ts` files at the source root for the `site` submodule to work (causing the build errors). See [layout.js](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss/layout.js) and [layout.d.ts](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss/layout.d.ts) for example.

ambrauer:
> @matkovskyi You'll need to add `site.js` and `site.d.ts` files at the source root for the `site` submodule to work (causing the build errors). See [layout.js](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss/layout.js) and [layout.d.ts](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss/layout.d.ts) for example.

@matkovskyi One more thing required: You'll need to add `site.d.ts` to the [tsconfig.json exclude](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss/tsconfig.json#L30).

FYI, you can check the build locally for individual packages or entire monorepo using `yarn build`.

ambrauer (approved):
👍 Great work!

.gitignore
```diff
@@ -22,4 +22,7 @@ lerna-debug.log
...
+.next
```
illiakovalenko, line 28: Add new line

illiakovalenko (approved):
Looks good! Just one comment

PR #952, opened 2022-03-16
[Initializers] Add merge (concatenate) functionality for .env files

ambrauer:
## Description / Motivation
Allow add-on initializers to define only environment variables specific to the add-on. Currently, this is a simple file concatenation, which allows preservation of comments and whitespace. This means it's not possible to override existing values; however, this hasn't been identified as a use case.

## Types of changes


illiakovalenko (approved):
Looks good, approved 👍

PR #968, opened 2022-03-30
[React][EE] Placeholder key is not defined

sc-addypathania:
- Fixed issue for when trying to add a rendering to the second placeholder to the page, it will throw an error.
 - This issue is caused because the componentDidMount function to Placeholder class overrides the same function in PlaceholderCommon.

## Types of changes


ambrauer (approved):
Looks good, but this PR should target dev branch, right?

PR #971, opened 2022-04-01
[sitecore-jss-react] Make Image handle 'class' prop when it's passed down

art-alexeyenko:
Fixes the issue with 'class' property not being handled property, after it is passed into Image component. 
This could lead to a console error in Experience Editor when displaying a placeholder image.

## Description / Motivation
If "class" property is passed down from Layout Service it will be transformed into or appended to "className" instead.

## Testing Details
Unit test added.

## Types of changes

### DIFF
--- a/packages/sitecore-jss-react/src/components/Image.tsx
+++ b/packages/sitecore-jss-react/src/components/Image.tsx
@@ -99,6 +99,19 @@ const getImageAttrs = (
     return null;
   }
 
+  // we want to get rid of class prop in compliance with JSX
+  if (otherAttrs.class) {
+    // if any classes are defined properly already
+    if (otherAttrs.className) {
+      let className:string = otherAttrs.className as string;
+      className += ` ${otherAttrs.class}`;
+      otherAttrs.className = className;
+    }
+    else
+      otherAttrs.className = otherAttrs.class;
+    delete otherAttrs.class;
+  }
+
   const newAttrs: { [attr: string]: unknown } = {
     ...otherAttrs,
   };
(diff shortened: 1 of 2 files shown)


packages/sitecore-jss-react/src/components/Image.tsx
```diff
@@ -99,6 +99,19 @@ const getImageAttrs = (
     return null;
   }
 
+  // we want to get rid of class prop in compliance with JSX
```
ambrauer, line 102: This is an EE-only issue, correct? If so, can we do this inside the [EE-specific code block](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss-react/src/components/Image.tsx#L148-L155)? In another branch, we have this code [exported to be re-used in a Next.js image component](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss-react/src/components/Image.tsx#L148-L155). This would make it easier to merge this fix into this branch.

ambrauer (changes requested):
Looks good, just wondering if we can move this logic to ease concurrent work (see comment)

packages/sitecore-jss-react/src/components/Image.tsx
art-alexeyenko: Bug is EE-specific, but LayoutService can pass image JSON data with "class" param in every scenario. Whatever img tag is rendered by the renderField pipeline will be transformed into JSON without modifications. So it is important to do this check in all cases, not just EE.

ambrauer: Hmm. In normal mode, I don't know how the image field would have a 'class' prop in the JSON data from Layout Service, especially in the case the field is empty.

Also, whoops, I had incorrect URL for second link. Should be this: [exported to be re-used in a Next.js image component](https://github.com/Sitecore/jss/blob/feature/449842-nextjs-image/packages/sitecore-jss-react/src/components/Image.tsx#L124)

ambrauer: As discussed, let's stick with this, and I will add a note to account for this in the next/image component.

ambrauer (approved):

PR #978, opened 2022-04-12
[Nextjs] New NextImage component

sc-addypathania:
Previously we were importing the React image component to the next styleguide page.
With the latest changes in Nextjs and its Next/Image component, we need to leverage its latest out-of-the-box features.

- Added the new NextImage component to the @sitecore-jss-next package.
- With this change, Nextjs styleguide component will now use this new Next/Image component.
- Supports latest Next/Image features like improved performance, faster page loads, automatic image optimizations etc.

## Types of changes


packages/sitecore-jss-nextjs/package.json
```diff
@@ -73,7 +75,8 @@
     "@sitecore-jss/sitecore-jss-react": "^21.0.0-canary.30",
     "prop-types": "^15.7.2",
     "regex-parser": "^2.2.11",
-    "sync-disk-cache": "^2.1.0"
+    "sync-disk-cache": "^2.1.0",
+    "url-parse": "^1.5.10"
```
ambrauer, line 79: Shouldn't need `url-parse` dependency in this package any longer, right?

ambrauer (approved):
Approved. Think we can get rid of one added dependency, but otherwise looks great 👍

PR #988, opened 2022-04-18
[template/Nextjs] Add gitignore

sc-addypathania:
Added the ignore file to the nextjs template.

## Types of changes


ambrauer (approved):
Looks good. Just wondering if this belongs in `nextjs-styleguide` instead? Not sure, but looks like these ignored items might only be applicable to disconnected mode.

PR #1001, opened 2022-04-28
#526106: Extend SXA Headless layout with 3 placeholders instead of one

AntonKechashin:
- Added Header and Footer placeholders to the layout. Now placeholders are the same as in MVC SXA.
- The `<%- helper.getAppPrefix(appPrefix, appName) %>` prefix was removed. In SXA there are three predefined placeholders (Header, Main and Footer), which don't have any dynamic prefixes in their names.


ambrauer (approved):
Looks good to me

PR #1025, opened 2022-05-17
Adding nextjs-specific Layout Service configuration

art-alexeyenko:
## Description / Motivation
NextJS needs to have IncludeServerUrlInMediaUrls set to false without conflicting with other apps deployed to Sitecore.
This change ensures next uses its own configuration entry for Layout Service that inherits from default ones.

## Testing Details
Manual testing with angular and nextJS apps deployed at the same time: with REST and GraphQL fetch approaches.

## Types of changes

### DIFF
--- a/packages/create-sitecore-jss/src/templates/nextjs/sitecore/config/{{appName}}.config
+++ b/packages/create-sitecore-jss/src/templates/nextjs/sitecore/config/{{appName}}.config
@@ -88,7 +88,7 @@
           compatible with JSS versions < 18.0.0).
         -->
         <app name="<%- appName %>"
-            layoutServiceConfiguration="default"
+            layoutServiceConfiguration="next-default"
             sitecorePath="/sitecore/content/<%- appName %>"
             useLanguageSpecificLayout="true"
             graphQLEndpoint="/sitecore/api/graph/edge"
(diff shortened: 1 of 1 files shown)


packages/create-sitecore-jss/src/templates/nextjs/sitecore/config/{{appName}}.config
```diff
@@ -145,19 +145,21 @@
     -->
     <layoutService>
       <configurations>
-        <config name="default">
-          <rendering>
-            <renderingContentsResolver>
+        <config name="next-default">
```
ambrauer, line 148: I'm surprised this works for edge GraphQL endpoint. From what I remember, this is hard-coded to use "default". Did you try setting "default" to "true" along with this set to "false"? I'm guessing the server URL would be included in media URLs.

ambrauer, line 148: If this does work, we would need to also set the layout service to use it: https://github.com/Sitecore/jss/blob/dev/packages/create-sitecore-jss/src/templates/nextjs/src/lib/layout-service-factory.ts#L21

packages/create-sitecore-jss/src/templates/nextjs/sitecore/config/{{appName}}.config
```diff
@@ -145,19 +145,21 @@
...
+            <renderingContentsResolver ref="/sitecore/layoutService/configurations/config[@name='default']/rendering/renderingContentsResolver">
               <IncludeServerUrlInMediaUrls>false</IncludeServerUrlInMediaUrls>
             </renderingContentsResolver>
           </rendering>
+          <serialization ref="/sitecore/layoutService/configurations/config[@name='default']/serialization"/>
         </config>
-        <config name="jss">
-          <rendering>
-            <renderingContentsResolver>
+        <config name="next-jss">
```
ambrauer, line 156: I'm not sure this "jss" config is even used any longer for Next.js. I think we could actually remove it (would need to verify).

PR #1035, opened 2022-05-20
fixing query and output type with personalization enabled

art-alexeyenko:
Some fixes to graphQL site query - bringing it to common form, fixing personalization type placement

## Description / Motivation
Changes to personalize query and output type.

## Types of changes

### DIFF
--- a/packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
+++ b/packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
@@ -34,13 +34,16 @@ query ${usesPersonalize ? 'PersonalizeSitemapQuery' : 'DefaultSitemapQuery'}(
           endCursor
           hasNext
         }
-        results: routesResult{
+        results {
           path: routePath 
           ${
             usesPersonalize
-              ? `personalize: {
-            variantIds
-          }`
+              ? `
+              route {
+                personalization {
+                  variantIds
+                }
+              }`
               : ''
           }
         }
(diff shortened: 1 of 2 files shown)


packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -204,7 +209,9 @@ export class GraphQLSitemapService {
    */
   protected async fetchSitemap(
     languages: string[],
-    formatStaticPath: (path: string[], language: string) => StaticPath
+    formatStaticPath: (path: string[], language: string) => StaticPath,
+    includedPaths?: string[],
+    excludedPaths?: string[]
```
ambrauer, line 214: Why were these parms added? Doesn't look like these are used (by the 2 public functions that call it). If the intent was to add these to the public fetch functions, then we should exclude from the service options (would be confusing to have in both places).

ambrauer (changes requested):
See comment. Additionally, I realized we never added a `graphql-sitemap-service` plugin to the `nextjs-personalize` add-on template. The plugin would be identical/override [the base one](https://github.com/Sitecore/jss/blob/dev/packages/create-sitecore-jss/src/templates/nextjs/src/lib/sitemap-fetcher/plugins/graphql-sitemap-service.ts), but would have the `includePersonalizedRoutes` set to `true`.

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
art-alexeyenko: done

PR #1058, opened 2022-06-10
 #533938 added styles sass

matkovskyi:
added sass styles from Themes/Basic2 which uses in the SXA Headless


matkovskyi:
> Do we want to store all the styles in the app? Can styles be fetched from the sitecore? Similar question about `fontawesome` - should it be installed from the npm?

I will have meeting about managed classes and do we need them in this module. I let you know

What about fontawesome - agreed with you

packages/create-sitecore-jss/src/templates/nextjs-sxa/package.json
```diff
@@ -1,5 +1,9 @@
 {
   "dependencies": {
     "bootstrap": "^5.1.3"
```
ambrauer, line 3: Do we still need bootstrap (here and in `_app.tsx`)?

matkovskyi: Yes, the bootstrap uses with our styles.

matkovskyi:
> Do we want to store all the styles in the app? Can styles be fetched from the sitecore? Similar question about `fontawesome` - should it be installed from the npm?


I investigated this question. Unfortunately we use fontsawesome 4.7.0 version and this version not available on the official npm package server. I tried to update version but faced missing some variables in the newest version of fontsawesome.

matkovskyi:
> @matkovskyi Did you use this npm package? It's an official fontawesome 4.7.0 package and should be used https://www.npmjs.com/package/font-awesome

Yes! You're right! I saw official site and there was link for only source of 4.7.0 version. thanks. I removed local font-awesome

ambrauer (approved):
Looks good to me!

PR #1071, opened 2022-06-22
Proper building of query string inside EditingRenderMiddleware

Antonytm:
## Description / Motivation
I want to override default `resolvePageUrl` and add GET parameter.
I use the way described in the official documentation:
https://doc.sitecore.com/xp/en/developers/hd/200/sitecore-headless-development/override-default-protocols-using-the-editingrendermiddleware-in-next-js-jss-apps.html?utm_medium=rss&utm_source=dev.sitecore.net

```
const handler = new EditingRenderMiddleware({
  resolvePageUrl: (serverUrl: string, itemPath: string) => `${serverUrl}${itemPath}?someMyKey=${myKeyValue}&`,
}).getHandler();
```

Why do I need this parameter: I need to indicate to some components that the request came from EE/Horizon. 
When I add only `someMyKey=${myKeyValue}` it doesn't work because of adding `?timestamp=${Date.now()}`. The result string will have 2 `?` signs, which is not right. I can do some tricks and add `?someMyKey=${myKeyValue}&`(with ampersand an at the end). It will give me 2 parameters: `someMyKey` and `?timestamp`. 

As `timestamp` parameter is added only to avoid cache, it doesn't matter how it will be named: `timestamp` or `?timestamp`. It makes this change very minor. But if we can have a proper GET parameter name, why don't have it?

## Testing Details

1. `yarn install`
1. `yarn build`
1. Copy only `editing-render-middleware.js`
1. Add logging of URL (to make sure that it was changed in the proper way)
1. Try different actions in Experience Editor (adding and removing different components)
1. Review logs: look for "timestamp"

## Types of changes


ambrauer (approved):
Hi @Antonytm! The change makes sense to me. While we never anticipated query strings in the `resolvePageUrl` output, we should definitely be defensive and handle this case.

Thanks for the contribution 😊

However, in your case I'm curious why not use other means to detect editing context? We have the `pageEditing` prop available on the Sitecore Context (you can access either through a React [hook](https://github.com/Sitecore/jss/blob/release/20.0.0/ref-docs/sitecore-jss-nextjs/modules/index.md#usesitecorecontext) or [HOC](https://github.com/Sitecore/jss/blob/release/20.0.0/ref-docs/sitecore-jss-nextjs/modules/index.md#withsitecorecontext), see [here](https://github.com/Sitecore/jss/blob/release/20.0.0/packages/create-sitecore-jss/src/templates/nextjs-styleguide/src/components/styleguide/Styleguide-Layout-Tabs.tsx#L47) for example) and also [isEditorActive](https://github.com/Sitecore/jss/blob/release/20.0.0/ref-docs/sitecore-jss-nextjs/modules/index.md#iseditoractive) for detection in the browser.

ambrauer:
@Antonytm I agree, we should be clear in the docs that `isEditorActive` only works in the browser. I've added a backlog item for us to address this in an upcoming release. Great feedback, thanks!

Option 2 isn't really feasible since how you access the Sitecore Context in each framework differs.

PR #1081, opened 2022-06-28
#539752 The site can't be opened in the EE and Pages in case the rend…

sc-dawidrutkowski:
The graphql-redirects-service.ts service needs to be additionally protected in case some data will be removed for the Edge (e.g.: redirects information).

## Types of changes


illiakovalenko (changes requested):
You can write

```
.then((result: RedirectsQueryResult) => result?.site?.siteInfo?.redirects || [])
```
And you need to add unit test cover such case

PR #1107, opened 2022-07-16
[Vue] Fix compilers proxy options in template

iamandrewluca:
## Description / Motivation

After some analysis of the project, it seems the `compilers` option is not used anymore.  
It was renamed to `requireArg?: string | null;`

In general, I see that the Vue template is very outdated.
I'm working on a new template that uses Vite and supports TypeScript
https://github.com/jagaad/create-sitecore-jss-vue


ambrauer (approved):
Good catch, thanks for the fix!

As for the template in general, we do have it in our backlog to support TypeScript. We are also considering options for SSG support in the future (Nuxt or Vite), but this is not in the roadmap as of yet.

PR #1113, opened 2022-07-20
 #539620: [SXA] fixed navigation link in editing mode

matkovskyi:
fixed navigation link in editing mode


packages/create-sitecore-jss/src/templates/nextjs-sxa/src/components/Navigation.tsx
```diff
@@ -126,3 +139,15 @@ const NavigationList = (props: NavigationProps) => {
   );
 };
 
+const getLinkTitle = (props: NavigationProps): string | undefined => {
+  let title;
+  if (props.fields.NavigationTitle) {
```
ambrauer, line 144: Should we check `.value` as well with these if conditions?

ambrauer (approved):
Just one suggestion, otherwise looks good to me

PR #1120, opened 2022-08-03
Error boundaries for dynamic components

art-alexeyenko:
Adds error boundary component that wraps around dynamic components under placeholders

## Description / Motivation
Error boundaries would allow to contain rendering errors to one component only - making sure other renderings under a placeholder are displayed fine.
This update adds an out of the box default error boundary component, as well as the possibility to pass down custom error boundaries and error components (displayed on error) into a placeholder.

## Testing Details
Works in connected, disconnected modes. An extra unit test added, with tweaks to existing ones.

## Types of changes

### DIFF
--- a/packages/create-sitecore-jss/src/templates/react/src/index.js
+++ b/packages/create-sitecore-jss/src/templates/react/src/index.js
@@ -9,6 +9,14 @@ import AppRoot from './AppRoot';
 import GraphQLClientFactory from './lib/GraphQLClientFactory';
 import config from './temp/config';
 import i18ninit from './i18n';
+import { stopReportingRuntimeErrors } from 'react-error-overlay';
+
+/* as we handle and show errors via error boundaries and console we want to avoid error overlay showing up */
+/* in part because it's not working correctly at the moment: https://github.com/facebook/create-react-app/issues/11773 */
+/* please feel free to disable this if overlay is needed for your dev environments */
+if (process.env.NODE_ENV === 'development') {
+  stopReportingRuntimeErrors();
+}
 
 if (process.env.REACT_APP_DEBUG) {
   enableDebug(process.env.REACT_APP_DEBUG);
--- a/packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
+++ b/packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
@@ -21,6 +21,7 @@ export interface ErrorBoundaryProps {
 export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorState> {
   constructor(props: ErrorBoundaryProps) {
     super(props);
+    this.state = {};
   }
 
   previousError: Error;
--- a/packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
+++ b/packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
@@ -1,0 +1,57 @@
+import React, { Component, ErrorInfo, ReactNode } from 'react';
+
+interface ErrorState {
+  hasError?: boolean;
+  error?: Error;
+  errorInfo?: ErrorInfo;
+}
+
+type ErrorComponentProps = {
+  [prop: string]: unknown;
+};
+
+export interface ErrorBoundaryProps {
+  errorComponent?: React.ComponentClass<ErrorComponentProps> | React.FC<ErrorComponentProps>;
+  children?: ReactNode;
+}
+
+export class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorState> {
+  constructor(props: ErrorBoundaryProps) {
+    super(props);
+    this.state = { hasError: false };
+  }
+
+  defaultErrorComponent: React.FC<{ error: Error; errorInfo: ErrorInfo }> = ({
+    error,
+    errorInfo,
+  }) => {
+    return (
+        <div className="sc-jss-component-error">
+          <p>A rendering error occurred: {error.message}.</p>
+          <p>Stack trace: {errorInfo.componentStack}</p>
+        </div>
+    );
+  };
+
+  componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+    this.setState({ error, errorInfo });
+  }
+
+  render() {
+    if (this.state.hasError || this.state.error) {
+      if (this.props.errorComponent) {
+        return (
+          <this.props.errorComponent
+            errorMessage={this.state.error.message}
+            errorStack={this.state.errorInfo.componentStack}
+          />
+        );
+      } else {
+        return (
+          <this.defaultErrorComponent error={this.state.error} errorInfo={this.state.errorInfo} />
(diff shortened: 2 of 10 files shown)


packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -245,13 +253,15 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
...
-        renderingDefinition.params.FieldNames
-      );
+      return withErrorBoundary(
+        this.props.errorComponent,
+        this.props.errorBoundaryComponent
+      )(componentFactory(renderingDefinition.componentName, renderingDefinition.params.FieldNames));
     }
 
-    return componentFactory(renderingDefinition.componentName);
+    return withErrorBoundary(this.props.errorComponent)(
```
ambrauer, line 262: Looks like this is missing the `this.props.errorBoundaryComponent` parameter?

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -245,13 +253,15 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
 
     // Render SXA Rendering Variant
     if (renderingDefinition.params?.FieldNames) {
-      return componentFactory(
-        renderingDefinition.componentName,
-        renderingDefinition.params.FieldNames
-      );
+      return withErrorBoundary(
```
ambrauer, line 256: Might be better to add the Error Boundary outside of this function. Would avoid duplicate code (and let it focus solely on resolving the component from the component factory). In `getComponentsForRenderingData` around line 218, for example.

packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,65 @@
+import React, { Component, ErrorInfo, ReactNode } from 'react';
+
+interface ErrorState {
+  error?: Error;
+  errorInfo?: ErrorInfo;
+}
+
+type ErrorComponentProps = {
```
ambrauer, line 8: Duplicate of `ErrorComponentProps` defined in `PlaceholderCommon.tsx`. We should consolidate/re-use. 

And looking at this, shouldn't we be able to be more specific with this type? i.e.

```
export type ErrorComponentProps = {
  error?: Error;
};
```

packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,65 @@
...
+        return (
+          <this.props.errorComponent
+            errorMessage={this.state.error.message}
+            errorStack={this.state.errorInfo.componentStack}
+          />
+        );
+      } else {
+        return (
+          <this.defaultErrorComponent error={this.state.error} errorInfo={this.state.errorInfo} />
+        );
```
ambrauer, line 60: There shouldn't be any difference in prop type between a set errorComponent (`this.props.errorComponent`) and the default (`this.defaultErrorComponent`). In both cases, seems like we can only have a single `error` prop which is type `Error` (see related comment above). This would allow the re-use with our [existing error component logic](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss-react/src/components/Placeholder.tsx#L76).

packages/sitecore-jss-react/src/index.ts
```diff
@@ -64,6 +65,7 @@ export {
   WithSitecoreContextProps,
   WithSitecoreContextHocProps,
 } from './enhancers/withSitecoreContext';
+export { withErrorBoundary } from './enhancers/withErrorBoundary';
```
ambrauer, line 68: Wondering if it will be confusing to expose this HOC in the public API surface as we are _automatically_ using it for every component used in a placeholder. Trying to think of a use case other than our own internal usage in `PlaceholderCommon.tsx`. Do we think this will be used on components that are not dynamic?

packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return (
+      <div className="sc-jss-component-error">
+        <p>A rendering error occurred: {error.message}.</p>
+        <p>Stack trace: {errorInfo.componentStack}</p>
+      </div>
+    );
+  };
+
+  componentDidCatch(error: Error, errorInfo: ErrorInfo) {
+    if (JSON.stringify(error) !== JSON.stringify(this.previousError)) {
```
ambrauer, line 42: What's the reason for tracking/checking previous error? Is it required or extraneous optimization?

packages/create-sitecore-jss/src/templates/react/src/index.js
```diff
@@ -9,6 +9,14 @@ import AppRoot from './AppRoot';
...
 import config from './temp/config';
 import i18ninit from './i18n';
+import { stopReportingRuntimeErrors } from 'react-error-overlay';
+
+/* as we handle and show errors via error boundaries and console we want to avoid error overlay showing up */
+/* in part because it's not working correctly at the moment: https://github.com/facebook/create-react-app/issues/11773 */
+/* please feel free to disable this if overlay is needed for your dev environments */
+if (process.env.NODE_ENV === 'development') {
+  stopReportingRuntimeErrors();
+}
```
ambrauer, line 19: What's the implication if this isn't done? If it is just extra error notification, then might be willing to accept that to lose the added dependency to manage on `react-error-overlay`.

ambrauer (changes requested):

packages/create-sitecore-jss/src/templates/react/src/index.js
art-alexeyenko: react-error-overlay is already added via react itself - in here it's just a direct import
without this users would see an iframe with error description in connected mode, taking up the whole page for react apps.

packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
art-alexeyenko: I noticed that a single error from a single component is reported several times in browser console. As we're updating state when error occurs I thought it's best to only do that once for duplicating errors from the same component.

packages/sitecore-jss-react/src/index.ts
art-alexeyenko: I thought why not make this an option if anyone wants to use it any time in future, but you're probably right - if anyone needs the OOB HOC for this - they'll let us know.

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
art-alexeyenko: errorComponent is what will be shown in case of error
erroBoundaryComponent is custom error boundary - if anyone decides to use it

PR #1122, opened 2022-08-09
 #546451: [SXA] added fix when receive from backend haven't FiledsName

matkovskyi:


kendoce (commented):
Is it necessary to check the Default first? can we check it after .default?

matkovskyi:
> 

don't matter because there will be only one variant

kendoce (approved):

PR #1123, opened 2022-08-09
Configurable timeout for CDP requests

art-alexeyenko:
Adds an env variable to set CDP request timeout

## Types of changes


packages/create-sitecore-jss/src/templates/nextjs-personalize/.env
```diff
@@ -12,3 +12,6 @@ NEXT_PUBLIC_CDP_SCRIPT_URL=https://d1mj578wat5n4o.cloudfront.net/boxever-1.4.8.m
 
 # Your Sitecore CDP point of sale
 NEXT_PUBLIC_CDP_POINTOFSALE=
+
+# Timeout for CDP requests to respond within (ms)
+PERSONALIZE_MIDDLEWARE_TIMEOUT=500
```
sc-addypathania, line 17: Did we decide the default value to be 500?
I remember in some situations it took us more than 1000ms.

ambrauer, line 17: Yes, we might want to bump this to 10000. For local development we were seeing higher response times. 
Also, worth considering, should we introduce a `.env.development` file ([Next.js supports these](https://nextjs.org/docs/basic-features/environment-variables#default-environment-variables)) and add this here instead (so it only applies in dev)?

ambrauer (changes requested):

packages/create-sitecore-jss/src/templates/nextjs-personalize/.env
art-alexeyenko: set dev value for 10k now

packages/create-sitecore-jss/src/templates/nextjs-personalize/src/lib/middleware/plugins/personalize.ts
```diff
@@ -20,6 +20,8 @@ class PersonalizePlugin implements MiddlewarePlugin {
         endpoint: process.env.NEXT_PUBLIC_CDP_API_URL || '',
         clientKey: process.env.NEXT_PUBLIC_CDP_CLIENT_KEY || '',
         pointOfSale: process.env.NEXT_PUBLIC_CDP_POINTOFSALE || '',
+        timeout:
```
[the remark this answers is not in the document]
art-alexeyenko: edge endpoint differs from CDP endpoint - so I added an extra env value

packages/create-sitecore-jss/src/templates/nextjs-personalize/.env
art-alexeyenko: done

packages/create-sitecore-jss/src/templates/nextjs-personalize/.env.production
```diff
@@ -0,0 +1,5 @@
+# Timeout for CDP requests to respond within (ms)
```
ambrauer, line 1: Do we need the .env.production file? We already have a fallback of 250 in the personalze.ts file and in the SDK middleware itself. We can leave addl production config up to their deployment platform of choice (e.g. set in Vercel project).

ambrauer (approved):
See some final suggestions, otherwise looks good.

packages/create-sitecore-jss/src/templates/nextjs-personalize/.env.production
art-alexeyenko: Yes - if we want it to be configurable. If variables are put into regular env file that would be the value used over others. So I've put both dev and prod env files to differentiate between usage scenarios.

ambrauer: Hmm... I guess I'm just leery of introducing more of these .env files which start dictating a customer's setup for prod. 

Maybe it's best we just go back to how you had it originally in our single .env file and add a comment that these should be tuned down for production? This falls more in line with other variables which need to be adjusted per environment (e.g. PUBLIC_URL).

PR #1142, opened 2022-09-02
Disable scaffold apps during publish runs

art-alexeyenko:
Follow up to lint CI pull request

## Description / Motivation
Disable scaffold and lint apps for publish CI runs to avoid git errors.


ambrauer (approved):
Looks good. Just one thought: should we introduce another variable (e.g. "shouldCheckSamples") to break our tie to publishing? Currently this works, but may not in the future. I'm good either way :)

PR #1153, opened 2022-09-12
[Next.js][Personalize] Finalize targeting experience call

ambrauer:
## Description / Motivation
We need to make some adjustments to the CDP callFlows call (which occurs in personalize middleware) according to final specs.

- Removed `geo` from `params` (this is now being handled entirely on CDP-side via Cloudflare)
- Removed `ua` from `params` (moved to headers - see next)
- Added user agent to the callFlows fetch request headers ('User-Agent')
- Added ability to configure `channel` (default to 'WEB')

## Testing Details
Local testing using mock CDP

## Types of changes


illiakovalenko (approved):
Awesome 👍

PR #1157, opened 2022-09-15
Add Scripts component to SXA layout

art-alexeyenko:
Aligns Layout output between nexjs and sxa app templates so that Scripts component is used in both (instead of adding VisitorIdentification by default).

## Types of changes


sc-addypathania (approved):
Looks good

PR #1160, opened 2022-09-16
Revert "[Nextjs] Added safeHydrate component to fix hydration error"

sc-addypathania:
Reverts Sitecore/jss#1143


art-alexeyenko (approved):
Yup, let's revert this.

PR #1161, opened 2022-09-19
Getting rid of hydration errors for nextjs pages in ExpEditor

art-alexeyenko:
This adds a workaround for Experience Editor to make it work better with nextjs apps and avoid hydration errors that started occurring after next and react version upgrades.

## Description / Motivation
React will add specific markup for empty placeholders in Experience Editor on first page load. This way it won't be "confused" by this markup when Experience Editor client side scripts add it.

## Types of changes


ambrauer (approved):
Looks good to me. Just curious - do we know if the same solution will be necessary in Horizon/Pages?

art-alexeyenko:
@ambrauer it shouldn't be. Placeholders appear to behave slightly different on Pages and hydration error does not occur when I tested (at least on XM Cloud deployment).

PR #1166, opened 2022-09-23
[Next.js][Personalize] Switchover to new @sitecore/engage SDK for page view events

ambrauer:
## Description / Motivation
* Switch to use new SitecoreCloud Engage SDK
* Consolidate `NEXT_PUBLIC_CDP_API_URL` and `NEXT_PUBLIC_CDP_TARGET_URL` to single `NEXT_PUBLIC_CDP_TARGET_URL`
* Send (properly formatted) pageVariantId in page view event
* Use uppercase "BID_" browserId cookie prefix for consistency with SitecoreCloud Engage SDK
* Emphasize cookie consent management implementation for `disabled` hooks (in both page view events and personalize middleware)
* Moved `getContentId` to new `CdpHelper` (for centralized id formatting logic)
* Renamed `CdpIntegrationScript` to `CdpPageView` component

## Types of changes


packages/create-sitecore-jss/src/templates/nextjs-personalize/src/components/CdpIntegrationScript.tsx
```diff
@@ -1,89 +1,66 @@
...
+   * Creates a page view event using the Sitecore Engage SDK.
+   */
+  const createPageView = async (page: string, language: string, pageVariantId: string) => {
+    const engage = await init({
+      clientKey: process.env.NEXT_PUBLIC_CDP_CLIENT_KEY || '',
+      targetDomain: process.env.NEXT_PUBLIC_CDP_API_DOMAIN || '',
+      // Replace with the top level cookie domain of the website that is being integrated e.g ".example.com" and not "www.example.com"
+      cookieDomain: window.location.host.replace(/^www\./, ''),
+      // Cookie may be created in personalize middleware (server), but if not we should create it here
+      forceServerCookieMode: false,
```
elolos, line 27: I think you should set this to true, as client-set cookies won't be able to avoid ITP restrictions.

packages/create-sitecore-jss/src/templates/nextjs-personalize/src/components/CdpIntegrationScript.tsx
```diff
@@ -1,89 +1,66 @@
...
+    });
```
elolos, line 28: This may be temporary from your side but it's probably going to be more efficient to have initialization run less often

packages/create-sitecore-jss/src/templates/nextjs-personalize/src/components/CdpIntegrationScript.tsx
```diff
@@ -1,89 +1,66 @@
...
-  });
-}
-
+/**
+ * This is the CDP integration script which uses the Sitecore Engage SDK.
+ * It is used to enable page view events.
+ * See Sitecore Engage SDK documentation for details.
+ * https://www.npmjs.com/package/@sitecore/engage
+ */
 const CdpIntegrationScript = (): JSX.Element => {
```
elolos, line 12: You should probably rename this, right?

illiakovalenko (approved):
Looks good!

packages/create-sitecore-jss/src/templates/nextjs-personalize/src/components/CdpIntegrationScript.tsx
ambrauer: Yeah, if a customer (or us, in the future) were using engage in more places it would make sense to place this somewhere else (e.g. a React Context) so that it can be re-used. However, for now it's meant to be as condensed/concise as possible for an example. Also, we can't easily introduce changes to higher levels (e.g. `[[...path]].tsx`, `_app.tsx`, or `Layout.tsx`) as we have to consider how that affects the base `nextjs` template and other add-ons (e.g. `nextjs-sxa`).

packages/create-sitecore-jss/src/templates/nextjs-personalize/src/components/CdpIntegrationScript.tsx
ambrauer: Let's discuss. Right now, the personalize middleware _only_ creates/sets the browserId cookie if there were personalization on the page (not every time in interest of performance).

packages/create-sitecore-jss/src/templates/nextjs-personalize/src/components/CdpIntegrationScript.tsx
ambrauer: I assume you mean switching to something like `EngageIntegrationScript`? We discussed this a bit but that seems more of an implementation detail. Whereas this is meant to indicate integration with Sitecore _CDP_ product.

packages/create-sitecore-jss/src/templates/nextjs-personalize/src/components/CdpIntegrationScript.tsx
elolos: In that case it's fine and you can probably point them towards the Engage samples for nextjs which will hopefully show a bigger range of example use cases.

packages/create-sitecore-jss/src/templates/nextjs-personalize/src/components/CdpIntegrationScript.tsx
elolos: Sorry, my comment was a bit cryptic. No, I meant that this is no longer loading a script, so that last part of the name could be misleading.

In terms of mentioning "engage" or "CDP", I'd probably avoid mentioning either and go for something more descriptive. I know that tracking is a "forbidden" term, otherwise I'd suggest something that matches the comments like "PageViewEventTracker" or similar.

ambrauer: Ah, this is a good point. Even though this is included through our [Scripts](https://github.com/Sitecore/jss/pull/1166/files#diff-0dae17efb21ff6cfc0b89abf0a6bf169fecc27e153cd7318d2de05d1dd2ad27c) abstraction, it doesn't need to use this term itself. I went with "CdpPageView" (to avoid "tracking" as you say) :)

PR #1177, opened 2022-09-29
Ensure headers from middleare are passed into NativeFetcher

art-alexeyenko:
Fixing the issue when user-agent was ignored in fetcher when executing CDP experience.

## Testing Details
To be further tested on Vercel or with mock.

## Types of changes

### DIFF
--- a/packages/sitecore-jss/src/native-fetcher.ts
+++ b/packages/sitecore-jss/src/native-fetcher.ts
@@ -15,10 +15,6 @@ type NativeDataFetcherOptions = {
    * Optional request timeout.
    */
   timeout?: number;
-  /**
-   * optional headers to use in fetch calls
-   */
-  headers?: Record<string, string>;
 };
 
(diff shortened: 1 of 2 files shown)


packages/sitecore-jss/src/native-fetcher.ts
```diff
@@ -15,6 +15,10 @@ type NativeDataFetcherOptions = {
    * Optional request timeout.
    */
   timeout?: number;
+  /**
+   * optional headers to use in fetch calls
+   */
+  headers?: Record<string, string>;
```
ambrauer, line 21: Shouldn't need this here, it's already on `RequestInit` type (which is part of final `NativeDataFetcherConfig` - see below).

art-alexeyenko: oh yeah, you're right

ambrauer (approved):
Looks good!

PR #1196, opened 2022-10-07
[sitecore-jss] `rootItemId` not resolvable for different language versions 

sc-addypathania:
The dictionary service doesn't return `rootItemId` for other language versions other than English.
Added a fix to make another request using language as `en` to get the `rootItemId` as it is same in all versions.

## Types of changes

### DIFF
--- a/packages/sitecore-jss/src/graphql/app-root-query.ts
+++ b/packages/sitecore-jss/src/graphql/app-root-query.ts
@@ -1,6 +1,5 @@
 import { GraphQLClient } from './../graphql-request-client';
 import { SitecoreTemplateId } from '../constants';
-import debug from '../debug';
 
 /** @private */
 export const siteNameError = 'The site name must be a non-empty string';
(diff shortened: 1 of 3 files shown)


packages/sitecore-jss/src/graphql/app-root-query.ts
```diff
@@ -65,12 +65,20 @@ export async function getAppRootId(
 
   debug.dictionary('fetching site root for %s %s', language, siteName);
```
ambrauer, line 66: Noticed this debug log for dictionary namespace here. We should move this to the `GraphQLDictionaryService` (just before the `getAppRootId` call) instead.

packages/sitecore-jss/src/graphql/app-root-query.ts
```diff
@@ -65,12 +65,20 @@ export async function getAppRootId(
...
 
-  const fetchResponse = await client.request<AppRootQueryResult>(appRootQuery, {
+  let fetchResponse = await client.request<AppRootQueryResult>(appRootQuery, {
     jssAppTemplateId: jssAppTemplateId || SitecoreTemplateId.JssApp,
     siteName,
     language,
   });
 
+  if (!fetchResponse?.layout?.homePage?.rootItem?.length) {
```
ambrauer, line 74: We should add ` && language !== 'en'` (no need to try with en if we already did)

ambrauer (changes requested):
See comments. Also, please [add unit tests](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss/src/graphql/graphql.test.ts#L14) for this new case.

packages/sitecore-jss/src/graphql/graphql.test.ts
```diff
@@ -44,7 +44,7 @@ describe('graphql', () => {
             },
           });
 
-        const result = await getAppRootId(client, 'siteName', 'language');
+        const result = await getAppRootId(client, 'siteName', 'en');
```
ambrauer, line 47: Seems like we should have 2 new tests which are variants of these existing "should return" ones.
1. 'valid app root ID' (existing - mocks a "hit" for the passed 'language')
2. 'valid app root ID with en fallback' (new - mocks a "miss" for the passed 'language' but "hit" for 'en')
3. 'null if no app root found' (existing - mocks a "miss" for 'en')
4. 'null if no app root found with en fallback' (new - mocks a "miss" for the passed 'language' and "miss" for 'en')

ambrauer (approved):
Approving, but see suggestion about tests :)

PR #1216, opened 2022-10-26
[React] export DateFieldProps

LeanTvener11:
Export DataFieldProps for React DateField component.

Description / Motivation
DataFieldProps have been added to the index fil for export, so that I mirrors all the other components in the jss-React. Then It can be used along side the DataField component. 

Change type
BugFix


ambrauer (approved):
👍

PR #1223, opened 2022-10-31
add async to editing data cache

erwinsmit:
## Description / Motivation
For deploying Sitecore JSS Next.js to Azure Static Web Apps I had to provide my own implementation of the EditingDataCache class (using Azure Blobstorage instead of disk cache):

See https://github.com/erwinsmit/swa-jss/blob/master/src/lib/editing/editing-data-cache.ts for an example.
In this example I had to copy more stuff out of the JSS repo to make it work. If async was supported that wouldn't be necessary. 

The problem I had was that the "EditingDataService" was expecting the "EditingDataCache" to be sync, not async. This limits the implementation of the EditingDataCache class, it's not possible for example to write/read stuff to Blobstorage in a sync manner. 

With my suggested changes it's possible to use sync and async EditingDataCache implementations.  

## Testing Details
I've tested my changes on this repo:
https://github.com/erwinsmit/swa-jss/

You can also read my blogpost that explains it further:
https://www.erwinsmit.com/sitecore-jss-nextjs-on-azure-static-web-app/

## Types of changes

### DIFF
--- a/packages/sitecore-jss-nextjs/src/editing/editing-data-cache.ts
+++ b/packages/sitecore-jss-nextjs/src/editing/editing-data-cache.ts
@@ -8,8 +8,8 @@ import { EditingData } from './editing-data';
  * Defines an editing data cache implementation
  */
 export interface EditingDataCache {
-  set(key: string, editingData: EditingData): void;
-  get(key: string): EditingData | undefined;
+  set(key: string, editingData: EditingData): Promise<void>;
+  get(key: string): Promise<EditingData | undefined>;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


erwinsmit:
Could someone look at this please? I would be happy to make any changes if necessary.

packages/sitecore-jss-nextjs/src/editing/editing-data-cache.ts
```diff
@@ -27,22 +27,28 @@ export class EditingDataDiskCache implements EditingDataCache {
...
 
-  set(key: string, editingData: EditingData): void {
+  set(key: string, editingData: EditingData): Promise<void> {
     const filePath = this.cache.set(key, JSON.stringify(editingData));
-    if (!filePath || filePath.length === 0) {
-      throw new Error(`Editing data cache not set for key ${key} at ${this.cache.root}`);
-    }
+    return new Promise((resolve) => {
+      if (!filePath || filePath.length === 0) {
+        throw new Error(`Editing data cache not set for key ${key} at ${this.cache.root}`);
```
ambrauer, line 34: Should we use `reject` here for good measure (even though it wouldn't matter in our current usage)?

ambrauer (approved):
Just one suggestion, otherwise looks good!

BTW, in a future release (currently, JSS 21 is only compatible with XM Cloud), you may be able to optimize your solution for Azure hosting even further by skipping the Data API Route + `EditingDataMiddleware` and provide an `EditingDataService` implementation instead (see `ServerlessEditingDataService` vs `BasicEditingDataService` in [XMC docs](https://doc.sitecore.com/xmc/en/developers/xm-cloud/architecture-and-apis-for-integrating-jss-next-js-apps-with-sitecore-editors.html)).

packages/sitecore-jss-nextjs/src/editing/editing-data-cache.ts
erwinsmit: Yes, it makes sense to use reject here, throw new Error will also fall into a catch, but I think it's cleaner to use reject.

erwinsmit:
> Just one suggestion, otherwise looks good!
> 
> BTW, in a future release (currently, JSS 21 is only compatible with XM Cloud), you may be able to optimize your solution for Azure hosting even further by skipping the Data API Route + `EditingDataMiddleware` and provide an `EditingDataService` implementation instead (see `ServerlessEditingDataService` vs `BasicEditingDataService` in [XMC docs](https://doc.sitecore.com/xmc/en/developers/xm-cloud/architecture-and-apis-for-integrating-jss-next-js-apps-with-sitecore-editors.html)).

Thanks, good to know!

erwinsmit:
@ambrauer Closed the PR by accident, so I've re-opened it again. The suggestion with the promise has been resolved.

PR #1232, opened 2022-11-09
 #559289: added implementation getEEMarkup into sitecore-jss-nextjs

matkovskyi:
## Types of changes


ambrauer (commented):
Was there no way to solve this (make it editor-friendly) without using `getEEMarkup`? As mentioned, this wasn't intended to be used externally... If this is the only way, then we can go with this.

matkovskyi:
> Was there no way to solve this (make it editor-friendly) without using `getEEMarkup`? As mentioned, this wasn't intended to be used externally... If this is the only way, then we can go with this.

Yeah! I removed this `getEEMarkup` and now there is using <JssImage />

ambrauer (approved):
Looks good!

PR #1233, opened 2022-11-09
add test task

erwinsmit:
## Description / Motivation
When I worked on a PR recently:
https://github.com/Sitecore/jss/pull/1223

I had an issue running all the tests from the root of the repository. In the contributing guidelines, it states that you can run all the tests using `yarn test-packages`. But there doesn't seem to be an NPM task for this.

## Types of changes
Adding a test task to the root of the repository for running tests across all the packages.


ambrauer (approved):
Good catch. This changed to "coverage-packages" since our CONTRIBUTING doc was written. While that could be used, it's nice to have an option without coverage included.

PR #1239, opened 2022-11-13
[NextJS] Proof of concept multisite solution

mjarosch:
## Description / Motivation
We host multiple websites in our Sitecore instance. When transitioning to NextJS and headless, we need to have multiple NextJS apps, one for each website. Also, our websites share enterprise components that provide brand consistency across sites, so this creates the requirement to either re-implement the shared components for each site or to have a shared component library.

## Testing Details
Still working on this.

## Types of changes


ambrauer:
@mjarosch this is great stuff. It just so happens that we are in the middle of a Next.js multi-site discovery and POC internally (with plans to introduce this capability in a future release), so your PR is quite timely :) We will definitely be looking through your PR for input. If you'd like, we can also include you on PR reviews of our own once we start implementation.

mjarosch:
@ambrauer It's great to hear this is on your roadmap. I am very interested in contributing any way I can, contributing code, PR reviews, documentation, any where I can be helpful. Please let me know.

A couple of notes about my implementation.

1. I tried to make the multisite a template addon, similar to the personalization, so that it is more opt in and reduce the number of changes to the main nextjs template.
2. I used a site config file, with the .env override mainly because the edge api doesn't have a site query where you can get all of the sites configured in Sitecore. But after thinking about it for a while, it did seem nice that you could limit what sites it responds to, and not every site configured in Sitecore.
3. I tried to keep my changes at the nextjs level and not add to much to Jss itself. This way it would reduce the amount changes that would affect other rendering options at this time. I thought reducing the change area might improve the chances of a PR getting accepted. 😃

PR #1256, opened 2022-12-07
Unit test coverage for VisitorIdentification

art-alexeyenko:
Unit test coverage for VisitorIdentification

## Description / Motivation
Defer prop has been added recently. This PR adds tests for that and other VI component functionality.

## Types of changes

### DIFF
--- a/packages/sitecore-jss-react/src/components/VisitorIdentification.test.tsx
+++ b/packages/sitecore-jss-react/src/components/VisitorIdentification.test.tsx
@@ -5,13 +5,15 @@ import { mount } from 'enzyme';
 
 import { VisitorIdentification, resetEmittedVI } from './VisitorIdentification';
 import { SitecoreContextReactContext } from './SitecoreContext';
+import { spy } from 'sinon';
 
 describe('<VisitorIdentification />', () => {
-  const defaultContext = { visitorIdentificationTimestamp: 123 };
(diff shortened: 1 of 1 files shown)


packages/sitecore-jss-react/src/components/VisitorIdentification.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  it('should render script with defer when defer is enabled', async () => {
+    const rendered = mount(
+      <SitecoreContextReactContext.Provider value={mockContext(defaultContext)}>
+        <VisitorIdentification defer={true} />
+      </SitecoreContextReactContext.Provider>
+    );
+    const script = document.getElementsByTagName('script')[0];
+    expect(script).to.not.be.equal(undefined);
+    expect(script.defer).to.equal(true);
```
ambrauer, line 53: We should check the `defer` default somewhere. Probably fine to just add an expectation to the `should render in <head>` test?

ambrauer (approved):
Looks good! Just one suggestion about additional defer coverage.

packages/sitecore-jss-react/src/components/VisitorIdentification.test.tsx
art-alexeyenko: Good idea. I've added the extra check.

PR #1262, opened 2022-12-16
sitecore-jss extra unit test coverage

art-alexeyenko:
Small increase for base package unit test coverage

- Ignore files that don't have executable logic
- Add tests where it makes sense


ambrauer (approved):
Looks good 👍

PR #1265, opened 2022-12-19
Additional unit test coverage for jss-dev-tools

art-alexeyenko:
A few extra tests for dev tools to bump coverage a little.

### DIFF
--- a/packages/sitecore-jss-dev-tools/src/resolve-scjssconfig.test.ts
+++ b/packages/sitecore-jss-dev-tools/src/resolve-scjssconfig.test.ts
@@ -1,0 +1,45 @@
+// saved for the future, when resolve can be better mocked
+// issue link: https://github.com/browserify/resolve/issues/293
+
+// /* eslint-disable @typescript-eslint/no-implicit-any */
+// import * as resolveUtils from 'resolve';
+// import { resolveScJssConfig } from './resolve-scjssconfig';
+// const Module = require('module');
+// import sinon from 'sinon';
+// import { expect } from 'chai';
+
+// describe('resolve-scjssconfig', () => {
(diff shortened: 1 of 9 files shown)


ambrauer (approved):
Looks good to me. Great work!

packages/sitecore-jss-dev-tools/src/resolve-scjssconfig.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    const mockConfigPath = './src/test-data/scjssconfig-notexists.json';
+    const resolveInput = {
+      configPath: mockConfigPath,
+      configName: 'sitecore',
+      assert: true,
+    };
+
+    try {
+      await resolveScJssConfig(resolveInput);
+      expect(true).to.be.false; // should be not reachable
```
ambrauer, line 27: Don't think this assertion is all that necessary. You could do an assertion on the `err` inside the catch if possible (though looks like this would be undefined), but the log assertion there is good enough.

packages/sitecore-jss-dev-tools/src/resolve-scjssconfig.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    const mockConfigPath = './src/test-data/scjssconfig-empty.json';
+    const resolveInput = {
+      configPath: mockConfigPath,
+      configName: 'sitecore',
+      assert: true,
+    };
+
+    try {
+      await resolveScJssConfig(resolveInput);
+      expect(true).to.be.false; // should be not reachable
```
ambrauer, line 47: Same here (not sure this is useful)

PR #1270, opened 2022-12-21
[React] Fix `eslint` compilation error

sc-addypathania:
**Issue Description**: The react plugin was conflicted between ".eslintrc" and "BaseConfig(in `react-scripts` module) i.e 2 copies are being fed to the ESLint.
**Fix**: Remove `eslint-plugin-react` and its configuration.

## Types of changes

### DIFF
--- a/packages/create-sitecore-jss/src/templates/react/.eslintrc
+++ b/packages/create-sitecore-jss/src/templates/react/.eslintrc
@@ -4,7 +4,6 @@
     "prettier",
     "plugin:import/errors",
     "plugin:import/warnings",
-    "plugin:react/recommended",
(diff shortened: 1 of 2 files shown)


ambrauer (commented):
Since we aren't using CRA's [built-in eslint config](https://create-react-app.dev/docs/setting-up-your-editor/#displaying-lint-output-in-the-editor), wouldn't we lose out on all of these rules if we remove it here?

Seems like we would want to keep this. Is the issue that we [have different `eslint-plugin-react` version than CRA](https://stackoverflow.com/questions/70449712/create-react-app-error-in-plugin-react-was-conflicted-between-eslintrc-jso)?

Issue #1291, opened 2023-01-16
With multisite enabled in 21.1-canary, the GRAPH_QL_ENDPOINT is required


larserhardsen:
## Description
In previous versions before adding the new multisite functionality, it was enough to provide the SITECORE_API_HOST and SITECORE_API_KEY environment variables, and with multisite enabled, the GRAPH_QL_ENDPOINT variable must be filled in, according to the file here:
https://github.com/Sitecore/jss/blob/dev/packages/create-sitecore-jss/src/templates/nextjs-multisite/scripts/config/plugins/multisite.ts

If omitted, the build fails, as the config.graphQLEndpoint is only containing a path.

## Expected behavior
It should be enough to provide SITECORE_API_HOST and SITECORE_API_KEY

ambrauer:
@larserhardsen Thanks for reporting, I've added this to our backlog for further investigation. You are correct, we lost previous dynamic aspects of the config `graphQLEndpoint` prop.

ambrauer:
@larserhardsen This issue is resolved with #1296. Please try with latest canary build.

larserhardsen:
@ambrauer It works! Thanks!

PR #1296, opened 2023-01-18
Ensure app config variables have values at build-time

art-alexeyenko:
This PR slightly reworks how a config in nextjs app is built. 
Each value is resolved with the following logic:
1) Attempt to retrieve value from env
2) Attempt to retrieve value from supplementary files (package.json / scjssconfig)
3) If certain values (that can't be empty, i.e. language) are still empty - use fallback value.

This roughly follows pre-plugin approach to building config where env values take precedence, other sources follow as fallback.
This also ensures that graphQL endpoint will have a correct value, if hostname values is present in either env or scjssconfig.

## Types of changes

### DIFF
--- a/packages/create-sitecore-jss/src/templates/nextjs-multisite/scripts/config/plugins/multisite.ts
+++ b/packages/create-sitecore-jss/src/templates/nextjs-multisite/scripts/config/plugins/multisite.ts
@@ -14,13 +14,12 @@ class MultisitePlugin implements ConfigPlugin {
   async exec(config: JssConfig) {
     let sites: SiteInfo[] = [];
 
-    // graphQL endpoint can have a dynamic value in the config - so we resolve it in a special way at build time
     const endpoint = config.graphQLEndpoint;
     const apiKey = config.sitecoreApiKey;
 
     if (!endpoint || !apiKey) {
       console.warn(
-        chalk.yellow('Skipping site information fetch (missing GraphQL connection details). Endpoint or API key missing.')
+        chalk.yellow('Skipping site information fetch (missing GraphQL endpoint or API key).')
       );
     } else {
       console.log(`Fetching site information from ${endpoint}`);
--- a/packages/create-sitecore-jss/src/templates/nextjs-multisite/scripts/config/plugins/multisite.ts
+++ b/packages/create-sitecore-jss/src/templates/nextjs-multisite/scripts/config/plugins/multisite.ts
@@ -24,12 +24,12 @@ class MultisitePlugin implements ConfigPlugin {
     };
 
     // graphQL endpoint can have a dynamic value in the config - so we resolve it in a special way at build time
-    const buildTimeGraphQlEndpoint = process.env.GRAPH_QL_ENDPOINT || computeConfigValue(config.computed?.graphQLEndpoint || '');
+    const endpoint = process.env.GRAPH_QL_ENDPOINT || computeConfigValue(config.computed?.graphQLEndpoint || '');
     const apiKey = process.env.SITECORE_API_KEY || config.sitecoreApiKey;
 
     if (!endpoint || !apiKey) {
       console.warn(
-        chalk.yellow('Skipping site information fetch (missing GraphQL connection details)')
+        chalk.yellow('Skipping site information fetch (missing GraphQL connection details). Endpoint or API key missing.')
       );
     } else {
       console.log(`Fetching site information from ${endpoint}`);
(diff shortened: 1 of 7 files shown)


ambrauer (changes requested):
The `computed` prop and `computeConfigValue` required to evaluate the string seems unnecessary. I think you're onto something with the fallback to env values in `scjssconfig.ts`...

Could we just make this a first step to populate all of the config defaults? 
For example, in [generate-config.ts](https://github.com/Sitecore/jss/blob/dev/packages/create-sitecore-jss/src/templates/nextjs/scripts/generate-config.ts#L12-L18), something like:

```
const defaultConfig: JssConfig = {
  sitecoreApiKey: process.env[`${constantCase('sitecoreApiKey')}`] || 'no-api-key-set',
  sitecoreApiHost: process.env[`${constantCase('sitecoreApiHost')}`] || '',
  jssAppName: process.env[`${constantCase('jssAppName')}`] || 'Unknown',
  graphQLEndpointPath: process.env[`${constantCase('graphQLEndpointPath')}`] || '',
  defaultLanguage: process.env[`${constantCase('defaultLanguage')}`] || 'en',
};
```

Or maybe even better, a `default` config plugin that does this? Could probably then lose the defaultConfig altogether in generate-config.ts.

art-alexeyenko:
Reworked the approach, moving away from dynamic evaluation.

packages/create-sitecore-jss/src/templates/nextjs-multisite/scripts/config/plugins/multisite.ts
```diff
@@ -13,12 +13,14 @@ class MultisitePlugin implements ConfigPlugin {
...
-    const apiKey = process.env.SITECORE_API_KEY || config.sitecoreApiKey;
+
+    // graphQL endpoint can have a dynamic value in the config - so we resolve it in a special way at build time
+    const endpoint = config.graphQLEndpoint;
+    const apiKey = config.sitecoreApiKey;
 
     if (!endpoint || !apiKey) {
       console.warn(
-        chalk.yellow('Skipping site information fetch (missing GraphQL connection details)')
+        chalk.yellow('Skipping site information fetch (missing GraphQL connection details). Endpoint or API key missing.')
```
ambrauer, line 23: Can simplify this I think
```suggestion
        chalk.yellow('Skipping site information fetch (missing GraphQL endpoint or API key)')
```

packages/create-sitecore-jss/src/templates/nextjs/scripts/config/plugins/fallback.ts
```diff
@@ -0,0 +1,19 @@
+import { ConfigPlugin, JssConfig } from '..';
+
+/**
+ * This config will set fallback values for properties that were left empty
+ * If neither env, nor other places had a proper value, this will ensure a fallback is set
+ */
+class ComputedPlugin implements ConfigPlugin {
```
ambrauer, line 7: Looks like c/p from computed plugin was done :) Should be "fallback" in this file.

packages/create-sitecore-jss/src/templates/nextjs/scripts/config/plugins/fallback.ts
```diff
@@ -0,0 +1,19 @@
...
+  // should always comes last
+  order = 99;
+
+  async exec(config: JssConfig) {
+    return Object.assign({}, config, {
+      defaultLanguage: config.defaultLanguage || 'en',
+      sitecoreApiKey: config.sitecoreApiKey || 'no-api-key-set'
```
ambrauer, line 14: I know we had 'no-api-key-set' before, but now questioning if there's really any reason to have this? i.e. wouldn't an empty string result in the same "invalid API key" error when making requests in the app? Fine if we want to leave it too.

packages/create-sitecore-jss/src/templates/nextjs/scripts/generate-config.ts
```diff
@@ -10,11 +10,11 @@ import { JssConfig, jssConfigFactory } from './config';
 */
 
 const defaultConfig: JssConfig = {
```
ambrauer, line 12: Could we make this more dynamic - iterate through `defaultConfig` keys and set it using `constantCase` module?

ambrauer (changes requested):
Looking good, see a few comments

packages/create-sitecore-jss/src/templates/nextjs/scripts/config/plugins/fallback.ts
art-alexeyenko: That's a good point. Technically, we'd have the same result with empty API key before all these changes too. I think we were using 'no-api-key-set' value for better visibility and clarity when error occurs?

ambrauer: Yeah, let's just leave it for now.

ambrauer (approved):
Looks good!

ambrauer:
Resolves #1291

PR #1299, opened 2023-01-20
Ensure cookie is only used in subsequent middleware plugins, after the initial resolution

art-alexeyenko:
Considering a cookie value in initial site resolution leads to some issues. This PR fixes it.
We should use the cookie in post-multisite middleware plugins, but only the things immediately available to a user (like hostname, query string, pathname etc) should affect the initial site resolution at the start of request/middleware processing.

## Types of changes

### DIFF
--- a/packages/sitecore-jss-nextjs/src/middleware/multisite-middleware.test.ts
+++ b/packages/sitecore-jss-nextjs/src/middleware/multisite-middleware.test.ts
@@ -355,45 +355,6 @@ describe('MultisiteMiddleware', () => {
       nextRewriteStub.restore();
     });
 
-    it('sc_site cookie is provided', async () => {
-      const req = createRequest({ cookieValues: { sc_site: 'foobar' } });
-
-      const res = createResponse();
-
-      const nextRewriteStub = sinon.stub(nextjs.NextResponse, 'rewrite').returns(res);
-
-      const { middleware, getSite } = createMiddleware();
-
-      const finalRes = await middleware.getHandler()(req, res);
-
-      validateDebugLog('multisite middleware start: %o', {
-        pathname: '/styleguide',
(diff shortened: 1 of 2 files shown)


ambrauer (changes requested):
This cookie-based site lookup was added to solve the problem of browsing a specific site on [Vercel Preview Deployments](https://vercel.com/docs/concepts/deployments/generated-urls). Since Vercel doesn't generate unique URLs for each domain, this would allow you to browse a specific site using the query string parameter (without having to continuously add it while browsing).

Since cookies are domain specific, this seems like more of a localhost / dev problem. Maybe we can skip this cookie usage in development only?

art-alexeyenko:
Alternative approach: resolve site via cookie only when it makes sense. Determine via function.

packages/create-sitecore-jss/src/templates/nextjs-multisite/src/lib/middleware/plugins/multisite.ts
```diff
@@ -26,6 +26,10 @@ class MultisitePlugin implements MiddlewarePlugin {
       excludeRoute: () => false,
       // This function resolves site based on hostname
       getSite: siteResolver.getByHost,
+      // This function allows resolving site from cookies, which could be useful in case of Vercel preview URLs. Accepts NextRequest.
+      useCookieResolution: () => {
+        return process.env.VERCEL_ENV === 'preview';
+      },
```
ambrauer, line 32: Can simplify
```suggestion
      // This function allows resolving site from sc_site cookie, which could be useful in case of Vercel preview URLs. Accepts NextRequest.
     useCookieResolution: () => process.env.VERCEL_ENV === 'preview',
```

ambrauer (approved):
Looks great!

PR #1302, opened 2023-01-24
[Next.js][Multi-site] Added default for graphQLEndpoint config prop

ambrauer:
## Description / Motivation
This is a follow-up to #1296, adding the same default (pulling from env var) for `graphQLEndpoint` config prop.

## Types of changes


art-alexeyenko (approved):
Yup, all good. Thanks for the follow up - I missed that prop.

PR #1303, opened 2023-01-24
[sitecore-jss] Use site name value from field for multisite

art-alexeyenko:
Use name from field instead of an item name when getting site data via GraphQL

## Types of changes


illiakovalenko (approved):
👍

PR #1315, opened 2023-01-31
[create-sitecore-jss] Incompatibility message when installing sxa and styleguide

art-alexeyenko:
Adding incompatibility message when initializing nextjs sample app with both sxa with styleguide addons.

## Types of changes


illiakovalenko (changes requested):
Looks good to me. Just add Prompts handling as it is works for Personalize add-on

PR #1320, opened 2023-02-02
Avoid FEAAS Stylesheets parsing multiple times

Inviz:
## Description / Motivation
FEAAS Components require stylesheet of their libraries to run. Multiple components from the same library can be used on a single page. To ensure that stylesheets start loading as soon as possible, I suggested including stylesheet tag for each component. However with this approach even if stylesheet is loaded once (browser deduplicates requests) it's parsed and initialized multiple times (which can be seen in chrome dev console, having css rules multiplied). 

The solution to this problem is to use `rel="preload"` which will initiate loading during page parsing, so by the time JS is loaded, the static files are already pre-fetched. We also do it for preloading HTML and JS. Combined with SSG, this makes the files load as soon as possible, and still on demand.

## Testing Details
The result can be seen on any environment using feaas component.

Example of stylesheet preloading with **cold cache:**
1 request: Preloading
2 request: Fetching, 0 bytes - served from cache
3 request: background HEAD request, checks if the cache needs to be busted (feaas does stale-while-revalidate-hot-reload strategy for handling cache). This head request is not absolutely necessary conceptually for the case of cold cache, but it's lightweight (only headers) and it's a trade-off of preloading the static file.

MicrosoftTeams-image (10)

Example with **hot cache**
1 request: 5ms - served from disk cache
2 request: 4ms - served from disk cache
3 request: background HEAD request, checks if the cache needs to be busted. Important, as it's the only way to validate the freshness of cached file with immutable policy

Screen Shot 2023-02-03 at 00 32 51

## Types of changes


Inviz:
@ambrauer I think this is ready for review. It should be non-breaking as it just changes type of preloading we're doing.

packages/sitecore-jss-react/src/components/FEaaSComponent.tsx
```diff
@@ -84,9 +84,10 @@ export const FEaaSComponent = ({ params, fields }: FEaaSComponentProps): JSX.Ele
 
   return (
     <>
-      <feaas-stylesheet library={reqProps.library} cdn={reqProps.cdn} >
-        <link rel="stylesheet" href={`${reqProps.cdn}/styles/${reqProps.library}/published.css`} />
-      </feaas-stylesheet>
+      <link rel="preload" as="style" href={`${reqProps.cdn}/styles/${reqProps.library}/published.css`} />
+      <link rel="preload" as="fetch" href={`${reqProps.cdn}/components/${reqProps.library}/${reqProps.component}/${reqProps.version}/${reqProps.revision}.html`} />
```
ambrauer, line 88: Should this be as `document`?
```suggestion
      <link rel="preload" as="document" href={`${reqProps.cdn}/components/${reqProps.library}/${reqProps.component}/${reqProps.version}/${reqProps.revision}.html`} />
```

ambrauer (approved):
@Inviz Just one question, otherwise looks good to me

packages/sitecore-jss-react/src/components/FEaaSComponent.tsx
Inviz: These type match the content type that is being requested with fetch() call. So there's actually a requirement that this type needs to be the same on both ends (in our SDK and in here). Otherwise it can cause duplicate request. So maybe document could work, but since we do fetch() call to fetch the content i tested that it does duplicate request.

Maybe i could make it work with document too, but that needs some extra testing. Maybe i can do it on monday

Inviz: Apparently as="document" doesnt even work. Still open bug in firefox, and i couldn't make it in Chrome. It's supposed to be used for iframes also, so it seems not a good match semantically

ambrauer: @Inviz Got it, fetch it is then!

Inviz:
thanks Adam

PR #1339, opened 2023-02-14
Multisite SSG should always include site prefix

art-alexeyenko:
This PR ensures GraphQLSitemapService will return paths with site prefix applied even if one site is provided in array during class initialization. This, in combination with previous multisite changes, leads to some changes in the class' behavior:
1) fetchExportSitemap and fetchSSGSitemap will return paths with site prefix applied by default;
2) pass siteName into the methods to only return paths for a specific single site, without prefix.

nextjs and nextjs-multisite templates were updated to accommodate the change.

## Testing Details
Manual testing with extra unit tests.

## Types of changes
- [?] Breaking change (fix or feature that would cause existing functionality to change)


packages/create-sitecore-jss/src/templates/nextjs-multisite/src/lib/sitemap-fetcher/plugins/graphql-sitemap-service.ts
```diff
@@ -0,0 +1,34 @@
...
+import { siteResolver } from 'lib/site-resolver';
+
+class GraphqlSitemapServicePlugin implements SitemapFetcherPlugin {
+  _graphqlSitemapService: GraphQLSitemapService;
+
+  constructor() {
+    this._graphqlSitemapService = new GraphQLSitemapService({
+      endpoint: config.graphQLEndpoint,
+      apiKey: config.sitecoreApiKey,
+      sites: siteResolver.sites.map((site: SiteInfo) => site.name),
```
ambrauer, line 19: We should select distinct here (in case the default/config site is the same as a fetched site).

packages/create-sitecore-jss/src/templates/nextjs-multisite/src/lib/sitemap-fetcher/plugins/graphql-sitemap-service.ts
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  async exec(context?: GetStaticPathsContext): Promise<StaticPath[]> {
+    if (process.env.EXPORT_MODE) {
+      // Disconnected Export mode
+      if (process.env.JSS_MODE !== constants.JSS_MODE.DISCONNECTED) {
+        return this._graphqlSitemapService.fetchExportSitemap(config.defaultLanguage);
+      }
+    }
+    return this._graphqlSitemapService.fetchSSGSitemap(context?.locales || []);
```
ambrauer, line 30: Noticed this comment is invalid. Let's take the opportunity to clean up, maybe even simplify (in both copies of `graphql-sitemap-service.ts`).
```suggestion
    if (process.env.JSS_MODE === constants.JSS_MODE.DISCONNECTED) {
      return [];
    }
    return process.env.EXPORT_MODE
      ? this._graphqlSitemapService.fetchExportSitemap(config.defaultLanguage)
      : this._graphqlSitemapService.fetchSSGSitemap(context?.locales || []);
```

packages/create-sitecore-jss/src/templates/nextjs/src/lib/sitemap-fetcher/plugins/graphql-sitemap-service.ts
```diff
@@ -24,10 +24,13 @@ class GraphqlSitemapServicePlugin implements SitemapFetcherPlugin {
     if (process.env.EXPORT_MODE) {
```
ambrauer, line 24: We should remove the `sites` config above

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Our versioning strategy is as follows:
   * `import { editingDataService } from '@sitecore-jss/sitecore-jss-nextjs/editing';`
   * `import { EditingRenderMiddleware } from '@sitecore-jss/sitecore-jss-nextjs/editing';`
 
+* `[sitecore-jss-nextjs]` GraphqlSitemapService's constructor accepts a sites array by default, instead of siteName. fetchExportSitemap and fetchSSGSitemap methods will now return paths with site prefix applies by default. Provide a siteName parameter into the methods to get paths for a single site without prefix.
```
ambrauer, line 22: Since we are now duplicating `graphql-sitemap-service.ts`, feel like we could we avoid the breaking change by an alternative design. 
1. Have both `siteName` (existing) and `sites` (new) as options on `GraphQLSitemapServiceConfig`, both of these are optional
2. Validate 1 (and only 1) of these are set in constructor
3. Base `next` template (single-site) will set `siteName` (as it does today); `nextjs-multisite` will set `sites`
4. Can then determine site prefix formatting based on config option used

Would avoid breaking change and might be more clear since determination is all based on service config vs combination of that + functions (w/ optional siteName param).

ambrauer (changes requested):

packages/create-sitecore-jss/src/templates/nextjs-multisite/src/lib/sitemap-fetcher/plugins/graphql-sitemap-service.ts
art-alexeyenko: I think this should be handled by siteResolver. I can modify siteResolver logic in scope of this PR to have a distinct collection returned in case of Sitecore site and jssAppName are the same?

CHANGELOG.md
art-alexeyenko: > Have both siteName (existing) and sites (new) as options on GraphQLSitemapServiceConfig, both of these are optional
I have thought of several ways to combine multi and single site functionality, this one included. I think this would not be an optimal solution, even though it helps us avoid breaking changes. The entire behavior of a class changes based on constructor value. It makes using the class more complicated, less clean. If an instance of a class is passed several times between methods/other classes, you are never sure what behavior you will get without looking up the code.
I believe the approach of "Set default behavior for a class, modify behavior per method, via parameters" is cleaner, more intuitive. Multisite is a big change in itself, I think it could become the default way of using JSS going forward, and the breaking change here could be justified.

ambrauer: That's the thing, it remains to be seen if multisite will be the norm. A rendering host per site is still valid and in many ways might be preferred. I see your point, but the same argument can be made for this current solution as well. It's confusing either way ☺. 

Wonder if we should really be introducing another service altogether for multisite (e.g. `GraphQLMultiSitemapService`)? Then it's 100% clear on what you're getting and APIs are catered towards each.

packages/create-sitecore-jss/src/templates/nextjs-multisite/src/lib/sitemap-fetcher/plugins/graphql-sitemap-service.ts
ambrauer: Having duplicate site names is valid tho for the site resolver. This is only a problem here.

art-alexeyenko: should have distinct filtering now - please check if this approach is OK

CHANGELOG.md
art-alexeyenko: Refactored it into two classes - please see if this is OK.

packages/create-sitecore-jss/src/templates/nextjs-multisite/src/lib/sitemap-fetcher/plugins/graphql-sitemap-service.ts
```diff
@@ -0,0 +1,36 @@
...
+  async exec(context?: GetStaticPathsContext): Promise<StaticPath[]> {
+    if (process.env.EXPORT_MODE) {
+      if (process.env.JSS_MODE === constants.JSS_MODE.DISCONNECTED) {
+        return [];
+      }
+      return process.env.EXPORT_MODE
+        ? this._graphqlSitemapService.fetchExportSitemap(config.defaultLanguage)
+        : this._graphqlSitemapService.fetchSSGSitemap(context?.locales || []);
+    }
+    return this._graphqlSitemapService.fetchSSGSitemap(context?.locales || []);
```
ambrauer, line 32: Some redundant logic here (and in `nextjs` template's version). Should be able to have just the following, right?

```suggestion
    if (process.env.JSS_MODE === constants.JSS_MODE.DISCONNECTED) {
      return [];
    }
    return process.env.EXPORT_MODE
      ? this._graphqlSitemapService.fetchExportSitemap(config.defaultLanguage)
      : this._graphqlSitemapService.fetchSSGSitemap(context?.locales || []);
```

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -138,9 +137,10 @@ export interface GraphQLSitemapServiceConfig
   apiKey: string;
 
   /**
-   * Names of the configured sites
+   * Name of the site to retrieve site paths for
+   * Make it optional
    */
-  sites: string[];
+  siteName?: string;
```
ambrauer, line 143: This should be back to being required for this service (can just remove `siteName` from the `Omit` above). We can probably drop the check/RangeError as well? Guess it would at least catch empty string.

packages/sitecore-jss-nextjs/src/services/mutisite-graphql-sitemap-service.ts
```diff
@@ -0,0 +1,91 @@
...
+  RouteListQueryResult,
+  StaticPath,
+} from './graphql-sitemap-service';
+
+export const sitesError = 'The list of sites cannot be empty';
+
+/**
+ * Configuration options for @see GraphQLSitemapService instances
+ */
+export interface MultisiteGraphQLSitemapServiceConfig extends GraphQLSitemapServiceConfig {
```
ambrauer, line 15: We can omit the siteName here.

packages/sitecore-jss-nextjs/src/services/mutisite-graphql-sitemap-service.ts
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Service that fetches the list of site pages using Sitecore's GraphQL API.
+ * Used to handle multiple sites
+ * This list is used for SSG and Export functionality.
+ * @mixes SearchQueryService<PageListQueryResult>
+ */
+export class MultisiteGraphQLSitemapService extends GraphQLSitemapService {
+  /**
+   * Creates an instance of graphQL sitemap service with the provided options
+   * @param {GraphQLSitemapServiceConfig} options instance
```
ambrauer, line 31: This type should be updated (and possibly the description on line above)
```suggestion
   * @param {MultisiteGraphQLSitemapServiceConfig} options instance
```

CHANGELOG.md
ambrauer: I like it 👍. The two classes seems better to me. Not much duplication with the class extension.

ambrauer (changes requested):
Looking good, just a few minor things noticed

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
art-alexeyenko: I made this optional, so that multisite service constructor does not require a siteName, uses only the sites array instead. Is there a better way to use single site for parent class (graphql-sitemap-service) and only use site array for child (multisite)?

CHANGELOG.md
ambrauer, line 22: Can remove this, right (no longer a breaking change)?

packages/sitecore-jss-nextjs/src/services/graphql-sitemap-service.ts
```diff
@@ -15,138 +15,15 @@ export function getSiteEmptyError(siteName: string) {
...
 /**
  * Configuration options for @see GraphQLSitemapService instances
  */
-export interface GraphQLSitemapServiceConfig
-  extends Omit<SiteRouteQueryVariables, 'language' | 'siteName'> {
+export interface GraphQLSitemapServiceConfig extends BaseGraphQLSitemapServiceConfig {
   /**
-   * Your Graphql endpoint
+   * Name of the site to retrieve site paths for
+   * Make it optional
```
ambrauer, line 24: Remove this comment?
```suggestion
```

ambrauer (changes requested):
Awesome! Just a few minor things noticed remaining.

ambrauer (approved):
👍

PR #1349, opened 2023-02-17
[sitecore-jss-react] Fix rendering issue in components using withPlaceholder

sc-addypathania:
This fixes the rendering issueWhen additional params are configured in sitecore, the the components using the withPlaceholder HOC crashes because the current implementation of PlaceholderCommon assumes that the params would always be present on any rendering which is not the case as it is an optional property.

## Types of changes

### DIFF
--- a/packages/sitecore-jss-react/src/enhancers/withPlaceholder.test.tsx
+++ b/packages/sitecore-jss-react/src/enhancers/withPlaceholder.test.tsx
@@ -212,7 +212,6 @@ describe('withPlaceholder HOC', () => {
         const props: PlaceholderProps = {
           name: phKey,
           rendering: component,
-          params: { test: 'test' },
           componentFactory: componentFactory,
         };
         const Element = withPlaceholder(phKey)(Home);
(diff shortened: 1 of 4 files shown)


packages/sitecore-jss-react/src/enhancers/withPlaceholder.test.tsx
```diff
@@ -193,4 +200,29 @@ describe('withPlaceholder HOC', () => {
...
+      it('should render a placeholder with given key when params are not passed', () => {
+        const component = (dataSet.data.sitecore.route.placeholders.main as (
+          | ComponentRendering
+          | RouteData
+        )[]).find((c) => (c as ComponentRendering).componentName);
+        const phKey = 'page-content';
+        const props: PlaceholderProps = {
+          name: phKey,
+          rendering: component,
+          params: { test: 'test' },
```
art-alexeyenko, line 215: params are passed here for this test. Is this an omission?

art-alexeyenko (commented):
Looks like a good solution!
Just one small comment below.

packages/sitecore-jss-react/src/enhancers/withPlaceholder.test.tsx
sc-addypathania: Ohh, good catch.
This got left out when I was doing initial testing. 
Removed :)

art-alexeyenko (approved):
👍

PR #1380, opened 2023-03-15
[nextjs] Prefer wildcard fallback for point of sale resolution

art-alexeyenko:

## Description / Motivation
Point of sale would default to site language value over wildcard fallback value before.
This PR ensures wildcard is used as first fallback.

## Testing Details
- [x] Unit Test Added (amended)


ambrauer (approved):
👍

PR #1427, opened 2023-04-10
[Nextjs] [RAV] Pre-push hook for lint check

sc-addypathania:
Developers may find it frustrating when developing locally, everything works fine (because they are using next dev), but when pushing code they find that next build fails because of TypeScript or linting errors.
This hook help prevent surprise build errors (TypeScript/ESLint/Prettier) on XM Cloud / Vercel.

- [ ] Changelog updated

## Testing Details

- [x] Unit Test Added

## Types of changes


ambrauer (commented):
Any way we can do this without adding a dependency on husky? Similar to what we do with this monorepo itself:
https://github.com/Sitecore/jss/blob/dev/scripts/install-hooks.js

husky is a great tool, but maybe a bit too opinionated for our starters?

sc-addypathania:
> Any way we can do this without adding a dependency on husky? Similar to what we do with this monorepo itself: https://github.com/Sitecore/jss/blob/dev/scripts/install-hooks.js
> 
> husky is a great tool, but maybe a bit too opinionated for our starters?

Husky is widely used though and a popular tool in community but definitely better to use something of our own rather than adding an extra dependency. :)

packages/create-sitecore-jss/src/bin.test.ts
```diff
@@ -63,7 +66,16 @@ describe('bin', () => {
...
-      process.argv = ['node', 'index.ts', 'foo,bar', '--appPrefix', '--appName', 'test'];
+      process.argv = [
+        'node',
+        'index.ts',
+        'foo,bar',
+        '--appPrefix',
+        '--appName',
+        'test',
+        '--preCommitHook',
+        'yes',
```
ambrauer, line 77: Can we make `--preCommitHook` a boolean/flag type (instead of string)?

packages/create-sitecore-jss/src/bin.ts
```diff
@@ -98,6 +113,23 @@ export const main = async (args: ParsedArgs) => {
     args.force = true;
   }
 
+  let preCommitHook = args.preCommitHook;
+  if (!preCommitHook) {
```
ambrauer, line 117: If `preCommitHook` is a string, then we'd need to check for 'no' value here as well.

packages/create-sitecore-jss/src/bin.ts
```diff
@@ -98,6 +113,23 @@ export const main = async (args: ParsedArgs) => {
...
+    if (args.yes) {
+      preCommitHook = DEFAULT_PRECOMMITHOOK;
+    } else {
+      const answer = await inquirer.prompt({
+        type: 'list',
```
ambrauer, line 122: I think we can use `type: 'confirm'` for a yes/no prompt, which then uses a boolean value. I think this might then resolve my previous 2 comments.

packages/create-sitecore-jss/src/init-runner.ts
```diff
@@ -35,12 +36,22 @@ export const initRunner = async (initializers: string[], args: BaseArgs) => {
...
     installPackages(args.destination, args.silent);
     lintFix(args.destination, args.silent);
   }
+
+  // if you opt in to use the pre-commit hook for linting check,
+  // we need to initialize a git repository and install the hook locally.
+  if (args.preCommitHook === 'yes') {
+    exec(`cd ${args.destination} && git init && npm run install-pre-commit-hook`, (err: Error) => {
+      if (err) {
+        console.log('Error:', err);
```
ambrauer, line 50: We might want to provide more context for this error in the message, and maybe de-escalate to a warning (not critical to the initializer overall). Something like:
```suggestion
        console.log(chalk.yellow(`Warning: Pre-commit hook may not be working due to error `${err}`));
  );
```

packages/create-sitecore-jss/src/templates/angular/package.json
```diff
@@ -34,7 +34,9 @@
     "build:server": "cross-env-shell ng run <%- appName %>:server:production --output-path=$npm_package_config_buildArtifactsPath/server --output-hashing=none --bundle-dependencies true",
     "postbuild:server": "move-cli ./dist/server/main.js ./dist/server.bundle.js && del-cli ./dist/server",
     "bootstrap": "ts-node --project src/tsconfig.webpack-server.json scripts/bootstrap.ts",
-    "graphql:update": "ts-node --project src/tsconfig.webpack-server.json ./scripts/update-graphql-fragment-data.ts"
+    "graphql:update": "ts-node --project src/tsconfig.webpack-server.json ./scripts/update-graphql-fragment-data.ts",
+    "install-pre-commit-hook": "node ./scripts/install-pre-commit-hook.js",
+    "pre-commit": "yarn lint --fix"
```
ambrauer, line 39: Suggest to remove the dash to follow [npm scripts lifecycle](https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-scripts) naming convention (just in case we want to come up with "commit" or "postcommit" actions).
```suggestion
    "precommit": "yarn lint --fix"
```

packages/create-sitecore-jss/src/templates/angular/scripts/install-pre-commit-hook.js
```diff
@@ -0,0 +1,28 @@
+﻿const fs = require('fs');
+
+const installHooks = () => {
+  // data to be written to the file
+  const data = `#!/bin/sh
+
+#
+
```
ambrauer, line 8: Remove extra spaces in comment?

packages/create-sitecore-jss/src/templates/angular/scripts/install-pre-commit-hook.js
```diff
@@ -0,0 +1,28 @@
...
+# pre-commit hook that runs our linter before every commit
+
+#
+
+echo "Running lint check..."
+yarn pre-commit`;
```
ambrauer, line 14: Switch to "precommit" name, and we should use npm (our current assumed package manager)
```suggestion
npm run precommit`;
```

ambrauer (changes requested):
See some comments. Also wondering if we want to go with pre-push instead of pre-commit (maybe discuss with the team)? I forgot we used pre-push in our monorepo.

packages/create-sitecore-jss/src/common/args/base.ts
```diff
@@ -29,6 +29,10 @@ export type BaseArgs = {
    * Default values will be used
    */
   yes?: boolean;
+  /**
+   * Pre-push hook for linting check
+   */
+  prePush?: boolean;
```
ambrauer, line 35: Do you think "prePush" is too generic? I kinda liked your original name with "Hook" at the end. 
```suggestion
  prePushHook?: boolean;
```

packages/create-sitecore-jss/src/templates/angular/package.json
```diff
@@ -34,7 +34,9 @@
     "build:server": "cross-env-shell ng run <%- appName %>:server:production --output-path=$npm_package_config_buildArtifactsPath/server --output-hashing=none --bundle-dependencies true",
     "postbuild:server": "move-cli ./dist/server/main.js ./dist/server.bundle.js && del-cli ./dist/server",
     "bootstrap": "ts-node --project src/tsconfig.webpack-server.json scripts/bootstrap.ts",
-    "graphql:update": "ts-node --project src/tsconfig.webpack-server.json ./scripts/update-graphql-fragment-data.ts"
+    "graphql:update": "ts-node --project src/tsconfig.webpack-server.json ./scripts/update-graphql-fragment-data.ts",
+    "install-pre-push-hook": "node ./scripts/install-pre-push-hook.js",
+    "prepush": "yarn lint --fix"
```
ambrauer, line 39: Do we need this additional 'prepush' command with the --fix or can we just call the existing 'lint' command? If we do keep this, would need to use npm.

packages/create-sitecore-jss/src/templates/angular/scripts/install-pre-push-hook.js
```diff
@@ -0,0 +1,24 @@
+﻿const fs = require('fs');
+
+const installHooks = () => {
+  // data to be written to the file
+  const data = `#!/bin/sh
+#
+# pre-push hook runs our linter before we push our changes
+#
```
ambrauer, line 8: Would be nice to mention --no-verify option to bypass like we do here: https://github.com/Sitecore/jss/blob/dev/scripts/install-hooks.js#L10-L11

packages/create-sitecore-jss/src/common/args/base.ts
sc-addypathania: Honestly, I was a bit confused about the naming. I thought `prePushHook` would be long and hard to remember but i guess it makes sense to add hook in the end for more clarity.

packages/create-sitecore-jss/src/templates/angular/package.json
sc-addypathania: hmm yeah right we don't need it. WE can directly use the lint script.

ambrauer (approved):
Just one final thing I noticed, otherwise approved. Looks great 👍

packages/create-sitecore-jss/src/bin.ts
```diff
@@ -98,6 +98,17 @@ export const main = async (args: ParsedArgs) => {
     args.force = true;
   }
 
+  if (args.yes) {
+    args.prePushHook = true;
```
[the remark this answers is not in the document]
sc-addypathania: This is when we pass the `--yes` flag then `prePushHook` should be set to `true`

packages/create-sitecore-jss/src/bin.ts
```diff
@@ -98,6 +98,17 @@ export const main = async (args: ParsedArgs) => {
...
+  } else {
+    const answer = await inquirer.prompt({
+      type: 'confirm',
+      name: 'prePushHook',
+      message: 'Would you like to use the pre-push hook for linting check?',
```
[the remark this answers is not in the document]
sc-addypathania: Yes default is true (opt-in)

packages/create-sitecore-jss/src/bin.ts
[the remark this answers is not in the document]
ambrauer: Hmm.. I guess we haven't had to consider the `--yes` flag for boolean/flag arguments yet, but here's my thinking:

- Any boolean/flag argument is `false` by default - it's `true` only when present
- The `--yes` flag suppresses prompting and uses default values
- So, unless the boolean/flag argument is provided, `--yes` should use `false` for its default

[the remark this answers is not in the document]
sc-addypathania: Ohh okay I get it, it's clear now.
In that case removed the extra `if`.

PR #1428, opened 2023-04-11
[React] Fix Compilation Error

sc-addypathania:
`eslint-config-react-app` in `react-scripts` already have the dependency `eslint-plugin-react` therefore because of the duplicate dependency and mismatch in the version causes this compilation error.
The workaround to resolve this error is to remove the `yarn.lock` and regenerate again.

- [x] Changelog updated

## Testing Details

- [ ] Unit Test Added

## Types of changes


CHANGELOG.md
```diff
@@ -25,6 +22,7 @@ Our versioning strategy is as follows:
 ### 🐛 Bug Fixes
 
 * `[sitecore-jss-nextjs]` Fix for Link component which throws error if field is undefined ([#1425](https://github.com/Sitecore/jss/pull/1425))
+* `[templates/react]` Fix compilation error ([#1428](https://github.com/Sitecore/jss/pull/1428))
```
ambrauer, line 25: Might help to clarify that this is only in monorepo.
```suggestion
* `[templates/react]` Fix compilation error when using React template in monorepo ([#1428](https://github.com/Sitecore/jss/pull/1428))
```

packages/create-sitecore-jss/scripts/watch-templates.ts
```diff
@@ -49,6 +81,9 @@ const initializeApps = async (noInstall: boolean) => {
     if (watch.args.restoreLockfile) {
```
ambrauer, line 81: We probably wouldn't need to restore lock file if we are regenerating. Seems like regenerating would take precedence.

packages/create-sitecore-jss/scripts/watch-templates.ts
```diff
@@ -40,6 +43,35 @@ function restoreLockfile() {
   }
 }
 
+/**
+ * Removes and then regenerates the yarn.lock file in order to resolve conflicting dependencies for react template.
+ */
+async function regenerateYarnLock() {
```
ambrauer, line 49: Rename to match `restoreLockFile`?
```suggestion
async function regenerateLockFile() {
```

packages/create-sitecore-jss/scripts/watch-templates.ts
```diff
@@ -49,6 +81,9 @@ const initializeApps = async (noInstall: boolean) => {
     if (watch.args.restoreLockfile) {
       restoreLockfile();
     }
+    if (initializers.includes('react')) {
+      await regenerateYarnLock();
```
ambrauer, line 85: Looks like rename to regenerateLockFile was missed here?

packages/create-sitecore-jss/scripts/watch-templates.ts
ambrauer: Right, this scenario I mentioned would only affect react and wouldn't really break anything would just be unnecessary work (i.e. if you had restoreLockFile set when initializing react, it would restore the lock file then delete and regen it).

ambrauer: So, I suppose restoreLockFile wouldn't really do much for react then? Which, honestly, I think is OK. Let's leave it as-is :)

ambrauer (approved):
Just one last missed rename, otherwise looks good to me 👍

PR #1454, opened 2023-04-26
﻿ #534241: [sitecore-jss-nextjs][middleware-redirect]added handling with token $siteLang

matkovskyi:
- [x] Changelog updated

## Description / Motivation

Added a new handling for token $siteLang(context site language) in middleware redirect

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/create-sitecore-jss/src/templates/nextjs-sxa/src/lib/middleware/plugins/redirects.ts
```diff
@@ -24,6 +24,8 @@ class RedirectsPlugin implements MiddlewarePlugin {
       disabled: () => process.env.NODE_ENV === 'development',
       // Site resolver implementation
       siteResolver,
+      // This list of sites need for searching context site language when we use token like $siteLang it's not locale of site
+      sites: JSON.parse(config.sites),
```
ambrauer, line 28: No need to add this config. `SiteResolver` class has a readonly `sites` prop you can use (i.e. use `this.config.siteResolver.sites`).

Here's an example where we use it:
https://github.com/Sitecore/jss/blob/dev/packages/create-sitecore-jss/src/templates/nextjs-multisite/src/lib/sitemap-fetcher/plugins/graphql-sitemap-service.ts#L19

packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.ts
```diff
@@ -86,6 +92,22 @@ export class RedirectsMiddleware extends MiddlewareBase {
...
+        );
+
+        if (!siteInfo) {
+          throw new Error(`Could not resolve context site language for site ${site.name}`);
+        }
+
+        existsRedirect.target = existsRedirect.target.replace(
+          REGEXP_CONTEXT_SITE_LANG,
+          siteInfo?.language
+        );
```
ambrauer, line 108: I don't follow why we need this extra `siteInfo` variable and site lookup. Don't we already have this with the `site` variable (above)?
```suggestion
        existsRedirect.target = existsRedirect.target.replace(
          REGEXP_CONTEXT_SITE_LANG,
          site.language
        );
```

packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.ts
```diff
@@ -86,6 +92,22 @@ export class RedirectsMiddleware extends MiddlewareBase {
...
+      }
+
```
ambrauer, line 110: Unrelated, we may want to consider adding the same [try/catch in `getHandler` setup](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss-nextjs/src/middleware/multisite-middleware.ts#L28-L38) we have for error handling in other middleware. Noticed we didn't apply that to this middleware.

ambrauer (changes requested):

packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.ts
matkovskyi: I wanted to use this `siteInfo` but I need get language from config file because in `siteInfo` is property `language` but it's from route or default language(EN). SXA has context site language where you can set in Settings of site and this parameter is not related to routing
image

packages/create-sitecore-jss/src/templates/nextjs-sxa/src/lib/middleware/plugins/redirects.ts
matkovskyi: Thank you

packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.ts
ambrauer: Ah. This is more a local development configuration problem. The resolved `site` is probably the default so has the default language (en). If you set the "emptysite" host name field in Sitecore to have "localhost", it'll then resolve to that one and have the da language you're expecting in the middleware. 

Check out the [docs SiteResolver section](https://doc.sitecore.com/xmc/en/developers/xm-cloud/the-next-js-multisite-add-on.html#:~:text=the%20sites%20property.-,SiteResolver,-The%20SiteResolver%20class) for a better explanation.

The middleware (and anywhere else that uses site resolver) needs to be able to trust / use the resolved site.

packages/create-sitecore-jss/src/templates/nextjs-sxa/src/lib/site-resolver/plugins/default.ts
```diff
@@ -0,0 +1,14 @@
+import { SiteInfo } from '@sitecore-jss/sitecore-jss-nextjs/site';
+import config from 'temp/config';
+import { SiteResolverPlugin } from '..';
+
+class DefaultPlugin implements SiteResolverPlugin {
+  exec(sites: SiteInfo[]): SiteInfo[] {
+    /**
+     * remove default logic. it breaks for getting site info about context site language
```
ambrauer, line 8: @matkovskyi Think we're still not on the same page 🤔 We shouldn't need to remove default plugin logic. Site resolution/context is working as expected. Let's discuss again.

ambrauer (changes requested):
Also, please check linting, the check is currently failing.

ambrauer (approved):
Looks good (approved), but I do have a question about the $siteLang token in general. If the intent is to always replace with the context site's _default_ language, then this satisfies the requirements. However, I imagine it's more useful to replace with the _context_ language (i.e. from NextUrl.locale). 

For example, if I have a single site (default language = 'en') with 2 language versions ('en' and 'da-DK') and am using Next.js i18n sub-path routing, I would want to be able to configure a redirect target like:

pattern: `/not-found` > target: `http://mysite.com/$contextLang/found`

So that this would be the result:
http://mysite.com/en/not-found >http://mysite.com/en/found
http://mysite.com/da-DK/not-found >http://mysite.com/da-DK/found

With $siteLang, this would be the result:
http://mysite.com/en/not-found >http://mysite.com/en/found
http://mysite.com/da-DK/not-found >http://mysite.com/en/found

matkovskyi:
> Looks good (approved), but I do have a question about the $siteLang token in general. If the intent is to always replace with the context site's _default_ language, then this satisfies the requirements. However, I imagine it's more useful to replace with the _context_ language (i.e. from NextUrl.locale).
> 
> For example, if I have a single site (default language = 'en') with 2 language versions ('en' and 'da-DK') and am using Next.js i18n sub-path routing, I would want to be able to configure a redirect target like:
> 
> pattern: `/not-found` > target: `http://mysite.com/$contextLang/found`
> 
> So that this would be the result: http://mysite.com/en/not-found >http://mysite.com/en/found http://mysite.com/da-DK/not-found >http://mysite.com/da-DK/found
> 
> With $siteLang, this would be the result: http://mysite.com/en/not-found >http://mysite.com/en/found http://mysite.com/da-DK/not-found >http://mysite.com/en/found

Yeah, but in this situation we not use token $siteLang it will work the same without token, right?

ambrauer:
> Yeah, but in this situation we not use token $siteLang it will work the same without token, right?

Oh really? Just to confirm... You're saying that if I have this redirect configured:

pattern: /not-found > target: http://mysite.com/found

then these requests would redirect as follows?
http://mysite.com/en/not-found -> http://mysite.com/en/found
http://mysite.com/da-DK/not-found -> http://mysite.com/da-DK/found

Guess I'm just trying to think of the business value in being able to redirect to the site's configured _default_ language (vs the route's context language).

matkovskyi:
> > Yeah, but in this situation we not use token $siteLang it will work the same without token, right?
> 
> Oh really? Just to confirm... You're saying that if I have this redirect configured:
> 
> pattern: /not-found > target: http://mysite.com/found
> 
> then these requests would redirect as follows? http://mysite.com/en/not-found -> http://mysite.com/en/found http://mysite.com/da-DK/not-found -> http://mysite.com/da-DK/found
> 
> Guess I'm just trying to think of the business value in being able to redirect to the site's configured _default_ language (vs the route's context language).

I checked it and it works so as you write

/not-found -> /redirected-link
/da/not-found /da/redirected-link

PR #1455, opened 2023-04-26
Internal dev/16/fix pipeline bootstrap

art-alexeyenko:
Part 2 of fixing ver16 pipeline: 
* Ensure node version is set to the supported v14
* Fix an SSH dependency for react-native so that npm ci can do its thing without access errors
* Add extra condition so that pre-release will publish when pushing to release branch


sc-addypathania (approved):
Looks good

PR #1458, opened 2023-04-27
[internal][v16] Fix branch checkout in azure pipeline

art-alexeyenko:
Part 3 of pipeline adjustments. This one ensures pipeline checks out to the existing and correct branch name before performing lerna version and lerna publish actions.


sc-addypathania (approved):
👍🏼

PR #1472, opened 2023-05-05
[nextjs][RAV] Install `pre-push-hook` to root .git

sc-addypathania:
- Install the the `prePushHook` script to the root .git folder.
- Currently the --yes flag defaults prePushHook to true because of which if prePushHook flag is passed explicitly then it is not considered. We want the cli to accept both flags`--yes` and `prePushHook`.

- [x] Changelog updated

## Testing Details

- [ ] Unit Test Added

## Types of changes

### DIFF
--- a/packages/create-sitecore-jss/src/templates/angular/scripts/install-pre-push-hook.ts
+++ b/packages/create-sitecore-jss/src/templates/angular/scripts/install-pre-push-hook.ts
@@ -2,14 +2,16 @@
 import * as path from 'path';
 import { promisify } from 'util';
 import { exec } from 'child_process';
-import chalk from 'chalk';
+
+const greenLog = '\x1b[32m%s\x1b[0m';
+const redLog = '\x1b[31m%o\x1b[0m';
 
 const installHooks = async () => {
   try {
     const appPath = path.join(__dirname, '..').replace(/\\/g, '/');
     const { stdout } = await promisify(exec)('git rev-parse --show-toplevel');
     const gitRootPath = stdout.trim();
-    console.log(chalk.green(`Writing data to local .git folder ${gitRootPath}...`));
+    console.log(greenLog, 'Writing data to local .git folder...');
 
     const data = `#!/bin/sh
 #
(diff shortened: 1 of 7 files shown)


packages/create-sitecore-jss/src/templates/angular/scripts/install-pre-push-hook.ts
```diff
@@ -1,29 +1,34 @@
...
-  console.log('\x1b[32m%s\x1b[0m', 'Writing to local .git folder...');
+    const data = `#!/bin/sh
+    # pre-push hook runs our linter before we push our changes
+    #
+    # To skip this hook, use the --no-verify flag
+    # when pushing.
+    echo "Running lint check..."
+    cd ${appPath}
+    npm run lint
+  `;
```
ambrauer, line 25: Do we need to remove indentation here (so that it does not show up in resulting file)?

ambrauer (changes requested):

packages/create-sitecore-jss/src/templates/angular/scripts/install-pre-push-hook.ts
```diff
@@ -1,29 +1,34 @@
...
-  // data to be written to the file
-  const data = `#!/bin/sh
-#
-# pre-push hook runs our linter before we push our changes
-#
-# To skip this hook, use the --no-verify flag
-# when pushing.
-#
+const gitRootDirMessage = '\x1b[32m%s\x1b[0m';
+const successMessage = '\x1b[32m%s\x1b[0m';
```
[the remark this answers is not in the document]
sc-addypathania: replaced it with chalk.

packages/create-sitecore-jss/src/templates/vue/scripts/install-pre-push-hook.js
```diff
@@ -1,28 +1,32 @@
-﻿const fs = require('fs');
+﻿﻿const fs = require('fs');
+const path = require('path');
+const { promisify } = require('util');
+const { exec } = require('child_process');
+const chalk = require('chalk');
```
ambrauer, line 5: I think vue is the only one without chalk. Would need to add it to devDependencies.

ambrauer (approved):
Looks good! Just the chalk dependency for vue needs to be added. Otherwise, approved.

kendoce (approved):
Passed QA

PR #1497, opened 2023-05-24
[nextjs] Re-introduce babelrc

art-alexeyenko:
The Wheel of Time weaves as the Wheel wills and we again need to add babelrc after the nextjs version update.
This should be a temporary measure to make our test agents work again, while we look for a better solution.

### DIFF
--- a/packages/create-sitecore-jss/src/templates/nextjs/.babelrc
+++ b/packages/create-sitecore-jss/src/templates/nextjs/.babelrc
@@ -1,0 +1,3 @@
+{
+    "presets": ["next/babel"]
+}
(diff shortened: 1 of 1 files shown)


sc-addypathania (approved):
👍🏼

packages/create-sitecore-jss/src/templates/nextjs/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
```
ambrauer, line 1: We should add the same comments we had back as well

art-alexeyenko: added

PR #1508, opened 2023-06-02
Revert "[nextjs] Re-introduce babelrc"

art-alexeyenko:
Now that we added C++ Redist to our internal containers we can revert Sitecore/jss#1497


ambrauer (approved):
👍 goodbye for good??

PR #1530, opened 2023-06-19
[nextjs] Add Vercel KV support for editing cache

art-alexeyenko:
This PR adds a Vercel KV editing data cache implementation, to avoid editing errors when Vercel deployment is used as editing host.

- [x] Changelog updated

## Description / Motivation
Default disk-based cache used for editing conflicts with the stateless nature of Vercel workers. For a more persistent storage we add support for KV and an example of how it can be enabled in an app.

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/sitecore-jss-nextjs/src/editing/vercel-editng-data-cache.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+  }
+
+  get(key: string): Promise<EditingData | undefined> {
+    debug.editing(`Getting editing data for ${key} from redis storage...`);
+    return new Promise<EditingData | undefined>((resolve, reject) => {
+      this.redisCache
+        .get(key)
+        .then((entry) => {
+          // We need to normalize the object we get from API then JSON-ify it, won't work otherwise
```
ambrauer, line 53: Interesting that we need to do extra `JSON.stringify` here. Is this a known issue with Redis/Upstash storage of JSON strings?

packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/editing/data/[key].ts
Comment (ambrauer) [inline packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/editing/data/[key].ts:1]: We should actually be able to use the `BasicEditingDataService`, so you wouldn't need to make changes in this file. 

Maybe we can mention in the appropriate starter file that the `VercelEditingDataCache` is an option, but for the most part feel like we should leave the code snippets and instructions up to the docs?

ambrauer (changes requested):

packages/sitecore-jss-nextjs/src/editing/vercel-editng-data-cache.ts
```diff
@@ -0,0 +1,60 @@
...
+    });
+  }
+
+  set(key: string, editingData: EditingData): Promise<void> {
+    debug.editing(`Putting editing data for ${key} into redis storage...`);
+    return new Promise<void>((resolve, reject) => {
+      this.redisCache
+        .set(key, JSON.stringify(editingData))
+        .then(() => {
+          this.redisCache.expire(key, this.ttl).then(() => {
```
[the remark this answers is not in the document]
art-alexeyenko: a get call will follow a set almost always - but it's not guaranteed. Is it still better to invalidate on get instead of set?
might be better to do both to ensure there's no orphan data left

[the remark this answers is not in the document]
ambrauer: I like the idea of doing it in both (something long like 60 in the set, 0 in the get). I think about the editing cache more like a stack data structure with push/pop so in that sense the get(pop) is the place to do immediate cleanup.

packages/sitecore-jss-nextjs/src/editing/vercel-editng-data-cache.ts
```diff
@@ -0,0 +1,60 @@
...
+export class VercelEditingDataCache implements EditingDataCache {
+  protected redisCache: VercelKV;
+  private ttl;
+  
+  /**
+   * @param {string} redisUrl KV endpoint URL. Usually stored in process.env.KV_REST_API_URL
+   * @param {string} redisToken KV endpoint tokem. Usually stored in process.env.KV_REST_API_TOKEN
+   * @param {string} [entryTtlSeconds] TTL of cache entries in second. Default: 60 
+   */
+  constructor(redisUrl: string | undefined, redisToken: string | undefined, entryTtlSeconds = 60) {
```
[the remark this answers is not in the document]
art-alexeyenko: I've looked for the object type - and it's not exported from the kv library. I think it's better to have two params here, this way they can be described in comments at least

packages/sitecore-jss-nextjs/src/editing/vercel-editng-data-cache.ts
art-alexeyenko: Actually nevermind - it didn't work without transformation previously, but not using any types here works,  simplified the call

packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/editing/data/[key].ts
Comment (art-alexeyenko) [reply packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/editing/data/[key].ts:1]: For Vercel cache we should be good only using the modifications in [key].ts - this is simple to set up too.
I'll expand on configuring with BasicEditingDataService in doc input.

packages/sitecore-jss-nextjs/src/editing/vercel-editng-data-cache.ts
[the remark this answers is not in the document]
ambrauer: It's fine, though I'd still recommend an object for consistency (we can introduce our own type if needed). An object-based parameter should be equally discoverable in IDEs if doc comments are in the right places (we do this all over JSS).

packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/editing/data/[key].ts
Comment (ambrauer) [reply packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/editing/data/[key].ts:1]: True, it is simpler to configure, but at the expense of extra complexity (and performance hit) at runtime. I think in this case our recommendation should be catered towards the authoring experience. The extra bit of configuration for devs required to use `BasicEditingDataService` IMHO is worth it.

PR #1572, opened 2023-07-24
[BYOC][nextjs-sxa] Allow scaffolding components for BYOC

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
This PR modifies jss scaffold command for nextjs-sxa apps:
- Added an extra --byoc flag to create a BYOC boilerplate component
- Component now allows scaffolding components outside of src/components. Provide full path i.e. 'src/my-new-path/MyComponent' to create component in a fully custom location
- Command output will notify of further steps when component was registered outside of src/components
- To simplify registration process, instructions to bulk-register components by path are added in scripts\generate-component-builder\plugins\packages.ts

## Testing Details
- [ ] Unit Test Added
- [x] Manual Test/Other (Please elaborate) - manual runs of bootstrap, verified later with connected mode

## Types of changes


packages/create-sitecore-jss/src/templates/nextjs-sxa/scripts/generate-component-builder/plugins/packages.ts
ambrauer, line 1: Do we need this override? This file looks the same as the one in `nextjs` template.

packages/create-sitecore-jss/src/templates/nextjs/scripts/generate-component-builder/plugins/packages.ts
```diff
@@ -19,6 +19,11 @@ class PackagesPlugin implements ComponentBuilderPlugin {
      *      }
      *    ]
      *  }
+     * Or you can register all components from a path:
```
ambrauer, line 22: Consider adding this comment to React template (to keep template parity): https://github.com/Sitecore/jss/blob/dev/packages/create-sitecore-jss/src/templates/react/scripts/generate-component-builder/plugins/packages.js

ambrauer (changes requested):
I'm seeing most scaffold-component files in `nextjs-sxa` are now overriding those in the `nextjs` template. We should try to reduce this (to avoid code duplication/maintenance). I would challenge you to make changes to the base `nextjs` template in order to accomodate new BYOC requirements. Some thoughts:
1. Allowing components outside of src/components should be introduced to `nextjs` template (or we delay this feature - unless this a requirement for BYOC?)
2. As opposed to the `isByoc` prop, introduce an `args` (which we populate with any additional args from process.argv) and `template` prop to `ScaffoldComponentPluginConfig`, and use these to make the template itself "pluggable"?
3. Next steps is already "pluggable". Instead of overriding `next-steps.ts`, can't we inject a new plugin (e.g. `next-steps-byoc.ts`) to push additional steps to the `config.nextSteps`?

art-alexeyenko:
@ambrauer did some changes based on your feedback
> Allowing components outside of src/components should be introduced to nextjs template (or we delay this feature - unless this a requirement for BYOC?)
This is introduced to nextjs template (comment is present in component builder there), and now next steps in scaffold plugins account for it too. Next steps for it are put into separate plugin to make it compatible with styleguide.
 
> As opposed to the isByoc prop, introduce an args (which we populate with any additional args from process.argv) and template prop to ScaffoldComponentPluginConfig, and use these to make the template itself "pluggable"
That's a better way of doing things, implemented.

> Next steps is already "pluggable". Instead of overriding next-steps.ts, can we inject a new plugin (e.g. next-steps-byoc.ts) to push additional steps to the config.nextSteps?
What we have with possibility to create and register components outside of src/components is not byoc-specific - and so it went into base nextjs template. It makes sense to give some direction when creating BYOC, so I added another entry for BYOC steps.

I also modified next-steps for styleguide template. Since we have some set of steps in most cases, the initial text ("Scaffolding of TestNew complete. Next steps:") has been put in nextSteps array by default. We can probably add some error handling to display errors in next steps if something goes wrong.

packages/create-sitecore-jss/src/templates/nextjs-sxa/scripts/generate-component-builder/plugins/packages.ts
ambrauer: @art-alexeyenko Any thoughts on this one? It looks like the file is still present in latest.

packages/create-sitecore-jss/src/templates/nextjs/scripts/scaffold-component/plugins/next-steps-registration.ts
ambrauer, line 1: Does this need a dedicated plugin? Seems like this could just be part of `next-steps.ts` (since allowing components outside src\components is now part of base `nextjs` template).

ambrauer (changes requested):
Big improvements! Just a couple new things noticed, and one previous question still valid.

packages/create-sitecore-jss/src/templates/nextjs-sxa/scripts/generate-component-builder/plugins/packages.ts
art-alexeyenko: @ambrauer oh no, that should have been removed, an omission from latest push.

packages/create-sitecore-jss/src/templates/nextjs/scripts/scaffold-component/plugins/next-steps-registration.ts
art-alexeyenko: Yes, this should be a separate one - to make sure next-steps from nextjs-styleguide add-on does not overwrite it.

packages/create-sitecore-jss/src/templates/nextjs-sxa/scripts/scaffold-component/plugins/byoc.ts
```diff
@@ -0,0 +1,19 @@
+import { ScaffoldComponentPlugin, ScaffoldComponentPluginConfig } from '..';
+import generateByocComponentSrc from '../../templates/byoc-component-src';
```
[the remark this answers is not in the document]
art-alexeyenko: generateComponentSrc is a default export - which can be renamed during import. In here I rename the function during import to indicate it is doing byoc generation specifically.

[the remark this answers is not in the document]
ambrauer: Ah, I missed the default export. Makes sense.

packages/create-sitecore-jss/src/templates/nextjs/scripts/scaffold-component/plugins/next-steps-registration.ts
ambrauer: Ah, I missed that. Or, combine these and rename the styleguide one to `next-steps-disconnected.ts` and remove the duplicate `* Implement the React component in ${chalk.green(componentOutputPath)}` from it. Seems strange to have 2 in the base, but I'll leave it up to you :)

ambrauer (approved):
Just one last thought, otherwise this looks great 👍

PR #1580, opened 2023-08-07
[sitecore-jss-dev-tools] Fix line endings for component builder

sc-addypathania:
Fixes the following issues:
 - Line ending to have ';' instead of ','
 - should start with new line

- [x] Changelog updated

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/sitecore-jss-dev-tools/src/templating/nextjs/generate-component-builder.test.ts
```diff
@@ -121,6 +136,10 @@ describe('generate-component-builder', () => {
                 moduleName: 'Foo',
               },
             ],
+            defaultImport: {
+              useDefault: true,
+              defaultAlias: 'CustomNamespace1',
+            },
```
ambrauer, line 142: Do we have a test case with `useDefault: true` _without_ `defaultAlias`?

ambrauer (changes requested):
Looks good, just a couple questions/suggestions. Was also thinking the spec might be simplified to a single `defaultImport: boolean | string` option. If string were provided it would represent the custom namespace/alias.

Also please check the pipeline, it looks as though unit tests might be failing.

ambrauer (approved):
Looks good!

PR #1603, opened 2023-08-28
[sitecore-jss-react] [template/nextjs-sxa] Use `FEAAS.ExternalComponent` for `BYOCWrapper`

sc-addypathania:
- In order to reduce the responsibilities of BYOC wrapper and to ensure extensibility across apps, the `BYOCRenderer` which is now `BYOCCompoent` uses `<FEAAS.ExternalComponent  />` wrapper from `@sitecore-feaas/clientside 0.3.12` to render the external registered component.
     - Re-named `BYOCRenderer` to `BYOCComponent` since rendering of registered component is now handled by `FEAAS.ExternalComponent`.
     - The `FEAAS.ExternalComponent ` takes a `fallback` prop which would now handle the missing component error cases.

- [x] Changelog updated

## Testing Details

- [x] Unit Test Added

## Types of changes


art-alexeyenko:
> Re-named BYOCRenderer to BYOCComponent since rendering of registered component is now handled by FEAAS.ExternalComponent.

This is not needed -  and is actually harmful since DevEx would need to modify their rednering for this in Sitecore once again.

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  let componentProps: { [key: string]: unknown } = {};
+
+  if (props.params?.ComponentProps) {
+    try {
+      componentProps = JSON.parse(props.params.ComponentProps) || {};
+    } catch (e) {
+      console.warn(`Parsing props for ${componentName} component failed. Error: ${e}`);
+      setError(e as Error);
```
art-alexeyenko, line 108: This captures an error from parsing ComponentProps. But what happens when child component throws an error?

packages/sitecore-jss-react/src/components/BYOCComponent.test.tsx
```diff
@@ -0,0 +1,113 @@
+import React from 'react';
+import { expect } from 'chai';
+import { mount } from 'enzyme';
+import { BYOCComponent } from './BYOCComponent';
+import { MissingComponent, MissingComponentProps } from './MissingComponent';
+
+describe('BYOCComponent', () => {
+  it('should render ExternalComponent with props when ComponentProps is provided', () => {
```
art-alexeyenko, line 8: Does ExternalComponent support SSR? If so - should we see the BYOC component rendered too?

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
sc-addypathania: It should mirror the functionality of the class-based component's `getDerivedStateFromError` and `componentDidCatch` methods. If an error occurs within the component or its child components, the `setError` function should update the state with the error object.

packages/sitecore-jss-react/src/components/BYOCComponent.test.tsx
sc-addypathania: aah, i don't think it supports SSR. Maybe I should change the description of the test case.

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
art-alexeyenko: Have you tested if this would be the case in case of functional component?

packages/sitecore-jss-react/src/components/BYOCComponent.test.tsx
art-alexeyenko: Verify with Yarik if it does -  we actually want SSR here.

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
```diff
@@ -0,0 +1,125 @@
...
+  };
+
+  const unRegisteredComponentProps = {
+    rendering: {
+      componentName: componentName,
+    },
+    errorOverride: 'BYOC: This component was not registered.',
+  };
+
+  const missingProps = !componentName ? noNameProps : unRegisteredComponentProps;
```
ambrauer, line 88: If `componentName` is missing, we should short-circuit and return the missing component here right away.

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
```diff
@@ -0,0 +1,125 @@
...
+
+  const missingComponent = missingComponentComponent ? (
+    <props.missingComponentComponent {...missingProps} />
+  ) : (
+    <MissingComponent {...missingProps} />
+  );
+
+  if (error) {
```
ambrauer, line 96: Should error be checked first (before missing component logic)?

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
ambrauer: It's a good point. As far as I know error boundaries are not possible with a function component.

packages/sitecore-jss-react/src/components/BYOCComponent.test.tsx
ambrauer: Pretty sure SSR wouldn't be working just yet. We have an upcoming task to do this for BYOC (JSS-491).

ambrauer:
> > Re-named BYOCRenderer to BYOCComponent since rendering of registered component is now handled by FEAAS.ExternalComponent.
> 
> This is not needed - and is actually harmful since DevEx would need to modify their rendering for this in Sitecore once again.

@art-alexeyenko This is true but might be worth doing if we can before release (the name makes more sense and is in line with `FEAASComponent/Wrapper`).

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
sc-addypathania: So are you suggesting to do error handling within this component itself. It makes sense but should we not make use of the `fallback` prop?

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
```diff
@@ -30,21 +30,17 @@ export type BYOCProps = {
...
 /**
- * Props for BYOCRenderer component. Includes components list to load external components from.
+ * Props for BYOCComponent. Includes components list to load external components from.
  */
-export type BYOCRendererProps = BYOCProps & {
-  /**
-   * Registered component collection. Would be taken from FEAAS.External.registered
-   */
-  components: RegisteredComponents;
+export type BYOCComponentProps = BYOCProps & {
```
ambrauer, line 43: Is there a reason we need the separate `BYOCProps` type? Or can it be combined with `BYOCComponentProps`?

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
```diff
@@ -30,21 +30,17 @@ export type BYOCProps = {
...
+  params?: BYOCComponentParams;
   /**
    * fields from datasource items to be passed as rendered child component props
    */
   fields?: ComponentFields;
 };
 
 /**
- * Props for BYOCRenderer component. Includes components list to load external components from.
+ * Props for BYOCComponent. Includes components list to load external components from.
```
ambrauer, line 41: This comment about the components list is no longer accurate

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
```diff
@@ -100,36 +96,25 @@ export class BYOCRenderer extends React.Component<BYOCRendererProps> {
...
+    const unRegisteredComponentProps = {
+      rendering: {
+        componentName: componentName,
+      },
+      errorOverride: 'BYOC: This component was not registered.',
+    };
+
+    const missingProps = !componentName ? noNameProps : unRegisteredComponentProps;
+
+    const missingComponent = props.missingComponentComponent ? (
```
ambrauer, line 113: Still doesn't look like we return immediately if `componentName` is missing. It shouldn't attempt to parse ComponentProps or render <FEAAS.ExternalComponent/> in this case.

ambrauer (changes requested):

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
sc-addypathania: The [BYOCWrapper](https://github.com/Sitecore/jss/blob/1ff02cb589ce2661819aeb748bc99187ceb12fa1/packages/create-sitecore-jss/src/templates/nextjs-sxa/src/components/BYOCWrapper.tsx#L4) props are of `BYOCProps` type.

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
sc-addypathania: Actually I asked a question following your last comment. Nvm, it's clear now.

packages/sitecore-jss-react/src/components/BYOCComponent.tsx
ambrauer: Right, but why not use `BYOCComponentProps` for `BYOCWrapper` props? We do this for FEaaSWrapper:
https://github.com/Sitecore/jss/blob/1ff02cb589ce2661819aeb748bc99187ceb12fa1/packages/create-sitecore-jss/src/templates/nextjs-sxa/src/components/FEaaSWrapper.tsx#L12

PR #1609, opened 2023-09-07
[nextjs-personalize] Upgrade `Engage` to `1.4.0`

sc-addypathania:
-  Removed cdpService from @sitecore-jss/sitecore-jss 
 - Replaced cdpService methods --> `executeExperience` and `generateBrowserId` with out of the box engage functions  --> `personalize` and `handleCookie` function.
feature/jss-301-upgrade-engage 

- [ ] Changelog updated

## Testing Details

- [ ] Unit Test Added

## Types of changes


packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -153,16 +208,12 @@ export class PersonalizeMiddleware extends MiddlewareBase {
       );
       return response;
     }
-
-    const site = this.getSite(req, res);
```
ambrauer, line 157: `getSite`, `initializeEngageServer` and especially `handleCookie` should all occur _after_ skip conditions above.

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -174,29 +225,27 @@ export class PersonalizeMiddleware extends MiddlewareBase {
...
-      browserId,
       ua,
-      pointOfSale,
       params
     );
 
+    // Execute targeted experience in CDP
+    const personalizationData = {
+      channel: 'WEB',
+      currency: 'USA',
```
ambrauer, line 241: This should probably be added to our config (like `channel`) and fallback to 'USA'?

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -174,29 +225,27 @@ export class PersonalizeMiddleware extends MiddlewareBase {
...
-      ? this.config.getPointOfSale(site, language)
-      : PosResolver.resolve(site, language);
-    const variantId = await this.cdpService.executeExperience(
+    const { ua } = userAgent(req);
+
+    debug.personalize(
+      'executing experience for %s %s %s %o',
       personalizeInfo.contentId,
-      browserId,
       ua,
```
ambrauer, line 234: Can you confirm that the Engage SDK sends the User-Agent header to CDP callflows request? I see we are only using `ua` here for logging. Previously, this was sent to our `CdpService` and added as a header on CDP request.

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -174,29 +225,27 @@ export class PersonalizeMiddleware extends MiddlewareBase {
...
-    const { ua } = userAgent(req);
     const params = this.getExperienceParams(req);
-    const pointOfSale = this.config.getPointOfSale
-      ? this.config.getPointOfSale(site, language)
-      : PosResolver.resolve(site, language);
-    const variantId = await this.cdpService.executeExperience(
+    const { ua } = userAgent(req);
+
+    debug.personalize(
+      'executing experience for %s %s %s %o',
```
ambrauer, line 232: Check the debug log parameters, looks like we have 1 too many string `%s` params?

ambrauer (changes requested):

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -2,15 +2,32 @@ import { NextResponse, NextRequest, userAgent } from 'next/server';
 import {
   GraphQLPersonalizeService,
   GraphQLPersonalizeServiceConfig,
-  CdpService,
-  CdpServiceConfig,
-  ExperienceParams,
   getPersonalizedRewrite,
   PosResolver,
 } from '@sitecore-jss/sitecore-jss/personalize';
-import { debug, NativeDataFetcher } from '@sitecore-jss/sitecore-jss';
```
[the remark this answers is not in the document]
illiakovalenko: @ambrauer I would prefer to leave this implementation in our SDK. In the scope of a "minor" release, we can deprecate only (can be removed in the scope of a major release). But I think it can be useful for some customers who implemented their own middleware and decided to use our _NativeDataFetcher_, can it be a case?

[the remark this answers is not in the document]
ambrauer: Correct, we'd only be able to deprecate until JSS 22. But I agree, let's leave it for now as I can see it being useful for customers and our future selves :)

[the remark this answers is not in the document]
illiakovalenko: Maybe we will reuse it if we will implement a new middleware as an example, who knows? I assume this functionality will never be broken, since it uses a native function

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -85,33 +97,64 @@ export class PersonalizeMiddleware extends MiddlewareBase {
...
+      campaign: req.nextUrl.searchParams.get('utm_campaign') || undefined,
+      content: req.nextUrl.searchParams.get('utm_content') || undefined,
+      medium: req.nextUrl.searchParams.get('utm_medium') || undefined,
+      source: req.nextUrl.searchParams.get('utm_source') || undefined,
+    };
+
     return {
-      // It's expected that the header name "referer" is actually a misspelling of the word "referrer"
-      // req.referrer is used during fetching to determine the value of the Referer header of the request being made,
-      // used as a fallback
```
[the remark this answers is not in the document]
sc-addypathania: Aah got removed while moving things around.

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
sc-addypathania: You mean add it as a environment variable?

ambrauer: No need for env variable, just make it configurable for the middleware by adding it as a prop on `CdpServiceConfig` (like `channel`). Use that if present and fallback to 'USA' here.
```suggestion
      currency: this.config.cdpConfig.currency ?? 'USA',
```

kendoce:
@sc-addypathania @ambrauer - we might need to hold the merging of this PR since we might not include this in the 21.4.0 release due to some changes priority.

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
sc-addypathania: Confirmed, we have UA headers in the CDP callflows request.
Will remove our UA.

packages/create-sitecore-jss/src/templates/nextjs/package.json
```diff
@@ -62,7 +62,6 @@
     "eslint": "^8.32.0",
     "eslint-config-next": "^13.1.5",
     "eslint-config-prettier": "^8.6.0",
-    "eslint-plugin-prettier": "^4.2.1",
```
ambrauer, line 65: Is this supposed to be removed?

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -154,15 +170,25 @@ export class PersonalizeMiddleware extends MiddlewareBase {
...
+    const site = this.getSite(req, response);
+    const engageServer = this.initializeEngageServer(site, language);
+
+    // creates the browser ID cookie on the server side
+    // and includes the cookie in the response header
+    try {
+      await engageServer.handleCookie(req, response, timeout);
+    } catch (error) {
+      debug.personalize('skipped (browser id generation failed)');
+      console.log(error);
```
ambrauer, line 182: If we are throwing the error we don't need this log here; it will get logged already in `getHandler`. Otherwise, we can `return response` here instead of throwing like we did previously. I like the throwing.

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -174,28 +200,28 @@ export class PersonalizeMiddleware extends MiddlewareBase {
...
-      personalizeInfo.contentId,
-      browserId,
-      ua,
-      pointOfSale,
-      params
-    );
+    try {
+      variantId = (await engageServer.personalize(personalizationData, req, timeout)) as string;
+    } catch (error) {
+      console.log(error);
```
ambrauer, line 222: Same here - don't need this log if we are throwing.

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -174,28 +200,28 @@ export class PersonalizeMiddleware extends MiddlewareBase {
       return response;
     }
 
-    if (!browserId) {
-      browserId = await this.cdpService.generateBrowserId();
+    const params = this.getExperienceParams(req);
+    const { ua } = userAgent(req);
```
ambrauer, line 204: As discussed, can get rid of the UA

ambrauer (approved):
Just a few minor things noticed this time, please take a look. Overall, this looks great. Approved!

packages/sitecore-jss-nextjs/package.json
```diff
@@ -30,6 +30,7 @@
     "url": "https://github.com/sitecore/jss/issues"
   },
   "devDependencies": {
+    "@sitecore/engage": "^1.4.0",
```
ambrauer, line 33: Missed this before - don't we need to add this as a peer dependency as well?

packages/create-sitecore-jss/src/templates/nextjs/package.json
```diff
@@ -63,6 +63,7 @@
     "eslint-config-next": "^13.1.5",
     "eslint-config-prettier": "^8.6.0",
     "eslint-plugin-prettier": "^4.2.1",
+    "eslint-plugin-react": "^7.31.11",
```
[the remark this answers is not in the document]
sc-addypathania: Coming back to this comment.
I added this because the CI fails with linting errors.
It's failing now.

PR #1620, opened 2023-09-26
Chain responses in middleware.

mjarosch:
- [ ] Changelog updated

## Description / Motivation
By chaining the responses in rewrites, we can access headers set in earlier middleware.

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other - I have implemented similar changes in our Sitecore solution. I don't have a personal Sitecore instance to fully test them on at this time.

## Types of changes

### DIFF
--- a/packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.test.ts
+++ b/packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.test.ts
@@ -352,13 +352,16 @@ describe('RedirectsMiddleware', () => {
           url: 'http://localhost:3000/found',
           status: 301,
           setCookies,
+          headers: new Headers({}),
         });
-        const nextRedirectStub = sinon.stub(NextResponse, 'redirect').callsFake((url, status) => {
+        const nextRedirectStub = sinon.stub(NextResponse, 'redirect').callsFake((url, init) => {
+          const status = typeof init === 'number' ? init : init?.status || 307;
+          const headers = typeof init === 'object' ? init?.headers : {};
           return ({
             url,
             status,
             cookies: { set: setCookies },
-            headers: res.headers,
+            headers: new Headers(headers),
           } as unknown) as NextResponse;
         });
         const req = createRequest({
@@ -411,7 +414,7 @@ describe('RedirectsMiddleware', () => {
           status: 301,
           setCookies,
         });
-        const nextRewriteStub = sinon.stub(NextResponse, 'rewrite').callsFake((url) => {
+        const nextRewriteStub = sinon.stub(NextResponse, 'rewrite').callsFake((url, init) => {
           return ({
             url,
             status: 301,
@@ -526,7 +529,8 @@ describe('RedirectsMiddleware', () => {
           status: 301,
           setCookies,
         });
-        const nextRedirectStub = sinon.stub(NextResponse, 'redirect').callsFake((url, status) => {
+        const nextRedirectStub = sinon.stub(NextResponse, 'redirect').callsFake((url, init) => {
+          const status = typeof init === 'number' ? init : init?.status || 307;
           return ({
             url,
             status,
@@ -610,7 +614,6 @@ describe('RedirectsMiddleware', () => {
 
         validateEndMessageDebugLog('redirects middleware end in %dms: %o', {
           headers: {
-            'set-cookie': 'sc_site=nextjs-app; Path=/',
             'x-middleware-next': '1',
           },
           redirected: false,
@@ -687,7 +690,8 @@ describe('RedirectsMiddleware', () => {
           status: 302,
           setCookies,
         });
-        const nextRedirectStub = sinon.stub(NextResponse, 'redirect').callsFake((url, status) => {
+        const nextRedirectStub = sinon.stub(NextResponse, 'redirect').callsFake((url, init) => {
+          const status = typeof init === 'number' ? init : init?.status || 307;
           return ({
             url,
             status,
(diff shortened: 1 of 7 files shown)


mjarosch:
By making this change, middleware should favor headers for passing data. Cookies would be reserved for passing data to future requests.

mjarosch:
Sorry for all the yarn.lock commits.

art-alexeyenko:
@mjarosch It took us a while to get to this, sorry.
This is great work and an amazing code cleanup. I might need to copy the changes into a separate branch for some cleanup and conflict resolution - or if you got time, let me know if you can adjust personalize-middleware and remove yarn.lock from changes in this PR (there are no new dependencies - so yarn.lock does not need to be updated)

mjarosch:
@art-alexeyenko Yes, I will update to remove the yarn.lock file and resolve the conflicts in personalize.

> Do you mean that cookies may be absent during middleware execution of other requests or that follow-up middleware plugins have no access to cookies?

I mean follow-up middleware plugins won't have access to the cookie. The cookie is still set on the client, and future requests will see it.

art-alexeyenko:
@mjarosch 
> I mean follow-up middleware plugins won't have access to the cookie. The cookie is still set on the client, and future requests will see it.

That's not what I'm seeing when testing locally - but perhaps I miss some scenarios. SITE_SYMBOL cookie is set in multisite plugin, and if you check the value in the follow up plugin like
`    console.log(res?.cookies.get(this.SITE_SYMBOL)?.value);`
 The value will be available, on server. I also tested this in incgonito to ensure I'm not reading cookies from previous requests.

mike-jarosch-aw:
@art-alexeyenko I could have sworn, when I originally did this, that it wasn't there. But I validated on my original implementation and they are accessible. Should I remove the comment?

art-alexeyenko:
@mjarosch perhaps this PR can just be wound down to code cleanup? There's probably no reason to duplicate site in headers and cookies - but it's good to avoid code duplication.

mjarosch:
@art-alexeyenko the chaining makes it so following middleware don’t need to re-set cookies. So I would consider that more than code cleanup.

I can remove adding the site name to the headers. I think it’s nice to pass things in a consistent place and not everything should be set in a cookie. So I think you could justify duplicating the site in this case.

I defer to your judgement on this.

art-alexeyenko:
@mjarosch I'm still not sure the site should be duplicated in two places. The cookie can be used for both Sitecore instance to be consumed by, and by middleware plugins. Header will just repeat what is already set.
Could you cleanup the part around setting the header please, and then we'll merge this into the product.

mjarosch:
@art-alexeyenko I have updated it to remove storing the site in the header.

packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.test.ts
```diff
@@ -411,7 +414,8 @@ describe('RedirectsMiddleware', () => {
           status: 301,
           setCookies,
         });
-        const nextRewriteStub = sinon.stub(NextResponse, 'rewrite').callsFake((url) => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const nextRewriteStub = sinon.stub(NextResponse, 'rewrite').callsFake((url, _init) => {
```
art-alexeyenko, line 418: do we need _init here?

packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.test.ts
```diff
@@ -857,7 +863,8 @@ describe('RedirectsMiddleware', () => {
           url: 'http://localhost:3000/found',
           setCookies,
         });
-        const nextRewriteStub = sinon.stub(NextResponse, 'rewrite').callsFake((url) => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const nextRewriteStub = sinon.stub(NextResponse, 'rewrite').callsFake((url, _init) => {
```
art-alexeyenko, line 867: similarly here, should _init be used?

packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.ts
```diff
@@ -129,22 +129,26 @@ export class RedirectsMiddleware extends MiddlewareBase {
       /** return Response redirect with http code of redirect type **/
       switch (existsRedirect.redirectType) {
         case REDIRECT_TYPE_301:
-          return NextResponse.redirect(redirectUrl, 301);
+          return NextResponse.redirect(redirectUrl, {
```
art-alexeyenko, line 132: @matkovskyi could you please take a look here? Would this be in line with what you expect of SXA rewrites?
I'm talking mainly about the statusText and headers (which might be set in middleware and preserved on redirect)

art-alexeyenko (commented):
@mjarosch thank you once again!
I left a couple small comments and pinged a dev from SXA team to give a thumbs up on redirects changes, as it's their domain. We should be able to merge once there's a green light.

packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.test.ts
mjarosch: @art-alexeyenko Yes, we are calling the two parameter version of this method, so matching the call signature for the fake.

packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.test.ts
mjarosch: @art-alexeyenko Same as above, we are calling the two parameter version of this method, so matching the call signature for the fake.

packages/sitecore-jss-nextjs/src/middleware/redirects-middleware.ts
matkovskyi: I think that this will not interfere with the main logic. The code(type) of redirect is important

art-alexeyenko (approved):

PR #1621, opened 2023-09-26
[sitecore-jss-react] Handle BYOC Component error

sc-addypathania:
The BYOC component should parse the passed props when fetchedDatais not present.

- [ ] Changelog updated

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,7 +29,7 @@ Our versioning strategy is as follows:
 * `[templates/nextjs]` `[sitecore-jss-nextjs]` Better error handling for component-level data fetching ([#1586](https://github.com/Sitecore/jss/pull/1586))
 * `[sitecore-jss-react]` Fetch Data for FEaaS Components as part of Component SSR/SSG ([#1586](https://github.com/Sitecore/jss/pull/1590))
 * `[sitecore-jss-dev-tools]` `[templates/nextjs]` `[templates/react]` Introduce "components" configuration for ComponentBuilder ([#1598](https://github.com/Sitecore/jss/pull/1598))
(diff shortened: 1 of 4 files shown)


art-alexeyenko:
One question though- do we need to modify yarn.lock each time with our PRs now?
Should we be fixing yarn.lock outside of a pull request to make it more cherry-pickable? @sc-addypathania

sc-addypathania:
> One question though- do we need to modify yarn.lock each time with our PRs now? Should we be fixing yarn.lock outside of a pull request to make it more cherry-pickable? @sc-addypathania

How would you fix the yarn.lock outside of PR?
The issue is whenever we switch branches or pull changes from the remote branch if there were any change in the dependencies then the node_modules gets corrupted. Therefore the only solution i see is to reset yarn.lock before pushing the PR.

art-alexeyenko:
> The issue is whenever we switch branches or pull changes from the remote branch if there were any change in the dependencies then the node_modules gets corrupted. 

I think the issue was caused by that yarn.lock commit launched from azure pipeline automation? Now that this commit is not triggered for the previous PR merge, there's a conflict in the current one.

> How would you fix the yarn.lock outside of PR?

We should only need to update yarn lock when dependencies change. So we need to ensure that unneccesary yarn.lock updates don't trigger. Until then we can try restoring a "broken" yarn.lock via a direct [skip-ci] commit perhaps.
@sc-addypathania

sc-addypathania:
> I think the issue was caused by that yarn.lock commit launched from azure pipeline automation? Now that this commit is not triggered for the previous PR merge, there's a conflict in the current one.

Hmm probably but looks like this change has been there in the azure pipeline file since a very long time.


> We should only need to update yarn lock when dependencies change. So we need to ensure that unneccesary yarn.lock updates don't trigger. Until then we can try restoring a "broken" yarn.lock via a direct [skip-ci] commit perhaps.

Yeah we should add something to investigate this.
Sounds good. 👍🏼

PR #1637, opened 2023-10-25
 #600951: [templates/nextjs] fixed plugin of cors header

matkovskyi:
- [x] Changelog updated

## Description / Motivation

Problem was with custom headers. There was missed extends common nextjs config of headers 

## Testing Details

- [ ] Unit Test Added

## Types of changes

### DIFF
--- a/packages/create-sitecore-jss/src/templates/nextjs/src/lib/next-config/plugins/cors-header.js
+++ b/packages/create-sitecore-jss/src/templates/nextjs/src/lib/next-config/plugins/cors-header.js
@@ -10,7 +10,7 @@ const corsHeaderPlugin = (nextConfig = {}) => {
   return Object.assign({}, nextConfig, {
     async headers() {
       return [
-        ...await nextConfig.headers(),
+        ...await nextConfig?.headers(),
         {
           source: '/_next/:path*',
           headers: [{ key: 'Access-Control-Allow-Origin', value: config.sitecoreApiHost }],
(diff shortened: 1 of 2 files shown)


packages/create-sitecore-jss/src/templates/nextjs/src/lib/next-config/plugins/cors-header.js
```diff
@@ -10,6 +10,7 @@ const corsHeaderPlugin = (nextConfig = {}) => {
   return Object.assign({}, nextConfig, {
     async headers() {
       return [
+        ...await nextConfig?.headers(),
```
ambrauer, line 13: Shouldn't need `?` here right? Given the default on line 6.
```suggestion
        ...await nextConfig.headers(),
```

ambrauer (approved):
Just one suggestion, otherwise 👍

packages/create-sitecore-jss/src/templates/nextjs/src/lib/next-config/plugins/cors-header.js
matkovskyi: If there isn't headers param in next.config.js there will be error about `nextConfig.headers is not a function`

ambrauer: But this would still be a problem with this code since the optional chaining only applies to the preceding prop (`nextConfig`), right? To handle undefined `headers`, think we'd want something like this:
https://github.com/Sitecore/jss/blob/dev/packages/create-sitecore-jss/src/templates/nextjs-styleguide/src/lib/next-config/plugins/disconnected.js#L16

matkovskyi: @ambrauer I added some changes because if there isn't custom headers in next.config.js there will show error. I added conditional for fix this situation

ambrauer (approved):

PR #1638, opened 2023-10-25
[nextjs] Update feaas clientside and adjust types in byoc bundlers

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
1) Updated sitecore-feaas dependency to latests to accomodate:
- ContextID changes
- New types for bundler
2) Adjusted comments
3) Adjusted BYOC import-export logic with @Inviz 's suggestions (ground work for AppRouter BYOC support)
4) Also locked the dependency version in sitecore-jss-react package to avoid breaking changes breaking the package.

## Testing Details
- [ ] Unit Test Added

## Types of changes


packages/create-sitecore-jss/src/templates/nextjs/src/byoc/index.ts
```diff
@@ -1,17 +1,20 @@
...
 const ClientBundle = dynamic(() => import('./index.client'), {
   ssr: false,
 });
+FEAAS.enableNextClientsideComponents(dynamic, ClientBundle);
 
-// Import your SSR/hybrid components via index.hybrid.ts
+// Import your hybrid (server rendering with client hydration) components via index.hybrid.ts
 import './index.hybrid';
 
 // As long as client bundle is exported and rendered on page (as an empty element), client-only BYOC components are registered and become available
```
ambrauer, line 18: Does this comment need some updating with the change? There's a bit of magic now between the `FEAAS.enableNextClientsideComponents` above and `FEAAS.ExternalComponentBundle` below.

packages/sitecore-jss-react/package.json
```diff
@@ -27,7 +27,7 @@
     "url": "https://github.com/sitecore/jss/issues"
   },
   "devDependencies": {
-    "@sitecore-feaas/clientside": "^0.4.4",
+    "@sitecore-feaas/clientside": "0.4.12",
```
ambrauer, line 30: Why the exact version here? Should at least match our peerDependencies (~). 
In general though, are we anticipating breaking changes in @sitecore-feaas/clientside minor versions? If following semver (which we should be), then ^ should be fine.

Inviz: yeah needs ^ probably, if it was copied from my app - sorry

packages/create-sitecore-jss/src/templates/nextjs/src/byoc/index.ts
art-alexeyenko: Hybrid import is unaffected - but it makes sense to update the client import comments, or move the enableClientside call

packages/sitecore-jss-react/package.json
art-alexeyenko: > In general though, are we anticipating breaking changes in @sitecore-feaas/clientside minor versions

We don't, but sometimes they happen. I've encountered two instances during Forms integration dev. I can bring the ^ back if we don't expect any other ones.
@ambrauer

art-alexeyenko: Updated the version

ambrauer (approved):
Looks good to me

PR #1639, opened 2023-10-25
[templates/nextjs] Fix monorepo sample failing on client

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
Monorepo was failing with 'useContext' error (cannot read properties of null). This PR should fix it.

## Testing Details
- [ ] Unit Test Added

## Types of changes

### DIFF
--- a/packages/create-sitecore-jss/src/templates/nextjs/tsconfig.json
+++ b/packages/create-sitecore-jss/src/templates/nextjs/tsconfig.json
@@ -7,7 +7,7 @@
       "temp/*": ["src/temp/*"],
       "assets/*": ["src/assets/*"],
       "graphql-types": ["node_modules/@types/graphql-let/__generated__/__types__"],
-      "react": ["node_modules/@types/react"]
+      "react": ["node_modules/react"]
     },
     "target": "es5",
     "lib": ["dom", "dom.iterable", "esnext"],


jflheureux:
When starting a new project on the latest xmcloud-foundation-head (with JSS 21.6.0), the first thing my team did was to install the latest version of Storybook (7.6.10). Installation went well but no component showing in Storybook and we got browser console errors:

```
Warning: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
1. You might have mismatching versions of React and the renderer (such as React DOM)
2. You might be breaking the Rules of Hooks
3. You might have more than one copy of React in the same app
See https://reactjs.org/link/invalid-hook-call for tips about how to debug and fix this problem.

Uncaught TypeError: Cannot read properties of null (reading 'useRef')
    at useRef (react.development.js:1630:1)
    at WithCallback (react-18.mjs:4:1)
    at renderWithHooks (react-dom.development.js:10732:1)
    at mountIndeterminateComponent (react-dom.development.js:15770:1)
    at beginWork$1 (react-dom.development.js:17354:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:19467:1)
    at Object.invokeGuardedCallbackImpl (react-dom.development.js:19516:1)
    at invokeGuardedCallback (react-dom.development.js:19591:1)
    at beginWork (react-dom.development.js:25733:1)
    at performUnitOfWork (react-dom.development.js:24557:1)
```

After comparing the xmcloud-foundation-head repo with its previous version using JSS 21.5.X, we identified the change that breaks Storybook is the change from this pull request.

Reverting this change in our project resolved the Storybook issues. All we had to do was to revert the `compilerOptions.paths.react` to `["node_modules/@types/react"]` in the `tsconfig.json` file.

PR #1647, opened 2023-10-31
[RAV][node-ssr] Update samples to use SITE_NAME instead of JSS_APP_NAME

yavorsk:
- [x] Changelog updated

## Description / Motivation

Switch from using JSS_APP_NAME to SITE_NAME - this PR includes renaming all environment/config variables in React, Vue, Angular templates as well as ssr node proxy app templates.

## Testing Details

- [ ] Unit Test Added

## Types of changes


packages/create-sitecore-jss/src/templates/vue/scripts/generate-config.js
```diff
@@ -45,7 +49,16 @@ const config = {};\n`;
...
   // Set base configuration values, allowing override with environment variables
   Object.keys(config).forEach((prop) => {
-    configText += `config.${prop} = process.env.VUE_APP_${constantCase(prop)} || "${config[prop]}",\n`;
+    if (prop === 'siteName') {
+      // for the sake of backwards compatibility - make sure to initialize siteName from jssAppName if necessary
+      configText += `config.${prop} = process.env.VUE_APP_${constantCase(prop)} || "${
+        config[prop]
+      } || process.env.VUE_APP_JSS_APP_NAME || ${config.jssAppName}",\n`;
+    } else {
+      configText += `config.${prop} = process.env.VUE_APP_${constantCase('jssAppName')} || "${
```
art-alexeyenko, line 58: I think this is an error - if prop is not siteName, we assign a jssAppName value to it. But there could be other non-siteName props.
In general, perhaps it makes sense to not make a special case for siteName and instead use what you already have on lines 38-39:
`config.siteName = config.siteName || config.jssAppName`
or use a value from process.env as default for siteName on line 21, similar to how it's done in nextjs template

packages/create-sitecore-jss/src/templates/angular/scripts/generate-config.ts
```diff
@@ -50,7 +54,14 @@ const config = {};\n`;
 
   // Set base configuration values, allowing override with environment variables
   Object.keys(config).forEach((prop) => {
-    configText += `config.${prop} = process.env.${constantCase(prop)} || "${config[prop]}",\n`;
+    if (prop === 'siteName') {
```
art-alexeyenko, line 57: Similar to vue comment - do you think it can be better if we just use process.env values as default ones for siteName on line 19? This would simplify the code since the extra clause won't be needed

art-alexeyenko (changes requested):

packages/create-sitecore-jss/src/templates/vue/scripts/generate-config.js
yavorsk: Yea, I added that ugly special case because on line 38 jssAppName could be not initialized yet. Initializing both in defaultConfig using process.env values should solve this, thanks.

packages/create-sitecore-jss/src/templates/react/scripts/generate-config.js
```diff
@@ -12,14 +12,17 @@ const { jssConfigFactory } = require('./config');
 const defaultConfig = {
   sitecoreApiKey: process.env[`${constantCase('reactAppSitecoreApiKey')}`],
   sitecoreApiHost: process.env[`${constantCase('reactAppSitecoreApiHost')}`],
-  jssAppName: process.env[`${constantCase('reactAppJssAppName')}`],
+  siteName:
+    process.env[`${constantCase('reactAppSiteName')}`] ||
+    process.env[`${constantCase('reactAppJssAppName')}`],
```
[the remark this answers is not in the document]
yavorsk: Added some comments in all updated templates.

packages/create-sitecore-jss/src/templates/react/scripts/generate-config.js
```diff
@@ -9,17 +9,21 @@ const { jssConfigFactory } = require('./config');
 /**
  * REACT_APP is the default prefix for environment variables that will be made available to the app.
  */
+// REACT_APP_JSS_APP_NAME env variable has been depracated since v.21.6, REACT_APP_SITE_NAME should be used instead
```
art-alexeyenko, line 12: small typo here: should be "deprecated"

art-alexeyenko (approved):
Looks good, just fix the typos please and it's good to merge.

PR #1674, opened 2023-11-22
[nextjs/template][sitecore-jss-nextjs] On demand ISR

sc-addypathania:
Adds On-demand ISR feature from Nextjs
- An API route is introduced (/api/revalidate). When triggered this will purge the updated page path on-demand
- We have the traditional ISR(revalidate: 5s) which is enabled by default and now on-demand ISR which can be configured using a webhook.
- Users need to pass a secret and configure their environment variable i.e. _ISR_REVALIDATE_SECRET_ in order to prevent unauthorized user access.
- Users need to setup a webhook to automatically trigger the API route. They need to pass the _multiSite_ or _personalize_ Boolean values through their webhook payload if multi-site or personalize is configured.
- [x] Changelog updated

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
```diff
@@ -0,0 +1,214 @@
...
+  variantId: string;
+};
+
+export type RevalidateConfig = {
+  clientFactory: GraphQLRequestClientFactory;
+  multiSite?: boolean;
+  personalize?: boolean;
+};
+
+const SITE_PREFIX = '_site_';
```
ambrauer, line 28: We shouldn't be manually constructing these rewrite segments for site and personalize. We have util functions that handle this:
https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss/src/site/utils.ts#L13
https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss/src/personalize/utils.ts#L14

ambrauer (commented):
I didn't have time to do a full review, but here is some preliminary feedback

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
sc-addypathania: Well I looked at them and these functions work slightly different than what we need. The data I have already has the `siteName` in the path whereas the util functions accepts them separately.

packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/revalidate.ts
```diff
@@ -0,0 +1,28 @@
...
+    clientFactory,
+    multiSite: body.multiSite || false, // override this value through your webhook payload if you're using multi-site add-on
+    personalize: body.personalize || false, // override this value through your webhook payload if you're configuring personalization on any page
+  }).getHandler();
+
+  // check if the secret is provided, if not, allow the request to proceed without authentication
+  if (secret === undefined) {
+    console.log('Secret not provided. Proceeding without authentication.');
+    await revalidateHandler(req, res);
+  } else if (secret !== process.env.YOUR_REVALIDATION_SECRET) {
```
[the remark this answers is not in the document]
sc-addypathania: Makes sense to me, 👍🏼

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
illiakovalenko: I tried to figure out, but I see that you construct the rewrite path manually here: https://github.com/Sitecore/jss/pull/1674/files#diff-1a3c56f1c4214f67a0fa03b77abb29d41607f92bd5d5b2b781e4b09cdef70f90R156
In general it's not really good to duplicate this configuration here, you can try to prepare your parameters in a middleware and pass it to utils in appropriate format

packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/revalidate.ts
```diff
@@ -0,0 +1,30 @@
+﻿import { RevalidateMiddleware } from '@sitecore-jss/sitecore-jss-nextjs/revalidate';
+import { NextApiResponse, NextApiRequest } from 'next';
+import clientFactory from 'lib/graphql-client-factory';
+
+const handler = async (req: NextApiRequest, res: NextApiResponse) => {
```
illiakovalenko, line 5: I'm thinking if we need to add a description, so the purpose of this api route will be clear for the customer

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
```diff
@@ -0,0 +1,256 @@
...
+      pathsToRevalidate.push(...defaultPaths);
+    }
+
+    try {
+      // revalidate all the collected paths
+      await Promise.all(pathsToRevalidate.map((path: string) => res.revalidate(path)));
+      console.log(`[api/revalidate] revalidate: ${pathsToRevalidate.join(', ')}`);
+      return res.status(200).json({ revalidated: true });
+    } catch (error) {
+      console.error(`[api/revalidate] error: ${error}`);
```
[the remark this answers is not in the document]
illiakovalenko: Also if customer will use a different path for this api route (not a api/revalidate) this prefix will not match. So "debug" approach also should solve that

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
```diff
@@ -0,0 +1,256 @@
...
+      return res.status(500).json({ message: 'Error revalidating' });
```
illiakovalenko, line 253: Do we want to return an error message in a response?

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
```diff
@@ -0,0 +1,256 @@
...
+    };
+  }
+
+  private handler = async (req: NextApiRequest, res: NextApiResponse): Promise<void> => {
+    // filter out updated paths and language from request.body
+    const filteredUpdates = this.getFilteredUpdates(req);
+
+    if (this.isEmpty(filteredUpdates)) {
+      // nothing to revalidate
+      return res.status(204).json({});
```
illiakovalenko, line 212: Do we want to return a message?

illiakovalenko (approved):

illiakovalenko (changes requested):
Approved by mistake, need to resolve comments first :)

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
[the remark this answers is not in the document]
sc-addypathania: Makes sense.

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
sc-addypathania: Are you suggesting to return the error response or return an empty json?

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
sc-addypathania: I have refactored to use our existing util functions.
Do we have a util function to construct the path with both multiSite and personalize prefixes?

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
```diff
@@ -0,0 +1,304 @@
...
+
+  protected handleNonMultiSitePersonalization(
+    personalizeInfo: {
+      personalized: PersonalizedResult[];
+      nonPersonalized: {
+        path: string;
+      }[];
+    },
+    pathsToRevalidate: string[],
+    context: RevalidateMiddleware
```
ambrauer, line 199: Why the need to pass `this` in to these functions as "context"?

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
```diff
@@ -0,0 +1,304 @@
...
+
+    return req.body?.updates
+      .filter(
+        (update: Entity) =>
+          update.entity_definition === EntityDefinition.LayoutData && update.entity_culture
+      )
+      .map((update: Entity) => {
+        return {
+          path: update.identifier,
+          language: update.entity_culture,
```
ambrauer, line 154: This just made me think about i18n routing... Do we need to also add language to the revalidate path if this is configured (our default)? I couldn't find anything documented in Next.js docs, but I did find some discussions.
https://github.com/vercel/next.js/discussions/40489

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    const middleware = new RevalidateMiddleware({ clientFactory, personalize: true });
+    await middleware.getHandler()(req, res);
+
+    expect(getPersonalizeInfoStub.callCount).to.equal(2);
+    expect(res.status).to.be.calledWith(200);
+    expect(res.json).to.be.calledWith({ revalidated: true });
+    expect(res.revalidate).to.be.calledWith(`/_variantId_${variantId_1}/About`);
+  });
+
+  it('should revalidate when a site has some personalized and others non-personalized pages', async () => {
```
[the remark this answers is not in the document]
sc-addypathania: I have refactored the personalization tests"

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
sc-addypathania: I passed it to make it look a bit cleaner but now passed the respective params.

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.ts
sc-addypathania: Hmm to my understanding the path would still be the same. I check the webhook response and I get the following response.
image

packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/revalidate.ts
```diff
@@ -0,0 +1,38 @@
...
+    multiSite: body.multiSite || false, // override this value through your webhook payload if using multi-site add-on
+    personalize: body.personalize || false, // override this value through your webhook payload if personalization is configured
+  }).getHandler();
+
+  /**
+   NOTE: It is highly recommended to add a secret to your revalidate endpoint before going to production.
+   This prevents unauthorized users from triggering revalidation requests.
+   */
+  if (process.env.ISR_REVALIDATE_SECRET) {
+    if (secret && secret !== process.env.ISR_REVALIDATE_SECRET) {
```
ambrauer, line 27: Should also be considered a 401 if the secret is missing
```suggestion
    if (!secret || secret !== process.env.ISR_REVALIDATE_SECRET) {
```

packages/sitecore-jss-nextjs/src/revalidate/revalidate-middleware.test.ts
```diff
@@ -204,11 +191,42 @@ describe('RevalidateMiddleware', () => {
...
+    const middleware = new RevalidateMiddleware({
+      clientFactory,
+      languages: { locales: ['en', 'fr-FR'], defaultLocale: 'en' },
+    });
+    await middleware.getHandler()(req, res);
+
+    expect(res.status).to.be.calledWith(200);
+    expect(res.json).to.be.calledWith({ revalidated: true });
+    expect(res.revalidate).to.have.been.calledWithExactly('/fr-FR/');
+    expect(res.revalidate).to.have.been.calledWithExactly('/fr-FR/About');
```
ambrauer, line 207: I don't think this is right. The mock payload only has updates made to the `en` version, so we wouldn't revalidate the `fr-FR` prefixed paths. However, if the mock payload were this:
```
 {
      identifier: 'my-site/About',
      entity_definition: 'LayoutData',
      operation: 'Update',
      entity_culture: 'fr-FR',
    },
    {
      identifier: 'my-site/',
      entity_definition: 'LayoutData',
      operation: 'Update',
      entity_culture: 'fr-FR',
    },
```
then it would.

The i18n settings should only be used to determine the locale prefix of the revalidated path. Something like this flow:
- grab locale from `entity_culture`
- is locale in i18n locales?
- if no, then don't revalidate this item (skip)
- is locale the i18n defaultLocale?
- if yes, then don't add locale prefix
- if no, then add locale prefix

Of course this would only apply to Sub-path routing (our default). But what if they were using Domain routing? Perhaps we should make the locale prefix the responsibility of the app layer. Let's discuss @sc-addypathania .

packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/revalidate.ts
```diff
@@ -0,0 +1,55 @@
+﻿import { RevalidateMiddleware } from '@sitecore-jss/sitecore-jss-nextjs/revalidate';
+import { NextApiResponse, NextApiRequest } from 'next';
+import clientFactory from 'lib/graphql-client-factory';
+import nextConfig from '../../../next.config';
+
+/**
+ * Nextjs API route /api/revalidate
+ * This is a revalidate endpoint that can be used to trigger Incremental Static Regeneration (ISR).
+ * of Next.js pages on-Demand.This endpoint is called by Sitecore when content is published,
+ * and can also be called manually to trigger revalidation of pages.
```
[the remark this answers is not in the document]
sc-addypathania: Yeah right. So should we remove manually or document it?

packages/create-sitecore-jss/src/templates/nextjs/src/pages/api/revalidate.ts
```diff
@@ -0,0 +1,55 @@
...
+  const revalidateHandler = new RevalidateMiddleware({
+    clientFactory,
+    // override this value through your webhook payload if using multi-site add-on
+    multiSite: body.multiSite || false,
+    // override this value through your webhook payload if personalization is configured
+    personalize: body.personalize || false,
+    // pass the following function to handle additional locales configured in next.config
+    // if i18n isn't available or the language matches the defaultLocale, it returns an empty string,
+    // otherwise, it returns the language itself as the prefix based on the other language configured in next.config.
+    languagePrefix: (language: string) => {
```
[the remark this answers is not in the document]
sc-addypathania: Makes sense.

PR #1712, opened 2024-01-10
[sitecore-jss-nextjs] Reject SDK promise when init was rejected

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Events SDK should not initialize in dev environment or server context. This PR ensures Events SDK promise is not pending in such situations and will reject instead. Context logic has been made to handle initialization rejects.

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/create-sitecore-jss/src/templates/nextjs-xmcloud/src/lib/context/sdk/events.ts
+++ b/packages/create-sitecore-jss/src/templates/nextjs-xmcloud/src/lib/context/sdk/events.ts
@@ -6,8 +6,11 @@ const sdkModule: SDK<typeof Events> = {
   init: async (props) => {
     // Events module can't be initialized on the server side
     // We also don't want to initialize it in development mode
-    if (typeof window === 'undefined' || process.env.NODE_ENV === 'development') return;
-
+    if (typeof window === 'undefined') 
+      throw "Events SDK can't be initialized in server context";
+    if (process.env.NODE_ENV === 'development')
+      throw 'Events SDK is not initialized in development environment';
+    
     await Events.init({
       siteName: props.siteName,
       sitecoreEdgeUrl: props.sitecoreEdgeUrl,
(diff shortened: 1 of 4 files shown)


packages/sitecore-jss-nextjs/src/context/context.ts
```diff
@@ -137,12 +137,17 @@ export class Context<SDKModules extends SDKModulesType> {
...
+      this.props.sdks[name]
+        .init(this)
+        .then(() => {
+          this.sdks[name] = this.props.sdks[name].sdk;
+          resolve(this.sdks[name]);
+        })
+        .catch((e) => {
+          console.log(`Initialization for SDK ${name.toString()} skipped. Reason: \n ${e}`);
+          reject(e);
+        });
```
ambrauer, line 150: Do we need to catch and reject explicitly here or can we just let it bubble up? If we are doing it just to be able to console.log, don't know if that is necessary.

packages/create-sitecore-jss/src/templates/nextjs-xmcloud/src/lib/context/sdk/events.ts
```diff
@@ -6,8 +6,11 @@ const sdkModule: SDK<typeof Events> = {
   init: async (props) => {
     // Events module can't be initialized on the server side
     // We also don't want to initialize it in development mode
-    if (typeof window === 'undefined' || process.env.NODE_ENV === 'development') return;
-
+    if (typeof window === 'undefined') 
+      throw "Events SDK can't be initialized in server context";
```
ambrauer, line 10: Should be single quotes

packages/sitecore-jss-nextjs/src/context/context.ts
art-alexeyenko: Removed the log here - the error will now only show when attempting a getSDK on rejecting SDK

PR #1731, opened 2024-02-09
[sitecore-jss] Retry policy to handle transient network errors

sc-addypathania:
Adds better retry handling approach on top of exiting:

- User can pass _retryStrategy_ to configure custom retry config to the services. They can customize the error codes and the number of retries. It consist of two functions _shouldRetry_ and _getDelay_
- If _retryStrategy_ is not passed then _defaultRetryStrategy_ will be used. The _defaultStrategy_ handles 429 and the 500 error codes e.g. 502, 503, 504, 520, 521, 522, 523, 524.
- _getDalay_ calculates the delay (in milliseconds) before the next retry based on the given error, attempt count and exponential factor which defaults to 2.

- [x] Changelog updated

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/create-sitecore-jss/src/templates/nextjs/src/lib/layout-service-factory.ts
+++ b/packages/create-sitecore-jss/src/templates/nextjs/src/lib/layout-service-factory.ts
@@ -21,21 +21,14 @@ export class LayoutServiceFactory {
           siteName,
           clientFactory,
           /*
-            GraphQL endpoint may reach its rate limit with the amount of Layout and Dictionary requests it receives and throw a rate limit error.
+            GraphQL endpoint may reach its rate limit with the amount of requests it receives and throw a rate limit error.
             GraphQL Dictionary and Layout Services can handle rate limit errors from server and attempt a retry on requests.
-            For this, specify the number of retries the GraphQL client will attempt.
+            For this, specify the number of 'retries' the GraphQL client will attempt.
             It will only try the request once by default.
-            retries: 'number'
-            Additionally, you have the flexibility to customize the retry strategy by passing your customized RetryStrategy
-            object using the DefaultRetryStrategy class. The DefaultRetryStrategy class, which can be imported from the @sitecore-jss-nextjs
-            package, provides two essential methods: 'shouldRetry' which returns a boolean indicating whether a retry should
-            be attempted, and 'getDelay' which calculates the delay (in milliseconds) before the subsequent retry based on
-            the encountered error and the current attempt.
-            Example:
-            retryStrategy: new DefaultRetryStrategy({
-              statusCodes: 'number[]',
-              factor: 'number' (The exponential factor to calculate backoff-time),
-            }),
+
+            Additionally, you have the flexibility to customize the retry strategy by passing a 'retryStrategy'.
+            By default it uses the `DefaultRetryStrategy` with exponential back-off factor of 2 for error codes 429,
+            502, 503, 504, 520, 521, 522, 523, and 524. You can use this class or your own implementation of `RetryStrategy`.
           */
           retries:
             (process.env.GRAPH_QL_SERVICE_RETRIES &&
--- a/packages/create-sitecore-jss/src/templates/nextjs/src/lib/layout-service-factory.ts
+++ b/packages/create-sitecore-jss/src/templates/nextjs/src/lib/layout-service-factory.ts
@@ -23,10 +23,19 @@ export class LayoutServiceFactory {
           /*
             GraphQL endpoint may reach its rate limit with the amount of Layout and Dictionary requests it receives and throw a rate limit error.
             GraphQL Dictionary and Layout Services can handle rate limit errors from server and attempt a retry on requests.
-            For this, specify the number of retries the GraphQL client will attempt. 
+            For this, specify the number of retries the GraphQL client will attempt.
             It will only try the request once by default.
-            retries: 'number' 
-            Additionally, you can customize the retry strategy by providing your own RetryStrategy object.
+            retries: 'number'
+            Additionally, you have the flexibility to customize the retry strategy by passing your customized RetryStrategy
+            object using the DefaultRetryStrategy class. The DefaultRetryStrategy class, which can be imported from the @sitecore-jss-nextjs
+            package, provides two essential methods: 'shouldRetry' which returns a boolean indicating whether a retry should
+            be attempted, and 'getDelay' which calculates the delay (in milliseconds) before the subsequent retry based on
+            the encountered error and the current attempt.
+            Example:
+            retryStrategy: new DefaultRetryStrategy({
+              statusCodes: 'number[]',
+              factor: 'number' (The exponential factor to calculate backoff-time),
+            }),
           */
           retries:
             (process.env.GRAPH_QL_SERVICE_RETRIES &&
--- a/packages/create-sitecore-jss/src/templates/nextjs/src/lib/layout-service-factory.ts
+++ b/packages/create-sitecore-jss/src/templates/nextjs/src/lib/layout-service-factory.ts
@@ -26,6 +26,7 @@ export class LayoutServiceFactory {
             For this, specify the number of retries the GraphQL client will attempt. 
             It will only try the request once by default.
             retries: 'number' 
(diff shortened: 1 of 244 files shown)


packages/sitecore-jss/src/graphql-request-client.test.ts
```diff
@@ -245,4 +255,154 @@ describe('GraphQLRequestClient', () => {
...
+        await graphQLClient.request('test');
+      } catch (error) {
+        expect(error).to.not.be.undefined;
+        expect(graphQLClient['client'].request).to.have.been.called.exactly(3);
+        spy.restore(graphQLClient);
+      }
+    };
+
+    // Test cases for each retryable status code
+    for (const statusCode of [502, 503, 504, 520, 521, 522, 523, 524]) {
```
ambrauer, line 285: Should 429 be included here?

packages/create-sitecore-jss/src/templates/nextjs/src/lib/layout-service-factory.ts
```diff
@@ -26,6 +26,7 @@ export class LayoutServiceFactory {
             For this, specify the number of retries the GraphQL client will attempt. 
             It will only try the request once by default.
             retries: 'number' 
+            Additionally, you can customize the retry strategy by providing your own RetryStrategy object.
```
art-alexeyenko, line 29: How should users pass custom retryStrategy?
Graphql layout and dictionary service contsructors do not accept a retry strategy param.

packages/sitecore-jss/package.json
```diff
@@ -52,13 +52,15 @@
...
   "dependencies": {
+    "@types/sinon": "^17.0.3",
     "axios": "^0.21.1",
     "chalk": "^4.1.0",
     "debug": "^4.3.1",
     "graphql": "^16.5.0",
     "graphql-request": "^4.2.0",
     "lodash.unescape": "^4.0.1",
     "memory-cache": "^0.2.0",
+    "sinon": "^17.0.1",
```
art-alexeyenko, line 63: It's better to have this as a dev dependency - or use an already imported module alternative, if possible

packages/sitecore-jss/src/graphql-request-client.test.ts
```diff
@@ -176,44 +179,51 @@ describe('GraphQLRequestClient', () => {
...
-
-    const graphQLClient = new GraphQLRequestClient(endpoint, { retries: 2 });
-    spy.on(graphQLClient['client'], 'request');
-    await graphQLClient.request('test').catch((error) => {
-      expect(graphQLClient['client'].request).to.be.called.exactly(2);
-      expect(error.name).to.equal('AbortError');
-      spy.restore(graphQLClient);
-    });
-  });
+  // it.only('should use [retry-after] header value when response is 429', async function() {
```
art-alexeyenko, line 182: we should keep the tests for 'retry-after' header to avoid regression

packages/sitecore-jss/src/graphql-request-client.test.ts
```diff
@@ -245,4 +255,154 @@ describe('GraphQLRequestClient', () => {
...
+
+    // Test cases for each retryable status code
+    for (const statusCode of [429]) {
+      it(`should retry and resolve for ${statusCode} if one of the request resolves`, async function() {
+        this.timeout(16000);
+        await retryableStatusCodeResolve(statusCode);
+      });
+    }
+
+    it('should use custom retryStrategy and retry accordingly', async function() {
```
art-alexeyenko, line 330: please clarify the test case description based on expected input and output, i.e. should use custom delay and retry strategy when custom strategy provided

packages/sitecore-jss/src/graphql-request-client.test.ts
```diff
@@ -176,44 +179,51 @@ describe('GraphQLRequestClient', () => {
...
+  //       'Error: %d. Rate limit reached for GraphQL endpoint. Retrying in %ds. This was your %d attempt.',
+  //       429,
+  //       2,
+  //       1
+  //     );
+  //     spy.restore(graphQLClient);
+  //   });
+  // });
+
+  // it.only('should throw error when request is aborted with default timeout value after retry', async () => {
```
art-alexeyenko, line 201: This is a test when request is aborted due to long response time. We should ensure retryer handles this case, please leave this test in.

packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -65,9 +88,29 @@ export class GraphQLRequestClient implements GraphQLClient {
...
   private headers: Record<string, string> = {};
   private debug: Debugger;
-  private retries: number;
   private abortTimeout?: TimeoutPromise;
   private timeout?: number;
+  private retries: number;
+  private retryStrategy: RetryStrategy;
+  private defaultRetryStrategy: RetryStrategy = {
+    shouldRetry: (error: ClientError, attempt: number) => {
+      const retryableStatusCodes = [429, 502, 503, 504, 520, 521, 522, 523, 524];
```
art-alexeyenko, line 97: Is our final decision to include all the 5xx codes by default? This will happen for both SXP and XMC

ambrauer (changes requested):

packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -15,6 +15,25 @@ export interface GraphQLClient {
    */
   request<T>(query: string | DocumentNode, variables?: { [key: string]: unknown }): Promise<T>;
 }
+/**
+ * Defines the strategy for retrying GraphQL requests based on errors and attempts.
+ */
+export interface RetryStrategy {
```
art-alexeyenko, line 21: Please make sure RetryStrategy type is exported by sitecore-jss-nextjs, react, angular, vue packages and can be used by sample apps.

art-alexeyenko (changes requested):

packages/sitecore-jss/src/graphql-request-client.ts
sc-addypathania: Yes we discussed that yesterday but will clarify more with Adam.

packages/sitecore-jss/src/graphql-request-client.ts
sc-addypathania: Good point

packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -33,13 +53,18 @@ export type GraphQLRequestClientConfig = {
    */
   fetch?: typeof fetch;
   /**
-   * GraphQLClient request timeout
+   * GraphQLClient request timeout (in milliseconds).
    */
   timeout?: number;
   /**
-   * Number of retries for client. Will be used if endpoint responds with 429 (rate limit reached) error
+   * Number of retries for client. Number of retries for client. Will use the specified `retryStrategy`.
```
ambrauer, line 60: Dup content here
```suggestion
   * Number of retries for client. Will use the specified `retryStrategy`.
```

packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -33,13 +53,18 @@ export type GraphQLRequestClientConfig = {
...
    */
   retries?: number;
+  /**
+   * Retry strategy for the client. Default will use an exponential back-off strategy for
+   * codes 429, 502, 503, 504, 520, 521, 522, 523, 524.
```
ambrauer, line 65:
```suggestion
   * Retry strategy for the client. Uses `DefaultRetryStrategy` by default with exponential
   * back-off factor of 2 for codes 429, 502, 503, 504, 520, 521, 522, 523, 524.
```

packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -57,6 +82,29 @@ export type GraphQLRequestClientFactory = (
  */
 export type GraphQLRequestClientFactoryConfig = { endpoint: string; apiKey?: string };
 
+export class DefaultRetryStrategy implements RetryStrategy {
```
ambrauer, line 85: Add jsdoc comment to class and constructor

ambrauer (approved):
Just some doc comments to touch up / add, otherwise approved. 👍

PR #1733, opened 2024-02-12
[sitecore-jss] Optimize initial Retry delay time

sc-addypathania:
At present, the initial retry delay evaluated to 2 seconds, which becomes relatively high for subsequent retries when assessed in an exponential manner. We aim to initiate the first retry after a 1-second wait instead.

- [x] Changelog updated

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/sitecore-jss/src/graphql-request-client.test.ts
+++ b/packages/sitecore-jss/src/graphql-request-client.test.ts
@@ -407,7 +407,7 @@ describe('GraphQLRequestClient', () => {
     });
   });
 
-  describe.only('DefaultRetryStrategy class', () => {
+  describe.only('DefaultRetryStrategy', () => {
     const mockClientError = new ClientError(
       {
         data: undefined,
@@ -422,7 +422,7 @@ describe('GraphQLRequestClient', () => {
 
     const retryStrategy = new DefaultRetryStrategy([429, 503]);
 
-    it('should return false from shouldRetry method when conditions are not met', () => {
+    it('should return true from shouldRetry when conditions are met', () => {
       mockClientError.response.status = 503;
 
       const shouldRetry = retryStrategy.shouldRetry(mockClientError, 1, 3);
--- a/packages/sitecore-jss/src/graphql-request-client.test.ts
+++ b/packages/sitecore-jss/src/graphql-request-client.test.ts
@@ -413,14 +413,14 @@ describe('GraphQLRequestClient', () => {
         data: undefined,
         errors: [{ message: 'GaphqlError' }],
         extensions: undefined,
-        status: 502,
+        status: 429,
       },
       {
         query: 'query',
       }
     );
 
-    const retryStrategy = new DefaultRetryStrategy([429, 503]);
+    const retryStrategy = new DefaultRetryStrategy({ statusCodes: [429, 503] });
 
     it('should return true from shouldRetry when conditions are met', () => {
       mockClientError.response.status = 503;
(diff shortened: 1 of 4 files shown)


illiakovalenko (changes requested):
I assume that it should affect some unit tests, did you check? is it covered?

packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -113,7 +113,7 @@ export class DefaultRetryStrategy implements RetryStrategy {
     const rawHeaders = error.response?.headers;
     const delaySeconds = rawHeaders?.get('Retry-After')
       ? Number.parseInt(rawHeaders?.get('Retry-After'), 10)
-      : Math.pow(this.factor, attempt);
+      : Math.pow(this.factor, attempt - 1);
```
art-alexeyenko, line 116: do we have any tests to ensure delay is correctly calculated? similar to "should use [retry-after] header value when response is 429" case
I think that's what Illia meant by additional tests

art-alexeyenko (changes requested):

packages/sitecore-jss/src/graphql-request-client.ts
sc-addypathania: the last test i added above tests the function getDelay if it return the value correctly.

art-alexeyenko (approved):

packages/sitecore-jss/src/graphql-request-client.test.ts
```diff
@@ -405,4 +406,40 @@ describe('GraphQLRequestClient', () => {
...
+      const shouldRetry = retryStrategy.shouldRetry(mockClientError, 2, 1);
+      expect(shouldRetry).to.equal(false);
+    });
+
+    it('should return delay using exponential backoff when Retry-After header is not present', () => {
+      const delay = retryStrategy.getDelay(mockClientError, 3);
+      const expectedDelay = Math.pow(retryStrategy['factor'], 3 - 1) * 1000;
+      expect(delay).to.equal(expectedDelay);
+    });
+  });
```
ambrauer, line 444: There are a few more tests we could add here for better coverage, for example:
- shouldRetry returns false when retries is 0
- shouldRetry returns false when attempt exceeds retries
- shouldRetry returns false when error response status undefined
- getDelay uses custom exponential factor

ambrauer (approved):

packages/sitecore-jss/src/graphql-request-client.ts
```diff
@@ -92,12 +92,13 @@ export class DefaultRetryStrategy implements RetryStrategy {
...
-   * @param {number[]} statusCodes HTTP status codes to trigger retries on
-   * @param {number} factor Factor by which the delay increases with each retry attempt
+   * @param {Object} options Configurable options for retry mechanism.
+   * @param {number[]} options.statusCodes HTTP status codes to trigger retries on
+   * @param {number} options.factor Factor by which the delay increases with each retry attempt
    */
-  constructor(statusCodes?: number[], factor?: number) {
-    this.statusCodes = statusCodes || [429];
-    this.factor = factor || 2;
+  constructor(options: { statusCodes?: number[]; factor?: number }) {
```
ambrauer, line 99: Need a default empty object to make `options` optional. Probably good to add a `DefaultRetryStrategy` unit test to cover that. e.g. 'should use statusCode 429 and factor 2 by default'
```suggestion
  constructor(options: { statusCodes?: number[]; factor?: number } = {}) {
```

PR #1776, opened 2024-04-18
[sitecore-jss-react] [Editing Integration] Add support for chrome hydration for Placeholders

sc-addypathania:
- [x] Changelog updated

## Description / Motivation
To enable Chrome's hydration process in Pages, we need to modify the JSS Placeholder component in the sitecore-jss-react SDKs. We need to emit shallow chromes and required data for placeholders and renderings.
- Introduced a new `PlaceholderMetadata` component to emit shallow chromes.
- Placeholder component now uses `sitecoreContext` to access `editMode: "chromes" | "metadata"` property from Layout Service.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -193,6 +201,10 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
           : `component-${index}`;
         const commonProps = { key };
 
+        if (this.props.sitecoreContext?.editMode === EditMode.Metadata) {
+          return <PlaceholderMetadata key={key} component={rendering as ComponentRendering} />;
```
art-alexeyenko, line 205: One other thought: would it be better to move this condition into the main Placeholder component file? We have a number of condition on how exactly JSS will render placeholder components in there, like:
```
 if (this.props.renderEmpty && this.isEmpty) {
    } else if (components.length && this.isEmpty) {
    } else if (this.props.render) {
    } else if (this.props.renderEach) {
```
It should fit nicely and be better maintainable that way. Or would we lose some SSR capabilities if we return a PlaceholderMetadata component in there directly?

art-alexeyenko (changes requested):
A good start, and the implementation can be made even better, ensuring we get it as right as possible. I've left some comments for current implementation but perhaps it can go the slightly other way.

Something you can consider for the overall design: we have logic in PlaceholderCommon for the traditional way of rendering components and a bunch of conditions in main Placeholder that calls that logic. 
We can use a separate component to add "placeholder" code blocks, or it can be a function in PlaceholderCommon, or (perhaps better) an HOC that wraps placeholders (I think Yavor does something similar for fields in his PR).
HOC that outputs chrome code blocks for components (chrometype: rendering) can also be added and used to wrap components in componentBuilder - ensuring every registered component has code blocks added.

packages/sitecore-jss-react/src/components/PlaceholderMetadata.tsx
```diff
@@ -0,0 +1,85 @@
...
+ * This component handles recursive rendering of components and their placeholders, encapsulating
+ * each part within specific code tags to outline their structure and metadata attributes.
+ *
+ * @param {PlaceholderMetadataProps} props - The props containing the component data to render.
+ * @returns {JSX.Element} - The rendered component with all nested components and placeholders.
+ */
+export const PlaceholderMetadata: React.FC<PlaceholderMetadataProps> = ({ component }) => {
+  const defaultAttributes = {
+    type: 'text/sitecore',
+    chrometype: 'field',
```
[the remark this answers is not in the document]
sc-addypathania: Good catch. Totally missed it.

packages/sitecore-jss-react/src/components/Placeholder.test.tsx
```diff
@@ -101,7 +101,9 @@ describe('<Placeholder />', () => {
         const phKey = 'page-content';
 
         const renderedComponent = mount(
-          <Placeholder name={phKey} rendering={component} componentFactory={componentFactory} />
+          <SitecoreContext componentFactory={componentFactory}>
```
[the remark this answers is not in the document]
sc-addypathania: I think context is always there for PlaceholderComponent. Now we need to wrap it with SitecoreContext in order for it to find the desired element.

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
sc-addypathania: Whatever gets evaluated in `PlaceholderCommon` is passed to `Placeholder` and it would anyway go through all those conditions. If we move this to `Placeholder` then we would have to create another function to process the input data which we already get ready here in `PlacehollderCommon`.

packages/sitecore-jss-react/src/components/PlaceholderMetadata.tsx
```diff
@@ -0,0 +1,85 @@
...
+      kind: `${kind}`,
+    };
+  };
+
+  const renderComponent = (component: ComponentRendering): JSX.Element => {
+    const ComponentName = component.componentName as React.ElementType;
+    return (
+      <>
+        <code {...codeAttributes('rendering', 'open')}>{`{uid: "${component.uid}"}`}</code>
+        <ComponentName />
```
illiakovalenko, line 34: Here you need to render a component, currently it's a string

packages/sitecore-jss-react/src/components/PlaceholderMetadata.tsx
```diff
@@ -0,0 +1,85 @@
...
+        <code {...codeAttributes('placeholder', 'close')} />
+      </React.Fragment>,
+    ]);
+  };
+
+  // check if the component has nested placeholders
+  return (
+    <>
+      {component.placeholders && Object.keys(component.placeholders).length > 0
+        ? renderComponentsInPlaceholder(
```
illiakovalenko, line 78: After rewriting unit tests, check and be sure that the contract is followed
especially the case when placeholders rendered inside of nested renderings.
It seems like that here some structure should be changed, from my pov

packages/sitecore-jss-react/src/components/PlaceholderMetadata.tsx
```diff
@@ -0,0 +1,85 @@
...
+    return (
+      <>
+        <code {...codeAttributes('rendering', 'open')}>{`{uid: "${component.uid}"}`}</code>
+        <ComponentName />
+        <code {...codeAttributes('rendering', 'close')}></code>
+      </>
+    );
+  };
+
+  const renderNestedComponents = (components: ComponentRendering[]): JSX.Element[] => {
```
illiakovalenko, line 40: This function is kind of similar to _renderComponent_, I mean by name
should we encapsulate it into the _renderComponentsInPlaceholder_?

packages/sitecore-jss-react/src/components/PlaceholderMetadata.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const codeAttributes = (chromeType: string, kind: string) => {
+    return {
+      type: 'text/sitecore',
+      chrometype: `${chromeType}`,
+      className: 'scpm',
+      kind: `${kind}`,
+    };
+  };
+
+  const renderComponent = (component: ComponentRendering): JSX.Element => {
```
illiakovalenko, line 29:
```suggestion
  const renderRendering = (rendering: ComponentRendering): JSX.Element => {
```

packages/sitecore-jss-react/src/components/PlaceholderMetadata.tsx
```diff
@@ -0,0 +1,85 @@
+﻿import React from 'react';
+import { ComponentRendering } from '@sitecore-jss/sitecore-jss/layout';
+
+/**
+ *  Props containing the component data to render.
+ */
+export interface PlaceholderMetadataProps {
+  component: ComponentRendering;
```
illiakovalenko, line 8:
```suggestion
  rendering: ComponentRendering;
```

packages/sitecore-jss-react/src/components/PlaceholderMetadata.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/**
+ * This component supports the chrome's hydration process in Pages by rendering the required metadata.
+ * This component handles recursive rendering of components and their placeholders, encapsulating
+ * each part within specific code tags to outline their structure and metadata attributes.
+ *
+ * @param {PlaceholderMetadataProps} props - The props containing the component data to render.
```
illiakovalenko, line 16:
```suggestion
 * @param {PlaceholderMetadataProps} props - The props containing the rendering data to render.
```

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -102,6 +109,7 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
       PropTypes.func as Requireable<React.FC<unknown>>,
     ]),
     modifyComponentProps: PropTypes.func,
+    sitecoreContext: PropTypes.object as Requireable<SitecoreContextValue>,
```
illiakovalenko, line 112: is there a way to use _PropTypes.object.isRequired_? this is the expected property

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
illiakovalenko: @art-alexeyenko It should be done in PlaceholderCommon, since we execute all the editing related logic here

packages/sitecore-jss-react/src/components/Placeholder.test.tsx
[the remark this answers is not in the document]
art-alexeyenko: The layout data context was not there in unit tests before - and it's still not needed for most test scenarios

packages/sitecore-jss-react/src/components/PlaceholderMetadata.tsx
sc-addypathania: I have done some refactoring, they don't look the same now. Its very hard to combine them into one function.

packages/sitecore-jss-react/src/components/Placeholder.test.tsx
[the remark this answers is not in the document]
sc-addypathania: Since now we wrap the placeholder component with `SitecoreContext` it is necessary to wrap each test with `SitecoreContext` otherwise it cannot get componentFactory. A common function is also not possible or maybe not very valuable since Placeholder Component accepts so many props and each test case tests different props so a common function would be a overkill.

packages/sitecore-jss-react/src/components/PlaceholderMetadata.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    const expectedOutput =
+      '<code type="text/sitecore" chrometype="placeholder" class="scpm" kind="open" id="main_123"></code><code type="text/sitecore" chrometype="placeholder" class="scpm" kind="close" id="main_123"></code>';
+
+    const actualOutput = normalizeWhitespace(wrapper.html());
+    const normalizedExpectedOutput = normalizeWhitespace(expectedOutput);
+
+    expect(actualOutput).to.equal(normalizedExpectedOutput);
+  });
+});
```
art-alexeyenko, line 109: Please also add a test case for placeholder that renders a component with missing implementation (not registered in componentBuilder).
This component should still be wrapped in placeholder and rendering code blocks.

packages/sitecore-jss-react/src/components/PlaceholderMetadata.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+  it('renders a single component without placeholders', () => {
+    const component = {
+      uid: '123',
+      componentName: 'RichText',
+    };
+
+    const wrapper = shallow(<PlaceholderMetadata rendering={component} />);
+
+    const expectedOutput =
+      '<code type="text/sitecore" chrometype="rendering" class="scpm" kind="open" id="123"></code><RichText></RichText><code type="text/sitecore" chrometype="rendering" class="scpm" kind="close" id="123"></code>';
```
art-alexeyenko, line 24: Would it be possible to make this more readable? Kind of like this:
https://github.com/Sitecore/jss/pull/1773/files#diff-77b3f451fd7b04ab45e786f4f22abe108550edb30a868b1949f8a814ac9d18d5R106-R114 ?

art-alexeyenko: Also, is it expected here for RichText component to not be rendered as HTML? Should this be full SSR instead?

sc-addypathania: It was just being rendered as a string. Now it is html.

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -277,6 +300,8 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
       );
     }
 
+    // console.log('real', componentFactory(renderingDefinition.componentName));
```
yavorsk, line 303: This line could be removed?

packages/sitecore-jss-react/src/components/Placeholder.test.tsx
```diff
@@ -677,18 +709,46 @@ it('should render custom HiddenRendering when rendering is hidden', () => {
...
+        hiddenRenderingComponent={CustomHiddenRendering}
+      />
+    </SitecoreContext>
   );
   expect(renderedComponent.find('.hidden-rendering').length).to.equal(1);
   expect(renderedComponent.find(HiddenRendering).length).to.equal(1);
   expect(renderedComponent.find('p').props().children).to.equal('Hidden Rendering');
 });
 
+const testDataWithoutEE = [
```
illiakovalenko, line 725:
```suggestion
const testDataWithMetadata = [
```

packages/sitecore-jss-react/src/test-data/non-ee-data.ts
```diff
@@ -510,3 +512,50 @@ export const feaasWrapperData = {
     },
   },
 };
+
+export const layoutDataWithMetadata = {
+  sitecore: {
+    context: {
+      pageEditing: false,
```
illiakovalenko, line 519:
```suggestion
      pageEditing: true,
```

packages/sitecore-jss-react/src/components/Placeholder.test.tsx
illiakovalenko: And something is definitely wrong here. You are relying on non-editing layout service data, but your unit test should cover editing scenario and inside of unit test you are referencing _layoutDataWithMetadata_ that is confusing

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -193,6 +203,19 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
           : `component-${index}`;
         const commonProps = { key };
 
+        // if editMode is equal to 'metadata' then emit shallow chromes for hydration in Pages
+        if (this.props.sitecoreContext?.editMode === EditMode.Metadata) {
+          return (
+            <PlaceholderMetadata
```
illiakovalenko, line 209: From my understanding the PlaceholderMetadata
should accept a minimum amount of props and we shouldn't duplicate most part of functionality from _PlaceholderCommon_
My expectation that you can render it here: https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx#L254
If your rendering has placeholders, it will render another _Placeholder_ component, so all this functionality will be executed again. But in your case you are trying to traverse all the tree at once (seems like)
Here is an example of rendering that contains a nested placeholder: https://github.com/Sitecore/jss/blob/dev/packages/create-sitecore-jss/src/templates/nextjs-styleguide/src/components/graphql/GraphQL-Layout.tsx#L31

```ts
const rendered = React.createElement<{ [attr: string]: unknown }>(
  component as React.ComponentType,
  this.props.modifyComponentProps ? 
  this.props.modifyComponentProps(finalProps) : finalProps
);

// if editMode is equal to 'metadata' then emit shallow chromes for hydration in Pages
if (this.props.sitecoreContext?.editMode === EditMode.Metadata) {
  return <PlaceholderMetadata {...someRequiredProps}>{rendered}</PlaceholderMetadata>;
}

return rendered;
```

packages/sitecore-jss-react/src/components/PlaceholderMetadata.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * @param {React.Component} component - The React component to set up for testing.
+ * @returns {object} An object containing the component, placeholderProps, and mock functions.
+ */
+function setupTest(component) {
+  const componentFactory: ComponentFactory = (componentName: string) => {
+    const components = new Map<string, React.FC>();
+
+    components.set('RichText', () => <div className="richtext-mock" />);
+    components.set('Header', () => <div className="header-mock" />);
```
illiakovalenko, line 25: According to test data that I can see, _Header_ component contains _logo_ placeholder. So here you need to render also a placeholder, like it's done in a sample. You need to revisit your test results

yavorsk (approved):

illiakovalenko (changes requested):

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
sc-addypathania: I updated it but then it breaks the Placeholder component. I think we need to ensure that the PropTypes in PlaceholderCommon component perfectly mirror the TypeScript interfaces.

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
sc-addypathania: All other comments depend on the decision here.
Suggested approach is great but we wont be able to handle nested placeholders.
Yes I am traversing the whole tree at once in order to handle nested placeholders, please let me know if you can think of something better than this?

illiakovalenko: @sc-addypathania Please, update your unit test as it suggested here: https://github.com/Sitecore/jss/pull/1776#discussion_r1579082115
You will see, that traversing the whole tree is a problem, since you can have some nested _Placeholders_, after you render renderings using _componentFactory_

packages/sitecore-jss-react/src/components/Placeholder.test.tsx
```diff
@@ -677,18 +706,313 @@ it('should render custom HiddenRendering when rendering is hidden', () => {
...
+        hiddenRenderingComponent={CustomHiddenRendering}
+      />
+    </SitecoreContext>
   );
   expect(renderedComponent.find('.hidden-rendering').length).to.equal(1);
   expect(renderedComponent.find(HiddenRendering).length).to.equal(1);
   expect(renderedComponent.find('p').props().children).to.equal('Hidden Rendering');
 });
 
+describe.only('PlaceholderMetadata', () => {
```
illiakovalenko, line 722:
```suggestion
describe('PlaceholderMetadata', () => {
```

packages/sitecore-jss-react/src/test-data/non-ee-data.ts
```diff
@@ -1,3 +1,5 @@
+import { EditMode } from '@sitecore-jss/sitecore-jss/layout';
```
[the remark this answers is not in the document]
illiakovalenko: I suppose you need to use this value for some _metadata_ related data shapes

packages/sitecore-jss-nextjs/src/components/Placeholder.tsx
```diff
@@ -11,7 +11,7 @@ export const Placeholder = (props: PlaceholderComponentProps) => {
   return (
     <ReactPlaceholder
       {...props}
-      modifyComponentProps={(initialProps) => {
+      modifyComponentProps={(initialProps: { rendering: { uid: string | number } }) => {
```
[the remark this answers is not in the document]
sc-addypathania: I left this in order for us to make a decision regarding the _SitecoreContext_ optional or required. Accordingly porotypes would change.

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -251,12 +261,42 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
...
           component as React.ComponentType,
           this.props.modifyComponentProps ? this.props.modifyComponentProps(finalProps) : finalProps
         );
+
+        // if editMode is equal to 'metadata' then emit shallow chromes for hydration in Pages
+        if (this.props.sitecoreContext?.editMode === EditMode.Metadata) {
+          return (
+            <PlaceholderMetadata
+              key={key}
+              uid={(rendering as ComponentRendering).uid}
```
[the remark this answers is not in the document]
illiakovalenko: Nvm, it's applicable here

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -251,12 +261,42 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
...
+        }
+
+        return rendered;
       })
       .filter((element) => element); // remove nulls
+
+    if (this.props.sitecoreContext?.editMode === EditMode.Metadata) {
+      return [
+        <PlaceholderMetadata
+          key={(this.props.rendering as ComponentRendering).uid}
```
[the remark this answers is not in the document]
illiakovalenko: Nvm, it's applicable here

packages/sitecore-jss-nextjs/src/components/Placeholder.tsx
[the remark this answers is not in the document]
sc-addypathania: I think the introduction of `sitecoreContext` to the `PlaceholderComponentProps` might have influenced TypeScript's type inference or exposed previously overlooked implicit 'any' usage. I don't see anything else causing this as `sitecoreContext` is optional prop.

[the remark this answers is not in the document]
art-alexeyenko: Previously, withComponentFactory HOC would return a placeholder with the same props type. withSitecoreContext returns Placeholder with "WithSitecoreContextHocProps" prop type.
I think this is the culprit. typescript considers this as "any" type because it cannot find the "modifyComponentProps" field in WithSitecoreContextHocProps.

packages/sitecore-jss-react/src/components/PlaceholderMetadata.tsx
```diff
@@ -0,0 +1,72 @@
...
+  const renderComponent = (uid: string, placeholderName?: string) => (
+    <>
+      <code
+        {...getCodeBlockAttributes('open', uid, placeholderName)}
+        key={`open-${placeholderName}-${uid}`}
+      />
+      {children}
+      <code
+        {...getCodeBlockAttributes('close', uid, placeholderName)}
+        key={`close-${placeholderName}-${uid}`}
```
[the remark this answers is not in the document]
sc-addypathania: So this key was more of a react thing than the feature because it was an array before. I have removed them now since its not a array anymore. Also id is only for opening code bocks and not for closing.

art-alexeyenko (approved):
Good work!

Let's figure out the purpose of WithSitecoreContextHocProps type and remove it if needed, and then merge

PR #1786, opened 2024-04-26
[sitecore-jss-react] Introduce ErrorBoundary to serve as error handling per component

yavorsk:
- [x] Changelog updated

## Description / Motivation

This PR introduces ErrorBoundary component. All rendered components are wrapped with it and will catch client or server side (uses < Suspense />) errors from any of its children, display appropriate message and prevent the rest of the application from failing. It accepts and can display custom error component and loading message if it is passed as a prop to parent Placeholder. 
Note that ErrorBoundary will not catch errors for:
- Event handlers (for that, you need to use regular try/catch)
- Asynchronous code (e.g., setTimeout or requestAnimationFrame callbacks)
- Server-side rendering
- Errors thrown in the error boundary itself (rather than its children)

If an error occurs in development mode or during page editing the ErrorBoundary will display error message with the name of the faulty component and the error's message.
If we are in production mode and the error happens not in editing mode ErrorBoundary will display generic error message without providing additional information about the error and the component.
If Custom error component has been provided to the Placeholder showing it will take precedence over the above scenarios.

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate) - test with both client and server side errors (example throw errors in GraphQL-ConnectedDemo.dynamic and GraphQL-IntegratedDemo styleguide components) both for live and edit modes; error message component should be shown instead of only the failing component - rest of the app should be fine;

## Types of changes


packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,74 @@
...
+          return (
+            <div>
+              <div className="sc-jss-placeholder-error">
+                A rendering error occurred in component <em>{this.state.error.message}</em>
+              </div>
+            </div>
+          );
+        }
+      }
+    }
```
ambrauer, line 62: Should probably use the `customErrorComponent` in edit mode as well ... but do we need a difference between edit/normal mode? We don't make that distinction in our error handling at the placeholder level (code [here](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss-react/src/components/Placeholder.tsx#L74-L84), docs [here](https://doc.sitecore.com/xmc/en/developers/jss/217/jss-xmc/working-with-placeholders-in-a-jss-next-js-app.html#error-components)).

Also, is there any reason to keep the error handling at the placeholder level? Seems like this effectively supersedes that.

yavorsk: Hey @ambrauer,
we had a discussion regarding this on Friday.
I have reworked it with the following logic:
- if custom error component has been provided - show it.
- if we are in development mode or in editing mode - show a message with additional information - error's message and the name of the faulty component (could be useful for troubleshooting).
- if error is in production and not editing mode - show a generic message for the final user, without details.
What's your opinion regarding the above?

And yes, I guess we could remove the error handling on placeholder level.

packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,80 @@
...
+  type: string;
+};
+
+class ErrorBoundary extends React.Component<ErrorBoundaryProps> {
+  defaultErrorMessage =
+    'There was a problem loading this section. Refresh your browser to try again.'; // eslint-disable-line
+  defaultLoadingMessage = 'Loading component...';
+  state: { error: Error };
+
+  constructor(props: any) {
```
sc-addypathania, line 27: Shouldn't we be using `ErrorBoundaryProps` here?
`any` should be avoided.

packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,80 @@
...
+    super(props);
+    this.state = { error: null };
+  }
+
+  static getDerivedStateFromError(error: Error) {
+    return { error: error };
+  }
+
+  componentDidCatch(error: Error, errorInfo: any) {
```
sc-addypathania, line 36: Same here, we should avoid type any.

packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export type ErrorBoundaryProps = {
+  children: ReactNode;
+  customErrorComponent?:
+    | React.ComponentClass<CustomErrorComponentProps>
+    | React.FC<CustomErrorComponentProps>;
+  rendering?: ComponentRendering;
+  sitecoreContext: SitecoreContextValue;
+  componentLoadingMessage?: string;
+  type: string;
```
sc-addypathania, line 18: Move required props at the top and optional at the bottom.

packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
ambrauer: @yavorsk Makes sense, just one last thought about that last one...
 
 > if error is in production and not editing mode - show a generic message for the final user, without details.
 
 I know this is how Placeholder-level error handling is working, but I'm wondering if we should render anything at all in this case? If an error occurs on your production website, wouldn't it be preferrable to hide the erroneous component and limit the error details to the console/logging?

yavorsk: That also makes sense. I'll ask for Kusuma's opinion regarding this.

rangkusu: @aweber1, @yavorsk 
If the error affects the user experience, we should show error on whats happening otherwise they won't know right?
For example, when I click on something it meant to take me to another component but due to error if we don't show anything user won't know what happened.
Am I missing something?

ambrauer: @rangkusu True as well. I think it really depends on the component and how it's used on a page.
@yavorsk I say let's leave it as-is (consistent with Placeholder-level error handling). If a customer _did_ want to hide it when in an error state on production, they could easily accomplish this with a custom `errorComponent`.

sc-addypathania (approved):
Great work! 👍🏼

ambrauer (approved):

PR #1831, opened 2024-06-24
[sitecore-jss-react] [sitecore-jss-nextjs] Default Placeholder Content for empty fields in editMode metadata 

yavorsk:
- [x] Changelog updated

## Description / Motivation

This PR enhances the rendering of fields in editing mode for editMode metadata. When a field value is empty, a placeholder content should be rendered to signify that the field has no value. The placeholder should be interactable, so user can add value to the field. Additionally developers can provide their own custom empty field placeholder component via the field's `emptyValueEditingPlaceholder` prop, which will be rendered instead of the default one.
The fields with this functionality are:
- [sitecore-jss-react] Date.tsx
- [sitecore-jss-react] Image.tsx
- [sitecore-jss-react] Link.tsx
- [sitecore-jss-react] RichText.tsx
- [sitecore-jss-react] Text.tsx
- [sitecore-jss-nextjs] Link.tsx
- [sitecore-jss-nextjs] NextImage.tsx
- [sitecore-jss-nextjs] RichText.tsx

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate): Test in XmCloud in editMode=metadata

## Types of changes

### DIFF
--- a/packages/sitecore-jss-nextjs/src/components/Link.tsx
+++ b/packages/sitecore-jss-nextjs/src/components/Link.tsx
@@ -30,7 +30,10 @@ export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
 
     if (
       !field ||
-      (!(field as LinkFieldValue).editable && !field.value && !(field as LinkFieldValue).href)
+      (!(field as LinkFieldValue).editable &&
+        !field.value &&
+        !(field as LinkFieldValue).href &&
+        !field.metadata)
     ) {
       return null;
     }
--- a/packages/sitecore-jss-nextjs/src/components/Link.tsx
+++ b/packages/sitecore-jss-nextjs/src/components/Link.tsx
@@ -41,7 +41,7 @@ export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
     const value = ((field as LinkFieldValue).href
       ? field
       : (field as LinkField).value) as LinkFieldValue;
-    const { href, querystring, anchor } = value;
+    const { href, querystring, anchor } = value || {};
 
     const isEditing =
       editable && ((field as LinkFieldValue).editable || (field as LinkFieldValue).metadata);
--- a/packages/sitecore-jss-nextjs/src/components/Link.tsx
+++ b/packages/sitecore-jss-nextjs/src/components/Link.tsx
@@ -41,6 +41,7 @@ export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
     const value = ((field as LinkFieldValue).href
       ? field
       : (field as LinkField).value) as LinkFieldValue;
+    // fallback to {} if value is undefined; could happen if field is LinkFieldValue, href is empty in metadata mode
     const { href, querystring, anchor } = value || {};
 
     const isEditing =
--- a/packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
+++ b/packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
@@ -4,7 +4,10 @@ import chaiString from 'chai-string';
 import { mount } from 'enzyme';
 import React from 'react';
 import { NextImage } from './NextImage';
-import { ImageField } from '@sitecore-jss/sitecore-jss-react';
+import {
+  ImageField,
+  DefaultEmptyFieldEditingComponentImage,
+} from '@sitecore-jss/sitecore-jss-react';
 import { ImageLoader } from 'next/image';
 import { spy, match } from 'sinon';
 import sinonChai from 'sinon-chai';
@@ -287,7 +290,7 @@ describe('<NextImage />', () => {
     });
   });
 
-  it('should render field metadata component when metadata property is present', () => {
+  describe('editMode metadata', () => {
     const testMetadata = {
       contextItem: {
         id: '{09A07660-6834-476C-B93B-584248D3003B}',
@@ -300,21 +303,78 @@ describe('<NextImage />', () => {
       rawValue: 'Test1',
     };
 
-    const field = {
-      value: { src: '/assets/img/test0.png', alt: 'my image' },
-      metadata: testMetadata,
-    };
+    it('should render field metadata component when metadata property is present', () => {
+      const field = {
+        value: { src: '/assets/img/test0.png', alt: 'my image' },
+        metadata: testMetadata,
+      };
 
-    const rendered = mount(<NextImage field={field} fill={true} />);
+      const rendered = mount(<NextImage field={field} fill={true} />);
 
-    expect(rendered.html()).to.equal(
-      [
-        `<code type="text/sitecore" chrometype="field" class="scpm" kind="open">${JSON.stringify(
-          testMetadata
-        )}</code>`,
-        '<img alt="my image" loading="lazy" decoding="async" data-nimg="fill" style="position: absolute; height: 100%; width: 100%; left: 0px; top: 0px; right: 0px; bottom: 0px; color: transparent;" sizes="100vw" srcset="/_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=640&amp;q=75 640w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=750&amp;q=75 750w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=828&amp;q=75 828w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=1080&amp;q=75 1080w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=1200&amp;q=75 1200w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=1920&amp;q=75 1920w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=2048&amp;q=75 2048w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=3840&amp;q=75 3840w" src="/_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=3840&amp;q=75">',
-        '<code type="text/sitecore" chrometype="field" class="scpm" kind="close"></code>',
-      ].join('')
-    );
+      expect(rendered.html()).to.equal(
+        [
+          `<code type="text/sitecore" chrometype="field" class="scpm" kind="open">${JSON.stringify(
+            testMetadata
+          )}</code>`,
+          '<img alt="my image" loading="lazy" decoding="async" data-nimg="fill" style="position: absolute; height: 100%; width: 100%; left: 0px; top: 0px; right: 0px; bottom: 0px; color: transparent;" sizes="100vw" srcset="/_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=640&amp;q=75 640w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=750&amp;q=75 750w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=828&amp;q=75 828w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=1080&amp;q=75 1080w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=1200&amp;q=75 1200w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=1920&amp;q=75 1920w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=2048&amp;q=75 2048w, /_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=3840&amp;q=75 3840w" src="/_next/image?url=%2Fassets%2Fimg%2Ftest0.png&amp;w=3840&amp;q=75">',
+          '<code type="text/sitecore" chrometype="field" class="scpm" kind="close"></code>',
+        ].join('')
+      );
+    });
+
+    it('should render default empty field placeholder for Image when field value is empty in edit mode metadata', () => {
+      const field = {
+        value: '',
+        metadata: testMetadata,
+      };
+
+      const rendered = mount(<NextImage field={field} />);
+      const defaultEmptyImagePlaceholder = mount(<DefaultEmptyFieldEditingComponentImage />);
+      expect(rendered.html()).to.equal(
+        [
+          `<code type="text/sitecore" chrometype="field" class="scpm" kind="open">${JSON.stringify(
+            testMetadata
+          )}</code>`,
+          defaultEmptyImagePlaceholder.html(),
+          '<code type="text/sitecore" chrometype="field" class="scpm" kind="close"></code>',
+        ].join('')
+      );
+    });
+
+    it('should render custom empty field placeholder when provided, when field value is empty in edit mode metadata', () => {
+      const field = {
+        value: '',
+        metadata: testMetadata,
+      };
+
+      const EmptyValueEditingPlaceholder: React.FC = () => (
+        <span className="empty-field-value-placeholder">Custom Empty field value</span>
+      );
+
+      const rendered = mount(
+        <NextImage field={field} emptyValueEditingPlaceholder={EmptyValueEditingPlaceholder} />
+      );
+
+      expect(rendered.html()).to.equal(
+        [
+          `<code type="text/sitecore" chrometype="field" class="scpm" kind="open">${JSON.stringify(
+            testMetadata
+          )}</code>`,
+          '<span class="empty-field-value-placeholder">Custom Empty field value</span>',
+          '<code type="text/sitecore" chrometype="field" class="scpm" kind="close"></code>',
+        ].join('')
+      );
+    });
+
+    it('should render nothing when field value is empty, when editing is explicitly disabled in edit mode metadata ', () => {
+      const field = {
+        value: '',
+        metadata: testMetadata,
+      };
+
+      const rendered = mount(<NextImage field={field} editable={false} />);
+
+      expect(rendered.html()).to.equal('');
+    });
   });
 });
--- a/packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
+++ b/packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
@@ -322,9 +322,9 @@ describe('<NextImage />', () => {
       );
     });
 
-    it('should render default empty field placeholder for Image when field value is empty in edit mode metadata', () => {
+    it('should render default empty field placeholder for Image when field value src is not present', () => {
       const field = {
-        value: '',
+        value: {},
         metadata: testMetadata,
       };
 
@@ -341,9 +341,28 @@ describe('<NextImage />', () => {
       );
     });
 
-    it('should render custom empty field placeholder when provided, when field value is empty in edit mode metadata', () => {
+    it('should render default empty field placeholder for Image when field src is not present', () => {
       const field = {
-        value: '',
+        src: undefined,
+        metadata: testMetadata,
+      };
+
+      const rendered = mount(<NextImage field={field} />);
+      const defaultEmptyImagePlaceholder = mount(<DefaultEmptyFieldEditingComponentImage />);
+      expect(rendered.html()).to.equal(
+        [
+          `<code type="text/sitecore" chrometype="field" class="scpm" kind="open">${JSON.stringify(
+            testMetadata
+          )}</code>`,
+          defaultEmptyImagePlaceholder.html(),
+          '<code type="text/sitecore" chrometype="field" class="scpm" kind="close"></code>',
+        ].join('')
+      );
+    });
+
+    it('should render custom empty field placeholder when provided, when field value src is not present', () => {
+      const field = {
+        value: {},
         metadata: testMetadata,
       };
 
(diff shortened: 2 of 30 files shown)


packages/sitecore-jss-react/src/enhancers/withEmptyValueEditingPlaceholder.tsx
```diff
@@ -0,0 +1,65 @@
...
+ * @param {ComponentType<FieldComponentProps>} FieldComponent the field component
+ * @param {React.FC} defaultEmptyFieldEditingComponent the default empty field placeholder component
+ * @param {boolean} isForwardRef set to 'true' if forward reference is needed
+ */
+export function withEmptyValueEditingPlaceholder<
+  FieldComponentProps extends WithEmptyValueEditingPlaceholderProps,
+  RefElementType = HTMLElement
+>(
+  FieldComponent: ComponentType<FieldComponentProps>,
+  defaultEmptyFieldEditingComponent: React.FC,
```
art-alexeyenko, line 27: if empty component is being passed into HOC anyway, would it be better to combine this and the emptyValueEditingPlaceholder props into one? Just have the field components decide which implementation to use and pass into HOC?

packages/sitecore-jss-react/src/components/Date.tsx
```diff
@@ -21,39 +23,48 @@ export interface DateFieldProps {
...
+  emptyValueEditingPlaceholder?: React.ComponentClass | React.FC;
 }
 
-export const DateField: React.FC<DateFieldProps> = withFieldMetadata(
-  ({ field, tag, editable = true, render, ...otherProps }) => {
-    if (!field || (!field.editable && !field.value)) {
-      return null;
-    }
+export const DateField: React.FC<DateFieldProps> = withFieldMetadata<DateFieldProps>(
+  withEmptyValueEditingPlaceholder<DateFieldProps>(
```
art-alexeyenko, line 35: instead of creating another HOC, can the empty stub logic be put into withFieldMetadata?

packages/sitecore-jss-react/src/index.ts
```diff
@@ -103,4 +103,9 @@ export { withDatasourceCheck } from './enhancers/withDatasourceCheck';
 export { EditFrameProps, EditFrame } from './components/EditFrame';
 export { ComponentBuilder, ComponentBuilderConfig } from './ComponentBuilder';
 export { withFieldMetadata } from './enhancers/withFieldMetadata';
+export { withEmptyValueEditingPlaceholder } from './enhancers/withEmptyValueEditingPlaceholder';
```
art-alexeyenko, line 106: if the HOC logic will not be merged into withFieldMetadata, please re-export it from jss-nextjs package too for JSS developers to use

packages/sitecore-jss-react/src/index.ts
```diff
@@ -103,4 +103,9 @@ export { withDatasourceCheck } from './enhancers/withDatasourceCheck';
...
 export { EditingScripts } from './components/EditingScripts';
+export {
+  DefaultEmptyFieldEditingComponentText,
+  DefaultEmptyFieldEditingComponentImage,
```
art-alexeyenko, line 110: same with the default empty field stubs

art-alexeyenko (changes requested):
Solid work, thank you!
Let's refine it a bit

packages/sitecore-jss-react/src/enhancers/withEmptyValueEditingPlaceholder.tsx
```diff
@@ -0,0 +1,65 @@
...
+  isForwardRef = false
+) {
+  const hasValue = (field: GeneralField | ImageFieldValue | LinkFieldValue) =>
+    field?.value || (field as ImageFieldValue)?.src || (field as LinkFieldValue)?.href;
```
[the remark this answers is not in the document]
yavorsk: hm interesting suggestion. It might be. Although wouldn't it be a possible case where we have metadata.rawValue but the href property is empty?

packages/sitecore-jss-nextjs/src/components/NextImage.tsx
```diff
@@ -9,77 +9,81 @@ import {
...
+  withEmptyValueEditingPlaceholder<NextImageProps>(
+    ({ editable = true, imageParams, field, mediaUrlPrefix, fill, priority, ...otherProps }) => {
+      // next handles src and we use a custom loader,
+      // throw error if these are present
+      if (otherProps.src) {
+        throw new Error('Detected src prop. If you wish to use src, use next/image directly.');
+      }
 
-    const dynamicMedia = field as ImageField | ImageFieldValue;
+      const dynamicMedia = field as ImageField | ImageFieldValue;
```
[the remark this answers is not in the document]
yavorsk: I also noticed, but spaces are added by the formatter because the intendation changed after adding the additional HOC

packages/sitecore-jss-react/src/components/Date.tsx
```diff
@@ -21,39 +23,48 @@ export interface DateFieldProps {
    */
   editable?: boolean;
   render?: (date: Date | null) => React.ReactNode;
+  /**
+   * -- Edit Mode Metadata --
+   *
+   * Custom element to render in Pages in Metadata edit mode if field value is empty
+   */
+  emptyValueEditingPlaceholder?: React.ComponentClass | React.FC;
```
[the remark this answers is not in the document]
yavorsk: i like "emptyEditingFieldComponent"

packages/sitecore-jss-react/src/components/Date.tsx
yavorsk: hm both HOCs work for metadata mode, but are doing two very different things...  I feel it is better to leave those separate?

art-alexeyenko: I look at it this way: both HOCs are required for editing and both need to be active in metadata mode. Having them separate could lead more error prone setuo and increased maintenance complexity. And I'd rather avoid stacking HOCs if we need to add yet another metadata functinality for fields - JSS developers would need to remember to apply all of them too for their custom fields, and it would be a bad experience.
Rendering empty values for fields in editing is still part of managing fields in metadata - in that sense it fits into withFieldMetadata HOC

packages/sitecore-jss-react/src/enhancers/withEmptyValueEditingPlaceholder.tsx
ambrauer: I agree. I think it's cleaner to have HOC take a single required prop (remove any knowledge of default) and the consumers should handle sending override or default (e.g. `emptyValueEditingPlaceholder || DefaultEmptyFieldEditingComponentText`).

packages/sitecore-jss-react/src/enhancers/withEmptyValueEditingPlaceholder.tsx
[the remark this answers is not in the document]
ambrauer: Might also consider making this a core (shared) field util. We have some field utilities already here: https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss/src/layout/utils.ts

packages/sitecore-jss-react/src/components/Date.tsx
ambrauer: I'd lean towards separate here ... In addition to Yavor's reasoning, these are composite in nature (one wraps the other in terms of React/DOM tree) so nesting them makes sense.

art-alexeyenko: @ambrauer should we consider introducing a composite HOC that is basically "all you need for wrapping fields in editing". I do think we need something like that to simplify custom implementations

ambrauer: @art-alexeyenko I think having them separate actually gives more flexibility (e.g. for custom field types). The field metadata is quite generic but I can imagine you may want something custom for the empty value placeholder (e.g. "isEmpty" logic is different).

packages/sitecore-jss-react/src/enhancers/withEmptyValueEditingPlaceholder.tsx
yavorsk: that makes sense. But seems we can't pass the default value via props - see what we need to do for editable here: https://github.com/Sitecore/jss/blob/4a047aa201182c8a0b678f949f8fb9f72c6a7783/packages/sitecore-jss-react/src/enhancers/withFieldMetadata.tsx#L40

packages/sitecore-jss-react/src/enhancers/withEmptyValueEditingPlaceholder.tsx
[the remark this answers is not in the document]
pzi: What if the value is through a `JssGraphQLField`? Then it'd `jss.value.href`, no?

packages/sitecore-jss-react/src/components/Link.test.tsx
```diff
@@ -140,21 +139,77 @@ describe('<Link />', () => {
...
+          )}</code>`,
+          '<a href="/lorem">ipsum</a>',
+          '<code type="text/sitecore" chrometype="field" class="scpm" kind="close"></code>',
+        ].join('')
+      );
+    });
+
+    it('should render default empty field placeholder when field value is empty in edit mode metadata', () => {
+      const field = {
+        value: '',
```
illiakovalenko, line 163: Need to add a test case to cover: when field is the value itself (GenericFIeldValue) type (applicable for rest of unit tests below)

illiakovalenko, line 163: Empty Link value is _undefined_ or { href: undefined }

packages/sitecore-jss/src/layout/utils.test.ts
```diff
@@ -1,7 +1,8 @@
 /* eslint-disable no-unused-expressions */
 import { expect } from 'chai';
 import { ComponentRendering } from '../../layout';
-import { getFieldValue, getChildPlaceholder } from './utils';
+import { getFieldValue, getChildPlaceholder, isFieldValueEmpty } from './utils';
+import { describe } from 'node:test';
```
illiakovalenko, line 5: Not sure if it works as expected, but we are using _mocha_ global _describe_. Not sure how this one will appear in test coverage report, let's remove it

packages/sitecore-jss/src/layout/utils.test.ts
```diff
@@ -53,4 +54,116 @@ describe('sitecore-jss layout utils', () => {
...
   });
+
+  describe('isFieldValueEmpty', () => {
+    it('should return true if passed parameter is not present', () => {
+      const field = {};
+      const result = isFieldValueEmpty(field);
+      expect(result).to.be.true;
+    });
+
+    it('should return true if src is empty for GenericFieldValue', () => {
```
illiakovalenko, line 65: Let's name these test cases a bit more explicitly. This one is related to _ImageField_.
May be you can structure _describe_ blocks by field type?
e.g.
```js
describe('Image' ...)
describe('Link' ...)
...
```

packages/sitecore-jss/src/layout/utils.test.ts
```diff
@@ -53,4 +54,116 @@ describe('sitecore-jss layout utils', () => {
       expect((result[0] as ComponentRendering).componentName).to.be.equal('placed');
     });
   });
+
+  describe('isFieldValueEmpty', () => {
```
illiakovalenko, line 58: Add test cases to cover:
* boolean value
* if _field_ is undefined

packages/sitecore-jss-react/src/enhancers/withEmptyFieldEditingComponent.tsx
```diff
@@ -0,0 +1,74 @@
...
+ * defaultEmptyFieldEditingComponent: the default empty field placeholder component
+ * isForwardRef: set to 'true' if forward reference is needed
+ * */
+export interface WithEmptyFieldEditingComponentOptions {
+  defaultEmptyFieldEditingComponent: React.FC;
+  isForwardRef?: boolean;
+}
+
+interface WithEmptyFieldEditingComponentProps {
+  // Parial is used here because field.value could be required or optional for the different field types
```
illiakovalenko, line 20:
```suggestion
  // Partial<T> type is used here because _field.value_ could be required or optional for the different field types
```

packages/sitecore-jss-react/src/enhancers/withEmptyFieldEditingComponent.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+        <div>
+          <h1>{props.field?.value}</h1>
+          <h2 ref={ref}>foo</h2>
+          <p>bar</p>
+        </div>
+      );
+    }
+  );
+
+  it('Should render provided default empty value placeholder component if field value is empty in metadata edit mode', () => {
```
illiakovalenko, line 54:
```suggestion
  it('Should render provided default empty component if field value is empty in metadata edit mode', () => {
```

packages/sitecore-jss-react/src/enhancers/withEmptyFieldEditingComponent.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+      defaultEmptyFieldEditingComponent: DefaultEmptyFieldEditingComponentText,
+    });
+
+    const rendered = mount(<WrappedComponent {...props} />);
+    const expected = mount(<DefaultEmptyFieldEditingComponentText />);
+
+    expect(rendered.html()).to.equal(expected.html());
+  });
+
+  it('Should render custom empty value placeholder if provided via props if field value is empty in metadata edit mode', () => {
```
illiakovalenko, line 72:
```suggestion
  it('Should render custom empty component if provided via props if field value is empty in metadata edit mode', () => {
```

packages/sitecore-jss-react/src/enhancers/withEmptyFieldEditingComponent.test.tsx
illiakovalenko, line 54: Let's add a _describe_ block called _Metadata mode_, so you can avoid adding this in a title of each test case

packages/sitecore-jss-react/src/enhancers/withEmptyFieldEditingComponent.test.tsx
```diff
@@ -0,0 +1,196 @@
...
+      }
+    );
+    const ref = React.createRef<HTMLDivElement>();
+    const rendered = mount(<WrappedComponent {...props} ref={ref} />);
+
+    expect(ref.current?.outerHTML).to.equal('<h2>foo</h2>');
+    expect(rendered.html()).to.equal('<div><h1>field value</h1><h2>foo</h2><p>bar</p></div>');
+  });
+
+  it('Should render component if field src value is provided (case for Image component)', () => {
```
illiakovalenko, line 165: You need to add unit tests to cover the cases when _field_ it's a value itself. For Image, Link components. So I would recommend to add separate _describe_ blocks for these components

packages/sitecore-jss-nextjs/src/components/Link.tsx
```diff
@@ -30,7 +30,10 @@ export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
 
     if (
       !field ||
-      (!(field as LinkFieldValue).editable && !field.value && !(field as LinkFieldValue).href)
+      (!(field as LinkFieldValue).editable &&
+        !field.value &&
+        !(field as LinkFieldValue).href &&
+        !field.metadata)
```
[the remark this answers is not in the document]
yavorsk: where is the similar check for nextjs RichText? (link does not show it). I think we don't need it for RichText - we don't have similar 'return null' as it is in here in Link.

packages/sitecore-jss/src/layout/utils.ts
```diff
@@ -72,3 +78,35 @@ export function getChildPlaceholder(
...
+
+/**
+ * Determines if the passed in field object's value is empty.
+ * @param {GenericFieldValue | Partial<Field>} field the field object; Parial is used here because field.value could be required or optional for the different field types
+ */
+export function isFieldValueEmpty(field: GenericFieldValue | Partial<Field>): boolean {
+  const isEmpty = (fieldValue: GenericFieldValue) => {
+    if (typeof fieldValue === 'object') {
+      if (
+        (fieldValue as { [key: string]: unknown }).src ||
```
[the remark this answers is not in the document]
yavorsk: not sure if suggested change above is the same - example, it will evaluate to true if src property is present but is empty string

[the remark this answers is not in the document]
illiakovalenko: @yavorsk True, for empty string you need to have an exact check

packages/sitecore-jss-nextjs/src/components/Link.tsx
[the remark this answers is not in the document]
illiakovalenko: @yavorsk See [here](https://github.com/Sitecore/jss/blob/dev/packages/sitecore-jss-nextjs/src/components/RichText.tsx#L35)

packages/sitecore-jss-nextjs/src/components/Link.tsx
```diff
@@ -30,15 +30,18 @@ export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
...
     ) {
       return null;
     }
 
     const value = ((field as LinkFieldValue).href
       ? field
       : (field as LinkField).value) as LinkFieldValue;
-    const { href, querystring, anchor } = value;
+    const { href, querystring, anchor } = value || {};
```
illiakovalenko, line 44: Was something changed with the _value_ type? Previously it wasn't required fallback to object

packages/sitecore-jss-react/src/enhancers/withEmptyFieldEditingComponent.tsx
```diff
@@ -0,0 +1,81 @@
...
+  isFieldValueEmpty,
+  FieldMetadata,
+} from '@sitecore-jss/sitecore-jss/layout';
+
+/**
+ * The HOC options
+ * */
+export interface WithEmptyFieldEditingComponentOptions {
+  /**
+   * the default empty field placeholder component
```
illiakovalenko, line 14: Let's avoid using _placeholder_ naming
```suggestion
   * the default empty field component
```

packages/sitecore-jss-nextjs/src/components/NextImage.test.tsx
illiakovalenko, line 1: Replace "placeholder" by "component" in each unit test title

packages/sitecore-jss-react/src/components/Image.test.tsx
illiakovalenko, line 1: Replace "placeholder" by "component" in each unit test title

packages/sitecore-jss-react/src/components/Link.test.tsx
illiakovalenko, line 1: Replace "placeholder" by "component" in each unit test title

illiakovalenko (changes requested):
Looks good!
See some other comments below
Also, revisit comments:
1- https://github.com/Sitecore/jss/pull/1831#discussion_r1667963075
2- https://github.com/Sitecore/jss/pull/1831#discussion_r1660727182

packages/sitecore-jss-nextjs/src/components/Link.tsx
yavorsk: some tests were failing - in case field is LinkFieldValue, href is empty and we don't have 'value' we want to fallback to react link and render empty component; however in that case the destructuring fails on that line

packages/sitecore-jss-nextjs/src/components/Link.tsx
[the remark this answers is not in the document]
yavorsk: I see. i added it

packages/sitecore-jss-nextjs/src/components/Link.tsx
illiakovalenko: @yavorsk It means that when field.value is empty it's not handled by HOC, isn't it?
this code shouldn't be executed if field.value is empty

yavorsk: so the think in nextjs link is that we are not wrapping it with the hoc, bc react link would be double wrapped in that case. which withEmptyFieldEditincComponent is not that big of a problem i guess - just in some cases the same code would be executed twice for the same field. Should we wrap nextjs link also?

packages/sitecore-jss-react/src/enhancers/withEmptyValueEditingPlaceholder.tsx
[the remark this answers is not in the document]
yavorsk: I have created the isFieldValueEmpty function in the base package; hey @pzi, I don't understand what you mean. and what do you mean with `JssGraphQLField` ?

packages/sitecore-jss-nextjs/src/components/Link.tsx
illiakovalenko: @yavorsk I see, now it makes sense. Can you add a comment there may be? To explain why we fallback the value
Should it be the same problem for RichText component as well? Since we reference _value_ but it's not wrapped by HOC

yavorsk: nextjs RichText is different in that, that all logic on what to return is made in the react rich text component, and we don't have Link's return null [here](https://github.com/Sitecore/jss/blob/b669ab5a913abcd7041b2240bb1c08ac50143c6d/packages/sitecore-jss-nextjs/src/components/Link.tsx#L38)

illiakovalenko: @yavorsk As we discussed, for RichText it's not a problem, since _value_ is a string

illiakovalenko (approved):
👍

packages/sitecore-jss-react/src/enhancers/withEmptyValueEditingPlaceholder.tsx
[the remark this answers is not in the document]
yavorsk: hey @pzi we can have both _file.href_ and _field.value.href_ and we cover both - if that's what you are asking... 
see https://github.com/Sitecore/jss/blob/f18852aa2b89327d0e79b35222094aa4ee96fff5/packages/sitecore-jss/src/layout/utils.ts#L113

PR #1853, opened 2024-07-24
[sitecore-jss][nextjs] Ensure hidden rendering is displayed when variant is hidden in metadata editing

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Modifies the layout personalizer to identify hidden rendering in layoutData, ensuring it's correctly rendered in metadata editing mode.
Side (non-breaking) effect of this PR is moving HIDDEN_RENDERING_NAME constant into base package.

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other  - connected mode with Postman

## Types of changes


packages/sitecore-jss/src/personalize/layout-personalizer.ts
```diff
@@ -86,7 +101,14 @@ export function personalizeComponent(
     return null;
```
ambrauer, line 101: Need to apply the same logic here as well (hidden by default case) and add accompanying unit test. 
I just tested this and it suffers from the same issue.

ambrauer (changes requested):
Looking good, just need to cover another use case (hidden by default)

Issue #1834, opened 2024-07-01
Link component is not checking if href is undefined


deansimcox:
### Describe the Bug

If the field passed to the Link component has an undefined href, it will render undefined as a string in the dom.

### To Reproduce

```
import {Link} from '@sitecore-jss/sitecore-jss-nextjs';

const field = {value: {href: undefined, text: 'This is a link with a href that is undefined'}};

<Link field={field} />
```

This will render a link like this:

```
<a href="undefined">This is a link with a href that is undefined</a>
```

### Expected Behavior

I would expect the undefined href to be filtered out, leaving behind an anchor that with no href attribute.

E.g.

```
<a>This is a link with a href that is undefined</a>
```

### Possible Fix

On line 102, we should be checking for undefined
https://github.com/Sitecore/jss/blob/v20.3.3/packages/sitecore-jss-react/src/components/Link.tsx#L102

E.g.
```
href: link.href ? `${link.href}${querystring}${anchor}` : undefined,
```

### Provide environment information

* Sitecore Version: XP 10.3 (at least I think from memory)
* JSS Version: "@sitecore-jss/sitecore-jss-nextjs": "^20.3.3",
* Browser Name and version: Arc 1.49.0
* Operating System and version (desktop or mobile): macOS 14.5
* Link to your project (if available):

sc-addypathania:
@deansimcox, thanks for reporting this bug. I will add it to our backlog and prioritize it accordingly. However, if you feel this is a blocker for you, feel free to contribute by issuing a PR to our repository.

deansimcox:
All good, appreciate the update @sc-addypathania

pzi:
> @deansimcox, thanks for reporting this bug. I will add it to our backlog and prioritize it accordingly. However, if you feel this is a blocker for you, feel free to contribute by issuing a PR to our repository.

If we PR a fix, would you consider patching it into jss@20.3.3?

sc-addypathania:
@pzi yes you can issue a fix on the dev branch for now and we will try to prioritize the patch releases soon.

sumon1991:
@sc-addypathania I created a PR of it please review it.

yavorsk:
hey @sumon1991 thanks for your contribution :) I've created an item in our backlog to review and merge your PR

art-alexeyenko:
This should be have been fixed by https://github.com/Sitecore/jss/commit/b4309e8c7f8464c642e9dbe478b909f73df89ab9 

[...]

PR #1855, opened 2024-07-25
updated the link href

sumon1991:


sumon1991:
#1834 updated the link href

sumon1991:
#1834 updated the link href

PR #1858, opened 2024-07-25
[create-sitecore-jss] Allow proxies to be initialized with main app and rework angular initializer

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
* Introduces new `nodeAppDestination` arg for `create-sitecore-jss`. It will be used as destination for node proxy apps, when present
* Allows proxy apps to be installed alongside main apps (devex win!)
* Adds some initial meat to `angular-xmcloud` and `node-xmcloud-proxy` templates, simplifying smoke testing

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/create-sitecore-jss/src/bin.ts
```diff
@@ -26,6 +36,60 @@ export const parseArgs = (): ParsedArgs => {
...
+  }
+  // validate/gather destination
+  const defaultBaseDestination = `${process.cwd()}${
+    args.appName ? sep + args.appName : `${sep}${templates[0]}`
+  }`;
+  const destination = args.destination
+    ? args.destination
+    : args.yes
+    ? defaultBaseDestination
+    : await promptDestination('Where would you like your new app created?', defaultBaseDestination);
```
sc-addypathania, line 51: Should we use regular if/else for the ease of reading this?

packages/create-sitecore-jss/src/bin.ts
```diff
@@ -26,6 +36,60 @@ export const parseArgs = (): ParsedArgs => {
...
+  if (proxyApp) {
+    const defaultProxyDestination = proxyApp && `${process.cwd()}${sep}${proxyApp}`;
+    let nodeAppDestination = args.nodeAppDestination
+      ? args.nodeAppDestination
+      : args.yes
+      ? defaultProxyDestination
+      : await promptDestination(
+          'Where would you like your proxy app created?',
+          defaultProxyDestination
+        );
```
sc-addypathania, line 64: same comment as above

packages/create-sitecore-jss/src/bin.ts
art-alexeyenko: I can switch to if/else. The way it's done now was to not have too large of a duplicating code block (same logic is used further down)

packages/create-sitecore-jss/src/bin.ts
```diff
@@ -6,13 +6,23 @@ import { initRunner } from './init-runner';
...
   // parse any command line arguments passed into `init sitecore-jss`
   // to pass to the generator prompts and skip them.
   // useful for CI and testing purposes
   const options = {
     boolean: ['appPrefix', 'force', 'noInstall', 'yes', 'silent', 'prePushHook'],
-    string: ['appName', 'destination', 'templates', 'hostName', 'fetchWith', 'language'],
+    string: [
+      'appName',
+      'destination',
+      'nodeAppDestination',
```
ambrauer, line 20: I think we should change this to `proxyDestination` for the CLI param for clarity and consistency with prompts (and technically even the main app is a node app).

packages/create-sitecore-jss/src/common/args/base.ts
```diff
@@ -33,6 +33,10 @@ export type BaseArgs = {
    * Pre-push hook for linting check
    */
   prePushHook?: boolean;
+  /**
+   * Optional destination for proxy app, whether it's initialized alongside the main one or standalone
+   */
+  nodeAppDestination?: string;
```
ambrauer, line 39: Same comment here (rename to `proxyDestination`)

packages/create-sitecore-jss/src/init-runner.ts
```diff
@@ -41,16 +41,19 @@ export const initRunner = async (initializers: string[], args: BaseArgs) => {
...
 
-  saveConfiguration(args.templates, path.resolve(`${args.destination}${sep}package.json`));
-  // final steps (install, lint, etc)
-  if (!args.noInstall) {
-    installPackages(args.destination, args.silent);
-    lintFix(args.destination, args.silent);
-  }
+  for (const destination of [args.destination, args.nodeAppDestination]) {
+    if (!destination) continue;
+    saveConfiguration(args.templates, path.resolve(`${destination}${sep}package.json`));
```
ambrauer, line 46: Not sure if saving the applied "templates" makes sense for the proxy app. Should we save this saveConfiguration operation only for the primary app?

packages/create-sitecore-jss/src/init-runner.ts
```diff
@@ -41,16 +41,19 @@ export const initRunner = async (initializers: string[], args: BaseArgs) => {
...
+    // final steps (install, lint, etc)
+    if (!args.noInstall) {
+      installPackages(destination, args.silent);
+      lintFix(destination, args.silent);
```
ambrauer, line 50: Our proxy apps don't have linting configured. Skip this for those?

packages/create-sitecore-jss/src/init-runner.ts
```diff
@@ -41,16 +41,19 @@ export const initRunner = async (initializers: string[], args: BaseArgs) => {
...
+    }
 
-  // install pre-push hook if user opts-in
-  if (args.prePushHook) {
-    await installPrePushHook(args.destination, args.silent);
+    // install pre-push hook if user opts-in
+    if (args.prePushHook) {
```
ambrauer, line 54: Same here (proxy apps don't include pre-push hooks). Skip this for those?

ambrauer (changes requested):

packages/create-sitecore-jss/src/bin.ts
art-alexeyenko: `proxyDestination` reads like a request destination. Let me use `proxyAppDestination` then

packages/create-sitecore-jss/src/init-runner.ts
art-alexeyenko: Do we intend it to not have lint, ever?

art-alexeyenko: I think it's better to call this and other util functions regardless and let the create logic check whether the commands are present or not in package.json. This is simpler, more intuitive and allows us worry less if we decide to add lint and other things to proxies later on.
Prior to this PR the functions would be launched anyway when create-sitecore-jss is called for proxy only.

ambrauer: Ah, you're right. I agree, makes sense to leave it then.

packages/create-sitecore-jss/src/init-runner.ts
art-alexeyenko: Following on my comment above, I added a check to see if install-hook command exists.
I believe the scaffold logic should be flexible and more universal with its flags and logic behind them

ambrauer (approved):
👍 Looks good!

PR #1865, opened 2024-07-30
[angular-xmcloud][create-sitecore-jss] Add sxa styles to angular-xmcloud add on

yavorsk:
- [x] Changelog updated

## Description / Motivation

- add sxa styles to angular-xmcloud addon and necessary updates to angular.json;
- update transform process so that json files from addon are merged to json from base template (similarly to package.json) during initialization;

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/create-sitecore-jss/src/common/utils/helpers.ts
```diff
@@ -32,11 +32,15 @@ export const getPascalCaseName = (name: string): string => {
  */
 export const openPackageJson = (pkgPath?: string) => {
   const filePath = path.resolve(pkgPath ?? `.${sep}package.json`);
+  return openJson(filePath);
+};
+
+export const openJson = (jsonPath: string) => {
```
ambrauer, line 38: Instead of introducing another helper, can we just make the current `openPackageJson` and `writePackageJson` generic (suggested names: `openJsonFile` and `writeJsonFile`)?
I checked usage of these and in all cases the `pkgPath` is being passed. So, we could make the path parameter on the generic functions required like you've got.

packages/create-sitecore-jss/src/common/processes/transform.ts
```diff
@@ -290,6 +291,12 @@ export const transform = async (
         // merge them and set the result to str which will then go through diff
         const merged = merge(currentPkg, templatePkg);
         str = JSON.stringify(merged, null, 2);
+      } else if (file.endsWith('.json') && fs.existsSync(pathToNewFile)) {
```
ambrauer, line 294: See comment in helpers. Seems like we could make this generic and just have single case which handles all .json files.

packages/create-sitecore-jss/src/templates/angular-xmcloud/angular.json
```diff
@@ -0,0 +1,34 @@
...
+              "node_modules/bootstrap/dist/css/bootstrap.min.css",
+              "src/styles.css",
+              "src/assets/styles/main.scss"
+            ],
+            "stylePreprocessorOptions": {
+              "includePaths": ["src/assets/styles", "node_modules"]
+            }
+          }
+        },
+        "test": {
```
ambrauer, line 17: Do we need "stylePreprocessorOptions" defined for "test" as well?

packages/create-sitecore-jss/src/templates/angular-xmcloud/angular.json
```diff
@@ -0,0 +1,34 @@
...
+    }
+  },
+  "schematics": {
+    "@schematics/angular:component": {
+      "prefix": "app",
+      "style": "scss"
+    },
+    "@schematics/angular:directive": {
+      "prefix": "app"
+    }
```
ambrauer, line 32: This looks the same as the base. Is it needed here?

ambrauer (changes requested):
Looks good overall! See some comments and also looks like there are now some merge conflicts.

packages/create-sitecore-jss/src/templates/angular-xmcloud/angular.json
yavorsk: yeap, i can remove it.

packages/create-sitecore-jss/src/templates/angular-xmcloud/angular.json
yavorsk: i think so, adding it.

packages/create-sitecore-jss/src/common/utils/helpers.ts
yavorsk: yea that makes sense, I will update that. Initially i was thinking to do that for _openPackageJson_ only.

packages/create-sitecore-jss/src/common/utils/helpers.ts
```diff
@@ -26,32 +26,30 @@ export const getPascalCaseName = (name: string): string => {
 };
 
 /**
- * Provides `package.json` data
- * @param {string} [pkgPath] path to `package.json`. Default is './package.json'.
- * @returns `package.json` data
+ * Provides json data from a file
+ * @param {string} [jsonFilePath] path to the .json file.
```
ambrauer, line 30: Remove brackets here (no longer optional)
```suggestion
 * @param {string} jsonFilePath path to the .json file.
```

packages/create-sitecore-jss/src/common/utils/helpers.ts
```diff
@@ -26,32 +26,30 @@ export const getPascalCaseName = (name: string): string => {
...
   }
 };
 
 /**
- * Creates `package.json` file and inserts provided data
- * @param {Object} data data to be written into package.json
- * @param {string} [pkgPath] a path to a file. Default is './package.json'.
+ * Creates a .json file and inserts provided data
+ * @param {Object} data data to be written into the .json file
+ * @param {string} [jsonFilePath] a path to a file.
```
ambrauer, line 46: Same here
```suggestion
 * @param {string} jsonFilePath a path to a file.
```

ambrauer (approved):
Just minor jsdoc change, otherwise 👍

PR #1880, opened 2024-08-08
[template/angular-xmcloud] Fix angular layout component

sc-addypathania:
- [x] Changelog updated

## Description / Motivation
- Fix build issues
- Fix console errors

## Testing Details

- [ ] Unit Test Added

## Types of changes

### DIFF
--- a/packages/create-sitecore-jss/src/templates/angular-xmcloud/src/app/jss-link.service.ts
+++ b/packages/create-sitecore-jss/src/templates/angular-xmcloud/src/app/jss-link.service.ts
@@ -4,11 +4,7 @@ import { HTMLLink } from '@sitecore-jss/sitecore-jss-angular';
 
 @Injectable()
 export class JssLinkService {
-  document: Document;
-
-  constructor() {
-    this.document = Inject(DOCUMENT);
-  }
+  constructor(@Inject(DOCUMENT) private document: Document) {}
 
   /**
    * Adds link element in the document head.
@@ -16,6 +12,11 @@ export class JssLinkService {
    */
   addHeadLinks(headLink: HTMLLink) {
     if (!headLink) {
+      return;
+    }
+
+    // Check if a link with the same rel and href already exists
+    if (this.isDuplicateLink(headLink)) {
       return;
     }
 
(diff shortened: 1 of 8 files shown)


packages/create-sitecore-jss/src/templates/angular-xmcloud/src/app/jss-link.service.ts
```diff
@@ -43,6 +43,9 @@ export class JssLinkService {
    * @param rel - The rel attribute of the links to be removed.
    */
   removeLinksByRel(rel: string) {
+    if (!this.document || !this.document.head) {
```
ambrauer, line 46: While this might be good to add, there appears to be an underlying issue with the document injection. I was testing this, and `this.document` was undefined when calling `addHeadLinks` as well. I believe the document needs to be injected in this way in the constructor:
```
constructor(
    @Inject(DOCUMENT) private document: Document
  ) {}
  ```

**Review by ambrauer — CHANGES_REQUESTED** [review]

File: packages/create-sitecore-jss/src/templates/angular/src/app/routing/routing.module.ts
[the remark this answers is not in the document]
Comment (sc-addypathania) [reply packages/create-sitecore-jss/src/templates/angular/src/app/routing/routing.module.ts:1]: but isn't it limited to routing i.e. layout?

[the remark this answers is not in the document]
Comment (sc-addypathania) [reply packages/create-sitecore-jss/src/templates/angular/src/app/routing/routing.module.ts:1]: Discussed. Should be in AppModule.

File: packages/create-sitecore-jss/src/templates/angular-xmcloud/src/app/jss-link.service.ts [hunk 2]
```diff
@@ -39,13 +40,14 @@ export class JssLinkService {
...
+   * @param headLink - An HTMLLink object to be checked.
+   * @returns {boolean} - True if a matching link exists, false otherwise.
    */
-  removeLinksByRel(rel: string) {
-    const links = this.document.head.querySelectorAll(`link[rel="${rel}"]`);
-    links.forEach((link) => {
-      this.document.head.removeChild(link);
-    });
+  private isDuplicateLink(headLink: HTMLLink): boolean {
+    const existingLink = this.document.querySelector(
```
art-alexeyenko, line 48: For this we'll be parsing a progressively larger document each time. 
Can this be limited to document.head only? Alternatively, is it possible to prevent duplicate links from being created?

sc-addypathania: Makes sense. Updated.

Yeah unfortunately Angular creates these canonical/accumulating links while navigating to different routes.
Previously, I tried to clear all the link when visiting a new route but unfortunately it even removed the main stylesheet of the app. So now the most we can do we remove duplicate links from the visited routes.

art-alexeyenko (approved):

PR #1887, opened 2024-08-13
[templates/nextjs-xmcloud] Disable image optimization for edit/preview

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Disable nextjs image optimization on NextImage component for edit and preview modes. Ensures better stability when rendering SXA images.

## Testing Details
- [ ] Unit Test Added

## Types of changes

### DIFF
--- a/packages/sitecore-jss-nextjs/src/components/NextImage.tsx
+++ b/packages/sitecore-jss-nextjs/src/components/NextImage.tsx
@@ -49,11 +49,9 @@ export const NextImage: React.FC<NextImageProps> = withFieldMetadata<NextImagePr
       if (!img) {
         return null;
       }
-      const {
-        sitecoreContext: { pageState },
-      } = useSitecoreContext();
       // disable image optimization for Edit and Preview
-      const unoptimized = pageState !== LayoutServicePageState.Normal;
+      const unoptimized =
+        otherProps.unoptimized !== undefined ? otherProps.unoptimized : field.metadata;
 
       const attrs = {
         ...img,
--- a/packages/sitecore-jss-nextjs/src/components/NextImage.tsx
+++ b/packages/sitecore-jss-nextjs/src/components/NextImage.tsx
@@ -7,11 +7,12 @@ import {
   ImageField,
   ImageFieldValue,
   withFieldMetadata,
+  useSitecoreContext,
 } from '@sitecore-jss/sitecore-jss-react';
 import Image, { ImageProps as NextImageProperties } from 'next/image';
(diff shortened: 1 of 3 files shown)


packages/sitecore-jss-nextjs/src/components/NextImage.tsx
```diff
@@ -48,6 +49,11 @@ export const NextImage: React.FC<NextImageProps> = withFieldMetadata<NextImagePr
       if (!img) {
         return null;
       }
+      const {
+        sitecoreContext: { pageState },
+      } = useSitecoreContext();
+      // disable image optimization for Edit and Preview
+      const unoptimized = pageState !== LayoutServicePageState.Normal;
```
ambrauer, line 56: In normal mode, we need to consider the customer may have specified `unopotimized={false}` (i.e. in `otherProps` here). We should use what they've specified in that case.

ambrauer (changes requested):
See one comment. Also, we should probably add a unit test?

packages/sitecore-jss-nextjs/src/components/NextImage.tsx
```diff
@@ -48,6 +49,9 @@ export const NextImage: React.FC<NextImageProps> = withFieldMetadata<NextImagePr
       if (!img) {
         return null;
       }
+      // disable image optimization for Edit and Preview
+      const unoptimized =
+        otherProps.unoptimized !== undefined ? otherProps.unoptimized : field.metadata;
```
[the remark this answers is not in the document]
ambrauer: Yeah, I'm confused by the assignment to `field.metadata`. I was expecting something like
```
const unoptimized = pageState === LayoutServicePageState.Normal ? otherProps.unoptimized || true : false;
```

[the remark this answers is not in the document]
art-alexeyenko: Yes, that will not work, but context was failing when I went for this. Made it work now though - should cover preview and edit.

packages/sitecore-jss-nextjs/src/components/NextImage.tsx
art-alexeyenko: It would be the other way around actually. unoptimized will be false in Normal mode by default. It is when customer set it to `true` we need to preserve the value. Adjusted for this case.

ambrauer (approved):
👍

packages/sitecore-jss-nextjs/src/components/NextImage.tsx
ambrauer: Ah, yep correct

PR #1918, opened 2024-09-05
[sitecore-jss-angular][angular-xmcloud] Metadata editing mode support for Placeholder

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
* Render "code" blocks with metadata for placeholder and placeholder renderings
* Modifies JssContetService logic. It now uses JssStateService injected in `sitecore-jss-angular`
* DEFAULT_PLACEHOLDER_UID was moved to `sitecore-jss` base package

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate) - skate park site

## Types of changes


packages/sitecore-jss-angular/src/services/models.ts
```diff
@@ -0,0 +1,17 @@
+import { LayoutServiceContextData, RouteData } from '../public_api';
+
+export class JssState {
```
illiakovalenko, line 3: Some definitions, like "LayoutServiceError", "serverRoute" should be defined in the app instead, since it fully depends on customers' implementation.
Here are a few questions for you to think about:
1. As a developer, If I'm using a different _LayoutServiceError_, how can I bypass the type that is defined here?
1. As a developer, If I'm using a custom JssState and some properties are not needed for me OR I want to set additional properties to the state, how can I integrate a new solution?
1. Think about existing customers and the customizations that they can have in their state

packages/sitecore-jss-angular/src/components/placeholder.component.ts
```diff
@@ -166,15 +175,90 @@ export class PlaceholderComponent implements OnInit, OnChanges, DoCheck, OnDestr
...
+      this.rendering &&
+      this.rendering.placeholders &&
+      Object.keys(this.rendering.placeholders).length > 0
+    ) {
+      return this.rendering.placeholders[phName];
+    }
+    return null;
+  }
+
+  getCodeBlockId = (kind: string, renderingId?: string): string | undefined => {
```
illiakovalenko, line 217: I'm wondering if it's better to create a `PlaceholderMetadata` class that will be extended by `PlaceholderComponent` class. In this case we can hide some initialization logic, variables, methods. Currently, everything is mixed here

illiakovalenko, line 217: Please, add JsDoc for new methods

packages/sitecore-jss-angular/src/components/placeholder.component.ts
```diff
@@ -265,17 +347,30 @@ export class PlaceholderComponent implements OnInit, OnChanges, DoCheck, OnDestr
       this.isLoading = false;
     } else {
       const factories = await this.componentFactory.getComponents(placeholder);
-
       try {
         const nonGuarded = await this.guardResolver(factories);
         const withData = await this.dataResolver(nonGuarded);
+        // not using index to ensure code blocks are rendered at correct positions
```
illiakovalenko, line 353: What happens in case we continue to use "index"? The index shouldn't affect and give you the wrong order. Also, I see that you made "index" optional in some functions, while we don't provide index anymore
```
index: The 0-based index at which to insert the new view into this container. If not specified, append the new view as the last entry.
```

packages/sitecore-jss-angular/src/components/placeholder.component.spec.ts
```diff
@@ -844,3 +897,320 @@ describe('SXA components', () => {
...
+      );
+      expect(elements[3].nativeElement.outerHTML).toBe(
+        '<code type="text/sitecore" class="scpm" kind="close" chrometype="placeholder"></code>'
+      );
+      expect(de.query(By.directive(MissingComponentComponent))).toBeDefined();
+    })
+  );
+});
+
+describe('Placeholder Metadata: dynamic placeholder:', () => {
```
illiakovalenko, line 1066:
```suggestion
describe('Placeholder Metadata: dynamic placeholder', () => {
```

illiakovalenko, line 1066: Need to verify the case:
```
renders placeholder blocks with DEFAULT_PLACEHOLDER_UID value when metadataType is dynamic placeholder and uid is not present
```

packages/sitecore-jss-angular/src/services/jss-state.service.ts
```diff
@@ -0,0 +1,23 @@
...
+  get state(): Observable<JssState> {
+    return this._state.asObservable();
+  }
+  private _state: BehaviorSubject<JssState>;
+
+  constructor(protected transferState: TransferState) {
+    this._state = new BehaviorSubject<JssState>(new JssState());
+  }
+
+  getStateValue() {
```
illiakovalenko, line 16: We can use getter as well to simplify the usage
```suggestion
  get stateValue() ...
```

docs/upgrades/unreleased.md
```diff
@@ -1,5 +1,40 @@
 ## Unreleased
 
+# Angular
+
+* Update several services to use updated Jss Context service:
```
illiakovalenko, line 5: As a reminder: make sure to revisit upgrade guide after the final changes

illiakovalenko (changes requested):
Good progress 👍 
Some comments:
* The application is broken if I initialize "angular", "angular-sxp". I can't start the app in "disconnected" mode.
And the comments below as well

CHANGELOG.md
```diff
@@ -41,9 +41,12 @@ Our versioning strategy is as follows:
   * Promo component ([#1897](https://github.com/Sitecore/jss/pull/1897))
   * PageContent component ([#1905](https://github.com/Sitecore/jss/pull/1905))
   * Navigation component ([#1894](https://github.com/Sitecore/jss/pull/1894))
+* `[sitecore-jss-angular]` Placeholder component will now render metadata code blocks when in Metadata Edit mode ([#1918](https://github.com/Sitecore/jss/pull/1918))
 
 ### 🛠 Breaking Change
 
+* `[templates/angular]` JssContextService now relies on JssStateService for state management. ([#1918](https://github.com/Sitecore/jss/pull/1918))
```
illiakovalenko, line 48: New _JssStateService_ should be mentioned in a changelog as a feature (since it's a new service) and explanation of the usage

packages/sitecore-jss-angular/src/components/placeholder.component.ts
```diff
@@ -52,41 +54,13 @@ import { PlaceholderLoadingDirective } from './placeholder-loading.directive';
 import { RenderEachDirective } from './render-each.directive';
 import { RenderEmptyDirective } from './render-empty.directive';
 import { isRawRendering } from './rendering';
+import { JssStateService } from '../services/jss-state.service';
 
 /**
  * @param {ComponentRendering} rendering
  * @param {string} name
  * @param {EditMode} [editMode]
  */
```
yavorsk, line 63: these can be removed

packages/sitecore-jss-angular/src/test-data/metadata-data.ts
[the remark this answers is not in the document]
art-alexeyenko: What do you mean? It's the same data we use for unit tests in sitecore-jss-react

packages/sitecore-jss-angular/src/components/placeholder.component.ts
art-alexeyenko: > What happens in case we continue to use "index"?

All code blocks will be rendered after all the components are added, instead of wrapping them

packages/sitecore-jss-angular/src/components/placeholder.component.spec.ts
art-alexeyenko: This is intentional. The way angular ng-test outputs unit test descriptions is `<Describe section> <it section>` instead of a more intuitive way of doing "describe" and "it" sections on two different lines. Having a colon here improves readability of test descriptions.

CHANGELOG.md
art-alexeyenko: But it is a breaking change that requires customers to adjust also?

packages/sitecore-jss-angular/src/components/placeholder.component.ts
art-alexeyenko: I've had trouble for some time previously navigating between Placeholder and PlaceholderCommon in React when debugging. 
And for all the mixing in, I believe this to be a better, more visible approach, with a single placeholder component. What you see in it is what you get, without splitting implementation between multiple files.

CHANGELOG.md
illiakovalenko: It's a new service that we introduced, so we need to provide more information on that, where it's available and why it was introduced. Customer can think that this change is not required to be done by them, currently there is no mentioning that this service is referenced by the components that are introduced on SDK level
I agree that it's a breaking change, I forgot that we started to reference this service in our SDK as well.

packages/sitecore-jss-angular/src/test-data/metadata-data.ts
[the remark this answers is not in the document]
illiakovalenko: The format of the examples that you introduced - they represent the normal mode data.
(Except of "editMode: metadata", that you added). "Metadata" format is.
```
{
  field: {
     metadata: { ... }
  }
}
```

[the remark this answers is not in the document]
illiakovalenko: I understand that we have the same in sitecore-jss-react, but it doesn't mean that we need to follow exactly the same if the data shape is incorrect

packages/create-sitecore-jss/src/templates/angular/src/app/app.module.ts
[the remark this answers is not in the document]
art-alexeyenko: indeed. removed

packages/create-sitecore-jss/src/templates/angular/src/app/jss-context.service.ts
```diff
@@ -16,14 +15,17 @@ export const jssKey = makeStateKey<JssState>('jss');
 export class JssContextService {
   // components can subscribe to this (or use getValue()) to get access to latest data from Layout Service,
   // as well as current language and server route
-  state: BehaviorSubject<JssState>;
-
-  constructor(protected transferState: TransferState, protected layoutService: JssLayoutService) {
-    this.state = new BehaviorSubject<JssState>(new JssState());
+  get state() {
```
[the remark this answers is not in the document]
art-alexeyenko: This is done to make upgrades easier. All the jss context usage is currently tied to JssContextService. Less hassle if instead of new injection/import we provide the option to reuse the same class, with slightly different getter.

packages/sitecore-jss-angular/src/services/models.ts
art-alexeyenko: All of this can be achieved with Angular's dependency injection, but you're right, it will be a pain.
I've modified the approach to still have the connection between packages state and app state, but JssState is back into sample

packages/sitecore-jss-angular/src/components/placeholder.component.ts
art-alexeyenko: And I'm still wary, not sure if this is a good idea. Angular has a lot of quirks around its template. The whole functionality resolves around the private view reference. I did a quick check - compiler doesn't like when view elements are defined on both levels. I think we would need the child to have its own view element in template ob child, but handle it the same as parent most of the time in _render() method. 
Or alternatively move placeholder code blocks renders from template into code which I also would like to avoid - this the only visual representation of code blocks in template, the only indication this is how they will be used.

packages/sitecore-jss-angular/src/test-data/metadata-data.ts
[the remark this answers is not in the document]
art-alexeyenko: I added them as commented out block. To be uncommented with Fields work - it causes type mismatch now.

art-alexeyenko:
@illiakovalenko addressed your comments except
> Update the PR description, since the description structure is messy

I have no idea what happened, this our regular template with the most regular raw text value imaginable.

illiakovalenko:
@art-alexeyenko I've updated, it's not allowed to use "code" tag directly in a description

packages/sitecore-jss-angular/src/test-data/metadata-data.ts
```diff
@@ -0,0 +1,145 @@
...
+      pageEditing: true,
+      editMode: EditMode.Metadata,
+      serverRoute: '/',
+      language: 'en',
+    },
+    route: {
+      name: 'main',
+      componentName: '',
+      fields: {
+        /* TODO: uncomment when doing Fields rework. Type mismatches otherwise
```
illiakovalenko, line 17: We can uncomment right now, since this object doesn't have exact type. Also the shape that should be is:
```
text: {
  metadata: {
    fieldId: '123',
    fieldType: 'text',
    rawValue: 'textField',
  },
},
```
Apply the same change for the comments added below in this file

packages/sitecore-jss/src/editing/utils.ts
```diff
@@ -1,5 +1,11 @@
 import isServer from '../utils/is-server';
 
+/**
+
```
illiakovalenko, line 4:
```suggestion
```

packages/sitecore-jss/src/editing/utils.ts
```diff
@@ -1,5 +1,11 @@
...
+ * default value of uid for root placeholder when uid is not present.
```
illiakovalenko, line 5:
```suggestion
 * Default value of uid for root placeholder when uid is not present.
```

packages/sitecore-jss-angular/src/components/placeholder.component.spec.ts
```diff
@@ -844,3 +896,368 @@ describe('SXA components', () => {
...
+        '<code type="text/sitecore" class="scpm" kind="close" chrometype="rendering"></code>'
+      );
+      expect(elements[3].nativeElement.outerHTML).toBe(
+        '<code type="text/sitecore" class="scpm" kind="close" chrometype="placeholder"></code>'
+      );
+    })
+  );
+
+  it(
+    'should render code blocks around nested placeholder components with provided placeholder ID',
```
illiakovalenko, line 973: This unit test doesn't cover the case that we expected to cover. It doesn't render nested placeholder and doesn't have "assert" statement for it

packages/sitecore-jss-angular/src/components/placeholder.component.spec.ts
```diff
@@ -844,3 +896,368 @@ describe('SXA components', () => {
...
+
+      comp = fixture.componentInstance;
+      const jssContext = de.injector.get(JssStateService);
+      jssContext.setState(layoutData);
+      fixture.detectChanges();
+    })
+  );
+
+  it(
+    'should render code blocks around nested placeholder components with DEFAULT_PLACEHOLDER_UID',
```
illiakovalenko, line 943: This unit test doesn't cover the case that we expected to cover. It doesn't render nested placeholder and doesn't have "assert" statement for it

packages/sitecore-jss-angular/src/components/placeholder.component.spec.ts
illiakovalenko, line 1: For all the unit tests, is it possible just to test the exact "innerHtml" markup? Like we did in React SDK.
The current approach doesn't give you a guarantee that the actual components are rendered in the right place, we are able to verify only "code" blocks and still we are not sure if there is something between these tags
In order to remove "angular tags" from "innerHTML" you can easily do that by using
```
innerHTML.replace(/<!--[\s\S]*?-->/g, '');
```

illiakovalenko (changes requested):

CHANGELOG.md
```diff
@@ -45,13 +45,17 @@ Our versioning strategy is as follows:
   * Promo component ([#1897](https://github.com/Sitecore/jss/pull/1897))
   * PageContent component ([#1905](https://github.com/Sitecore/jss/pull/1905))
   * Navigation component ([#1894](https://github.com/Sitecore/jss/pull/1894))
+* `[sitecore-jss-angular]` Placeholder component will now render metadata code blocks when in Metadata Edit mode ([#1918](https://github.com/Sitecore/jss/pull/1918))
 
 ### 🛠 Breaking Change
 
+
+*  `[templates/angular]` JssContextService now relies on JssStateService for state management. This allows for improved type safety and reverts the previous anti-pattern of modifying `state` object directly ([#1918](https://github.com/Sitecore/jss/pull/1918))
```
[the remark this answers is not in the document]
illiakovalenko: Also, since we introduced a new feature, we need to mention where it's exposed currently we don't have enough information here

packages/sitecore-jss-angular/src/components/placeholder.component.ts
illiakovalenko: Ok, let's leave it
My main request was to analyze if it's possible to improve it

packages/sitecore-jss-angular/src/components/placeholder.component.spec.ts
```diff
@@ -602,7 +628,21 @@ describe('<sc-placeholder /> with lazy loaded modules', () => {
             ]
           ),
         ],
-        providers: [MockUrlTreeGuard, MockUnexpectedErrorGuard],
+        providers: [
+          MockUrlTreeGuard,
+          MockUnexpectedErrorGuard,
+          JssStateService,
+          {
```
illiakovalenko, line 635: Can you double-check if this provider is needed? I believe that it's unused, we don't have such injection token

packages/sitecore-jss-angular/src/services/jss-state.service.ts
```diff
@@ -0,0 +1,45 @@
...
+  };
+  viewBag?: { [key: string]: unknown };
+}
+
+/**
+ * The service that is used to store and retrieve the current state of the app.
+ * Referenced by the SDK to get the current state of the app.
+ */
+@Injectable()
+export class JssStateService<State extends BaseJssState = BaseJssState> {
```
illiakovalenko, line 17: We don't have any tests for the new service, please, consider to add them

packages/sitecore-jss-angular/src/components/placeholder.component.spec.ts
sc-addypathania: Updated.

packages/sitecore-jss-angular/src/components/placeholder.component.spec.ts
sc-addypathania: fixed.

packages/sitecore-jss-angular/src/components/placeholder.component.spec.ts
sc-addypathania: fixed.

CHANGELOG.md
```diff
@@ -58,6 +59,7 @@ Our versioning strategy is as follows:
 
 ### 🛠 Breaking Change
 
+*  `[templates/angular]` `[sitecore-jss-angular]` A new `JssStateService` has been introduced in the angular package (_src/service/jss-state.service.ts_), enhancing type safety and enabling state sharing across the application and SDK components. As a result, the JssContextService now relies on JssStateService for state management.([#1918](https://github.com/Sitecore/jss/pull/1918))
```
illiakovalenko, line 62:
```suggestion
*  `[templates/angular]` `[sitecore-jss-angular]` A new `JssStateService` has been introduced in the `sitecore-jss-angular` package, enhancing type safety and enabling state sharing across the application and SDK components. As a result, the JssContextService now relies on JssStateService for state management.([#1918](https://github.com/Sitecore/jss/pull/1918))
```

illiakovalenko (approved):

PR #1937, opened 2024-10-01
[sitecore-jss-angular] Fix default empty field components to not render unwanted wrapping HTML tags

addy-pathania:
- [x] Changelog updated

## Description / Motivation
The empty field scenarios for Link and Image breaks Pages editor because angular components returns extra wrapper tags. Pages Editor does not accept these wrapper tags therefore this fix removes those unwanted wrappers.

## Testing Details

- [ ] Unit Test Added

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,7 @@ Our versioning strategy is as follows:
 
 - Patch: no breaking changes (e.g. bug fixes, minor improvements)
 - Minor: may include breaking changes in framework packages (e.g. framework upgrades, new features, improvements)
-- Major: may include breaking changes in core packages (e.g. major architectural changes, major features)
-
+- Major: may include breaking changes in core packages (e.g. major architectural changes, major features
 ## Unreleased
 
 ### 🐛 Bug Fixes
@@ -16,6 +15,8 @@ Our versioning strategy is as follows:
 * `[templates/nextjs]` `[templates/react]` `[templates/angular]` `[templates/vue]` Fixed an issue when environment variable is undefined (not present in .env), that produced an "undefined" value in generated config file ([#1875](https://github.com/Sitecore/jss/pull/1875))
 * `[templates/nextjs]` Fix embedded personalization not rendering correctly after navigation through router links. ([#1911](https://github.com/Sitecore/jss/pull/1911))
 * `[template/angular]` Prevent client-side dictionary API call when SSR data is available ([#1930](https://github.com/Sitecore/jss/pull/1930)) ([#1932](https://github.com/Sitecore/jss/pull/1932))
+* `[sitecore-jss-angular]` Fix default empty field components to not render the unwanted wrapping tags ([#1937](https://github.com/Sitecore/jss/pull/1937))
+
 
 ### 🎉 New Features & Improvements
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Our versioning strategy is as follows:
 - Patch: no breaking changes (e.g. bug fixes, minor improvements)
 - Minor: may include breaking changes in framework packages (e.g. framework upgrades, new features, improvements)
 - Major: may include breaking changes in core packages (e.g. major architectural changes, major features
+
 ## Unreleased
 
 ### 🐛 Bug Fixes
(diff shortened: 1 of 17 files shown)


yavorsk (changes requested):
Looks much cleaner than initial iterations - to keep everything in the default components. Some tests are failing and need to be updated - probably bc the tested element is not located properly after the changes

packages/sitecore-jss-angular/src/components/default-empty-text-field-editing-placeholder.component.ts
```diff
@@ -1,10 +1,27 @@
-import { Component } from '@angular/core';
+import { Component, ElementRef, OnInit, Renderer2 } from '@angular/core';
 
 /**
  * Default component that will be rendered in pages when field is empty; applies for text, richtext, date and link fields.
  */
 @Component({
   selector: 'sc-default-empty-text-field-editing-placeholder',
```
illiakovalenko, line 7: I'm trying to figure out the fix...so according to the item description

> Pages expects link field to be anchor or span element, but in Angular the technical sc-default-empty-text-field-editing-placeholder node is rendered as wrapper. Consider rework the DefaultEmptyFieldEditingComponent to use attribute selector and not generate excesive wrapper

So, you was able to use "attribute" selector for "default-image-field", however, here we have some tricks. Why we can't do the same?
In "ngOnInit" you are trying to remove base template (and wrapper?) and trying to re-create the same by adding "span"? Why we can't leave "span" in a template as it was before?
I think in general DOM manipulation is not a right way that we need to follow, it's more like a hack

illiakovalenko (changes requested):

packages/sitecore-jss-angular/src/components/default-empty-text-field-editing-placeholder.component.ts
addy-pathania: yes that's what I thought initially that making the component an attribute selector would make it work for both components but for Link, Pages editor only accepts span or anchor tag and after changing the component to attribute selector it wraps the component with div tag which again break Pages editor. Somehow we don't have the same issue for Image.

illiakovalenko: So Pages doesn't accept "div" wrapper in all the cases except of "Image"? We need to figure out that. Can you clarify that?
If it's true, can we just render it still as "attribute selector" and remove extra "div" instead?

addy-pathania: So div tags are not acceptable only for link component, for rest it is acceptable.
I have refactored the implementation in a more cleaner way. Please check.

packages/sitecore-jss-angular/src/components/default-empty-text-field-editing-placeholder.component.ts
```diff
@@ -4,7 +4,7 @@ import { Component } from '@angular/core';
  * Default component that will be rendered in pages when field is empty; applies for text, richtext, date and link fields.
  */
 @Component({
-  selector: 'sc-default-empty-text-field-editing-placeholder',
-  template: '<span>[No text in field]</span>',
+  selector: '<span>[sc-default-empty-text-field-editing-placeholder]</span>',
```
illiakovalenko, line 7: Is it possible to render selector by using this approach?
How it is rendered in HTML?

addy-pathania: Yes it is.
image

illiakovalenko: Cool, this looks much better!

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ Our versioning strategy is as follows:
 * `[templates/nextjs]` `[templates/react]` `[templates/angular]` `[templates/vue]` Fixed an issue when environment variable is undefined (not present in .env), that produced an "undefined" value in generated config file ([#1875](https://github.com/Sitecore/jss/pull/1875))
 * `[templates/nextjs]` Fix embedded personalization not rendering correctly after navigation through router links. ([#1911](https://github.com/Sitecore/jss/pull/1911))
 * `[template/angular]` Prevent client-side dictionary API call when SSR data is available ([#1930](https://github.com/Sitecore/jss/pull/1930)) ([#1932](https://github.com/Sitecore/jss/pull/1932))
+* `[sitecore-jss-angular]` Fix default empty field components to not render the unwanted wrapping tags ([#1937](https://github.com/Sitecore/jss/pull/1937))
+
```
illiakovalenko, line 20: Remove extra line

illiakovalenko (approved):
Appoved! However make sure to update the unit tests to make build green

yavorsk (approved):

PR #1947, opened 2024-10-10
[Angular][XMC] Ensure context is available at the time of placeholder rendering

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
ngOnChanges executes before ngOnInit within placeholder component. This is [expected](https://angular.dev/guide/components/lifecycle#ngonchanges) (but not obvious).
Considering we need metadataMode value to be available already, the context subscription is moved to constructor (can't use it in OnChanges as it's called multiple times).

## Testing Details
- [ ] Unit Test Added

## Types of changes


packages/sitecore-jss-angular/src/components/placeholder.component.spec.ts
```diff
@@ -1269,4 +1269,20 @@ describe('Placeholder Metadata: dynamic placeholder:', () => {
       expect(cleanedRenderedHTML).toEqual(expectedHTML);
     })
   );
+
+  fit(
```
yavorsk, line 1273: this should be "it(..." I guess?

art-alexeyenko: ah yes, I thought I cleaned up all spec changes. thanks for catching this

PR #1948, opened 2024-10-11
[sitecore-jss-angular] Fix Editing Scripts component rendered twice in pages 

yavorsk:
- [x] Changelog updated

## Description / Motivation

This PR fixes issue where editing scripts are rendered twice (for server and client) in pages edit mode metadata - by rendering the scripts only on the server.

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate) - setup local app as rendering host for pages, and make sure client editing scripts are rendered only once in the markup and there is no 'duplicate modules' error in the browser console

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,7 @@ Our versioning strategy is as follows:
    - `scRichTextEmptyFieldEditingTemplate` for _scRichText_
    - `scRouterLinkEmptyFieldEditingTemplate` for _scRouterLink_
    - `scTextEmptyFieldEditingTemplate` for _scText_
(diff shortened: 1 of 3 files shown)


stasmaxymov:
@yavorsk Is there a way to run Angular in full SPA mode, without server-side rendering? If yes, the fix might not work with it

yavorsk:
hey @stasmaxymov , yea I was also thinking about this. In editing mode we always have SSR and since we are rendering these scripts for editing mode only anyway I think it should be fine

art-alexeyenko (approved):
This looks good, thank you Yavor.
One thing I'd change is having to set a spy in each test - I'll adjust this in a moment.

PR #1963, opened 2024-11-04
[nextjs][xmcloud]  Introduce property `enablePersonalizeCookie` to PersonalizeMiddlewareConfig

yavorsk:
- [x] Changelog updated

## Description / Motivation

This PR adds a new property `enablePersonalizeCookie` to PersonalizeMiddlewareConfig and pass it to CloudSDK.addPersonalize() function, which is required to enable A/B testing and personalization; update personalize plugin to set it to `true` by default. 

## Testing Details

- [ ] Unit Test Added

## Types of changes

### DIFF
--- a/packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
+++ b/packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
@@ -46,6 +46,10 @@ export type PersonalizeMiddlewareConfig = MiddlewareBaseConfig & {
    */
   cdpConfig: CdpServiceConfig;
   /**
+   * Flag to set the enablePersonalizeCookie setting of cloud sdk personalize; if omitted, defaults to false
+   */
+  enablePersonalizeCookie?: boolean;
+  /**
    * Optional Sitecore Personalize scope identifier allowing you to isolate your personalization data between XM Cloud environments
    */
   scope?: string;
@@ -127,7 +131,7 @@ export class PersonalizeMiddleware extends MiddlewareBase {
       cookieDomain: hostname,
       enableServerCookie: true,
     })
-      .addPersonalize()
+      .addPersonalize({ enablePersonalizeCookie: this.config.enablePersonalizeCookie ?? false})
       .initialize();
   }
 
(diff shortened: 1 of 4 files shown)


packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -127,7 +131,7 @@ export class PersonalizeMiddleware extends MiddlewareBase {
       cookieDomain: hostname,
       enableServerCookie: true,
     })
-      .addPersonalize()
+      .addPersonalize({ enablePersonalizeCookie: this.config.enablePersonalizeCookie ?? false})
```
kendoce, line 134: shouldn't we set it to true if it's undefined so that we will not have breaking changes? they can just set the value to false if they want to opt out.

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -45,6 +45,10 @@ export type PersonalizeMiddlewareConfig = MiddlewareBaseConfig & {
    * Configuration for your Sitecore CDP endpoint
    */
   cdpConfig: CdpServiceConfig;
+  /**
+   * Flag to set the enablePersonalizeCookie setting of cloud sdk personalize; if omitted, defaults to false
+   */
+  enablePersonalizeCookie?: boolean;
```
art-alexeyenko, line 51: what happens when this cookie is absent? Is it only the A/B that doesn't work, or would personalize rules based on visitor ID be affected too?
Overall perhaps it makes sense to always set this cookie, unless personalization middleware is disabled in the seprate setting.

art-alexeyenko: i.e. could this cookie affect "The visitor has visited page name(s) during the current visit" and other similar rules?

yavorsk: I was thinking to provide a link to the csdk documentation for more details, because we are only just passing this through..?
You mean that opt out can be done via the disable function?

packages/sitecore-jss-nextjs/src/middleware/personalize-middleware.ts
yavorsk: makes sense, I will update

PR #1984, opened 2024-11-26
[sitecore-jss-angular] [template/angular] Integrate CloudSDK events firing

addy-pathania:
- [x] Changelog updated

## Description / Motivation
Send event info to CloudSdk when different interactions happen with the form component e.g. INTERACTED, SUBMITTED etc.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/sitecore-jss-angular/setup.js
```diff
@@ -0,0 +1,5 @@
+﻿window.process = {
+  env: {
+    NODE_ENV: 'production',
```
art-alexeyenko: Can you elaborate on this change? Do you envision all tests be performed in production context?

packages/sitecore-jss-angular/src/components/form.component.ts
```diff
@@ -136,6 +138,32 @@ export class FormComponent implements OnInit, OnDestroy {
     }
   }
 
+  /**
+   * Subscribes to the form submit event and sends data to CloudSDK.
+   */
+  subscribeToFormSubmitEvent() {
```
art-alexeyenko, line 145: strictly speaking, it's not only the submit event - more like 'engage' event

packages/create-sitecore-jss/src/templates/angular/webpack.config.js
```diff
@@ -1,7 +1,11 @@
...
 
 module.exports = {
-  plugins: [
-    new Dotenv(),
-  ],
-}
+  plugins: [new Dotenv()],
+  resolve: {
+    alias: {
+      '@sitecore-cloudsdk': path.resolve(__dirname, 'node_modules/@sitecore-cloudsdk'),
```
art-alexeyenko, line 8: is there a way to only apply this on monorepo?

packages/sitecore-jss-angular/src/components/form.component.spec.ts
```diff
@@ -173,6 +180,41 @@ describe('FormComponent', () => {
...
+    const consoleWarnSpy = spyOn(console, 'warn').and.callThrough();
+
+    component.subscribeToFormSubmitEvent();
+
+    expect(consoleWarnSpy).toHaveBeenCalledWith(
+      'No form element found to subscribe to submit event.'
+    );
+  });
+
+  it('should not trigger the form event if the component is in editing or preview mode', () => {
```
art-alexeyenko, line 195: is there a way to also test if form function was called?

art-alexeyenko (changes requested):
Great work Addy, especially in figuring out the monorepo problems with cloudsdk.
I left a few comments to refine the PR to completion.

packages/create-sitecore-jss/src/templates/angular/webpack.config.js
addy-pathania: I think so but then we might need to setup a global or maybe shared webpack configuration which is out of scope for this task.
Worth exploring when we setup Nx.

packages/sitecore-jss-angular/setup.js
addy-pathania: Added description.
No, i updated it to `test`.

packages/sitecore-jss-angular/src/components/form.component.ts
addy-pathania: makes sense.

packages/create-sitecore-jss/src/templates/angular/webpack.config.js
art-alexeyenko: Can you have another go at it? It would be wrong to ship this part to end users, if alternative exists. It can potentially break customers' monorepo setups too, when dependencies are hoisted

addy-pathania: But how? it's just an alias which is telling webpack how to resolve module paths during bundling. It's not changing how dependencies are installed or managed in node_modules. Dependency hoisting is concept associated with package managers.

art-alexeyenko: The scenario I have in mind is this: 
1. `@sitecore/cloudsdk` installed outside of angular folder
2. webpack cannot resolve `node_modules/@sitecore-cloudsdk` from within angular folder because of this
3. Error?

This is my last concern, I'll approve the PR, but please only merge after you verified this webpack configuration works with our pnpm setup (angular builds, runs in dev and prod modes).

art-alexeyenko (approved):
Approved, but please double-check the webpack config will not conflict with pnpm setup we have for starter.

packages/create-sitecore-jss/src/templates/angular/webpack.config.js
addy-pathania: `@sitecore/cloudSdk` is installed within the angular app and not outside.
This webpack configuration just makes sure that in case of multiple `@sitecore/cloudsdk` instances it gets resolved to one. It is working fine in my monorepo setup but we will be able to test for others once this gets merged.

addy-pathania:
> Approved, but please double-check the webpack config will not conflict with pnpm setup we have for starter.

I discussed this with Yavor, and we’ll anyway need the canary to test this with pnpm. I’ll proceed with merging it now and create a follow-up issue if any problems arise.

PR #2022, opened 2025-01-27
[sitecore-jss] Introduce separate flow for fetching streams

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Follow up for https://github.com/Sitecore/jss/pull/2020

Previous approach introduced a bug where every fetch operation response was considered a stream - because `response.body` always has type `ReadableStream`. 
An alterative approach is to explicitly state an intent of getting a stream - so we optionally pass a resneponse handler into the `fetch` operation and get what we need.

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Our versioning strategy is as follows:
 ### 🐛 Bug Fixes
 
 * `[sitecore-jss-nextjs]` Fix Chromes editing mode when rendering host URL is internally redirected in XMCloud ([#2019](https://github.com/Sitecore/jss/pull/2019))
-* `[sitecore-jss]` NativeDataFetcher doesn't return stream response ([#2020](https://github.com/Sitecore/jss/pull/2020))
+* `[sitecore-jss]` NativeDataFetcher doesn't return stream response ([#2020](https://github.com/Sitecore/jss/pull/2020)[#2022](https://github.com/Sitecore/jss/pull/2022))
 
 ## 22.4.0
 
--- a/packages/create-sitecore-jss/src/templates/nextjs-sxa/src/pages/api/sitemap.ts
+++ b/packages/create-sitecore-jss/src/templates/nextjs-sxa/src/pages/api/sitemap.ts
@@ -36,7 +36,7 @@ const sitemapApi = async (
 
     try {
       const fetcher = new NativeDataFetcher();
-      const response = await fetcher.fetch<ReadableStream<Uint8Array>>(sitemapUrl);
+      const response = await fetcher.fetchStream<Uint8Array>(sitemapUrl);
 
       const reader = response.data.getReader();
       if (reader) {
--- a/packages/sitecore-jss/src/native-fetcher.test.ts
+++ b/packages/sitecore-jss/src/native-fetcher.test.ts
@@ -145,6 +145,15 @@ describe('NativeDataFetcher', () => {
       expect(response.data).to.equal('response override');
     });
 
+    it('fetchStream should return response.body', async () => {
+      const fetcher = new NativeDataFetcher();
+
+      spy.on(global, 'fetch', mockFetch(200, { body: 'response body' }, { responseType: 'text' }));
+
+      const response = await fetcher.fetch('http://test.com/api');
+      expect(response.data).to.equal('response body');
+    });
+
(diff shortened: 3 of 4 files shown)


packages/sitecore-jss/src/native-fetcher.test.ts
```diff
@@ -136,6 +136,15 @@ describe('NativeDataFetcher', () => {
       expect(fetchInit?.body).to.be.undefined;
     });
 
+    it('should execute request with custom response parser', async () => {
+      const fetcher = new NativeDataFetcher();
+
+      spy.on(global, 'fetch', mockFetch(200, {}, { responseType: 'text' }));
+
+      const response = await fetcher.fetch('http://test.com/api', {}, () => 'response override');
```
addy-pathania, line 144: we should also add test for testing the actual method i.e fetchStream

packages/sitecore-jss/src/native-fetcher.ts
```diff
@@ -111,6 +121,16 @@ export class NativeDataFetcher {
     }
   }
 
+  async fetchStream<T>(
```
addy-pathania, line 133: add jsDoc

packages/sitecore-jss/src/native-fetcher.ts
```diff
@@ -111,6 +121,16 @@ export class NativeDataFetcher {
...
+    url: string,
+    options: RequestInit = {}
+  ): Promise<NativeDataFetcherResponse<ReadableStream<T>>> {
+    const parseStreamResponse = (response: Response) => {
```
addy-pathania, line 137: how are we handling this in case of an error?

packages/sitecore-jss/src/native-fetcher.ts
```diff
@@ -49,6 +49,11 @@ export type NativeDataFetcherFunction<T> = (
   data?: RequestInit
 ) => Promise<NativeDataFetcherResponse<T>>;
 
+export type ResponseParserFunc = (
```
addy-pathania, line 55: add jsDoc

addy-pathania (changes requested):
Please see some comments.
Also, from my understanding this would be a breaking change and we wont be able to release a patch with this change.
Can't we do anything simple by directly processing the `ReadStream`  in the previous if block the way we do in `Sitemap.ts` api file?

packages/create-sitecore-jss/src/templates/nextjs-sxa/src/pages/api/sitemap.ts
```diff
@@ -36,7 +36,7 @@ const sitemapApi = async (
 
     try {
       const fetcher = new NativeDataFetcher();
-      const response = await fetcher.fetch<ReadableStream<Uint8Array>>(sitemapUrl);
+      const response = await fetcher.fetchStream<Uint8Array>(sitemapUrl);
```
addy-pathania, line 39: I think the bug introduced by previous PR was in the [editing-render-middelware](https://github.com/Sitecore/jss/blob/9b0b19aaeb6c756fee9c1a603cb875120d285d3a/packages/sitecore-jss-nextjs/src/editing/editing-render-middleware.ts#L136). You would need to update the reference there too.

CHANGELOG.md
```diff
@@ -16,7 +16,7 @@ Our versioning strategy is as follows:
 ### 🐛 Bug Fixes
 
 * `[sitecore-jss-nextjs]` Fix Chromes editing mode when rendering host URL is internally redirected in XMCloud ([#2019](https://github.com/Sitecore/jss/pull/2019))
-* `[sitecore-jss]` NativeDataFetcher doesn't return stream response ([#2020](https://github.com/Sitecore/jss/pull/2020))
+* `[sitecore-jss]` NativeDataFetcher doesn't return stream response ([#2020](https://github.com/Sitecore/jss/pull/2020)[#2022](https://github.com/Sitecore/jss/pull/2022))
```
addy-pathania, line 19: Should this be under the breaking change section?

packages/sitecore-jss/src/native-fetcher.ts
art-alexeyenko: there's no error to be had, we only return a value

PR #2053, opened 2025-03-17
[Nextjs] Move ComponentLibrary from a separate route to Layout.tsx

addy-pathania:
- [x] Changelog updated

## Description / Motivation
Customers typically customize catch-all route and `Layout.tsx` with various things like React contexts.
That mean they need to re-apply those modification on the `DesignLibrary` route too, which makes user experience worse.
- Removed `component-library/render` file route and now redirectd to `/`
- Renamed `ComponentLibraryLayout` to `DesignLibrary` for distinction of Layouts and all related utils and functions.
- All rendering logic has been unified into a single `Layout.tsx` which now handles the rendering of `DesignLibrary` based on `renderingType`

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/create-sitecore-jss/src/templates/nextjs-sxa/src/Layout.tsx
```diff
@@ -31,24 +32,45 @@ const Layout = ({ layoutData, headLinks }: LayoutProps): JSX.Element => {
     <>
       <Scripts />
       <Head>
-        <title>{fields?.Title?.value?.toString() || 'Page'}</title>
-        <link rel="icon" href={`${publicUrl}/favicon.ico`} />
+        <title>
+        <% if (xmcloud) {-%>
+          {layoutData.sitecore.context.renderingType === 'component'
```
stasmaxymov: Is there any need for adjusting a Page title? It will be used only inside iframe, so it won't be visible. On the other hand it complicates Layout with additional conditions

packages/create-sitecore-jss/src/templates/nextjs-sxa/src/Layout.tsx
```diff
@@ -31,24 +32,45 @@ const Layout = ({ layoutData, headLinks }: LayoutProps): JSX.Element => {
...
-        </header>
-        <main>
-          <div id="content">{route && <Placeholder name="headless-main" rendering={route} />}</div>
-        </main>
-        <footer>
-          <div id="footer">{route && <Placeholder name="headless-footer" rendering={route} />}</div>
-        </footer>
+      <% if (xmcloud) {-%>
+      {layoutData.sitecore.context.renderingType === 'component' ? (
+        <ComponentLibrary {...layoutData} />
```
stasmaxymov: In the ticket description there was a proposal to renderer component library inline, rather than conditionally switch between component and regular layout. The assumption was, `ComponentLibrary` itself can check if it is a regular rendering and output `null`. And the some goes with regular `Placeholders` in the `Layout`, assumption that they output nothing when there is no components in them.
Did you try this approach?
 ```suggestion        
        <div id="content">
                <ComponentLibrary {...layoutData} />
                {route && <Placeholder name="headless-main" rendering={route} />}
         </div>
```

**Comment by stasmaxymov** [comment]
I think there are missing changes to `editing-render-middleware.ts`, as now it should redirect to a root path intead of  `/component-library/render`

**Review by stasmaxymov — CHANGES_REQUESTED** [review]

File: packages/create-sitecore-jss/watch.json.example [hunk 3]
```diff
@@ -12,9 +12,12 @@
     "language": "da-DK",
     "fetchWith": "GraphQL",
     "prerender": "SSG",
+    // when using the "nextjs-xmcloud" add-on, set "xmcloud" to true
+    "xmcloud": false,
```
art-alexeyenko: It's better to ensure 'xmcloud' is being set when xmcloud initializer is invoked on `create-sitecore-jss` side, rather than rely on xmcloud being explicitly set in yarn.watch _and_ `npx ...` command.
But if ComponentLibrary component is added inline, perhaps we won't need this at all.

addy-pathania: We are not using it inline and so I have changed the check from `xmcloud` to `templates.inlcudes(...)`
Tested it, works fine.

packages/create-sitecore-jss/src/templates/nextjs-sxa/src/Layout.tsx
addy-pathania: Yes i did try this, it works but then in library mode it tries to render the `headless-header`, `headless-main` and `headless-footer` which are empty and complains about them in the logs

stasmaxymov (approved):

packages/create-sitecore-jss/src/templates/nextjs-sxa/src/Layout.tsx
```diff
@@ -3,14 +3,15 @@
...
+import {
+  Placeholder,
+  LayoutServiceData,
+  Field,
+  DesignLibrary,
+  HTMLLink,
+} from '@sitecore-jss/sitecore-jss-nextjs';
 import Scripts from 'src/Scripts';
 
-// Prefix public assets with a public URL to enable compatibility with Sitecore Experience Editor.
```
art-alexeyenko: nextjs-sxa can still be used in SXP and ExpEditor. Let's keep this part

packages/sitecore-jss-react/src/components/DesignLibrary.test.tsx
```diff
@@ -84,11 +86,24 @@ describe('<ComponentLibraryLayout />', () => {
...
 
+  it('should render null when renderType is not "component"', () => {
+    const basicPage = getTestLayoutData(false, 'page');
+    rendered = mount(
+      <SitecoreContext componentFactory={componentFactory}>
+        <DesignLibrary {...basicPage.layoutData} />
+      </SitecoreContext>
+    );
+
+    console.log(rendered.html());
```
art-alexeyenko, line 95: extra log here

packages/sitecore-jss-react/src/components/DesignLibrary.test.tsx
```diff
@@ -84,11 +86,24 @@ describe('<ComponentLibraryLayout />', () => {
     );
   });
 
+  it('should render null when renderType is not "component"', () => {
```
art-alexeyenko: Probably a typo - should spell 'renderingType'

packages/create-sitecore-jss/src/templates/nextjs-sxa/src/Layout.tsx
```diff
@@ -32,23 +33,36 @@ const Layout = ({ layoutData, headLinks }: LayoutProps): JSX.Element => {
...
-          <div id="header">{route && <Placeholder name="headless-header" rendering={route} />}</div>
-        </header>
-        <main>
-          <div id="content">{route && <Placeholder name="headless-main" rendering={route} />}</div>
-        </main>
-        <footer>
-          <div id="footer">{route && <Placeholder name="headless-footer" rendering={route} />}</div>
-        </footer>
+      <% if (templates.includes('nextjs-xmcloud')) {-%>
+      {layoutData.sitecore.context.renderingType === 'component' ? (
```
art-alexeyenko: For this and the comparison in Bootstrap.tsx, should we perhaps use the `RenderingType` enum instead?

art-alexeyenko (approved):
I left some minor comments for your consideration, and please make sure azure pipeline completes.
Regardles, approved. Good job and well done Addy.

packages/create-sitecore-jss/src/templates/nextjs-sxa/src/Layout.tsx
addy-pathania: updated

PR #2056, opened 2025-03-19
[sitecore-jss-nextjs] Ensure locales are compared case-insensitively in redirects middleware

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
We must ensure redirect rules specified like 'pl-pl/mypath' would still activate if matching 'pl-PL' locale, despite difference in case.

## Testing Details
- [x] Unit Test Adjusted

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Our versioning strategy is as follows:
 ## Unreleased
 
 ### 🐛 Bug Fixes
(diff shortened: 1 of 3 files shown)


yavorsk (approved):
Looks good. However I am thinking, does it make sense to add a test for case sensitive path redirect with and without locale?

art-alexeyenko:
@yavorsk 
>However I am thinking, does it make sense to add a test for case sensitive path redirect with and without locale?

The existing unit test already tests for lowercase locale in redirect pattern, I think we're covered here.

PR #2061, opened 2025-04-01
[sitecore-jss-nextjs]: Link component emptyFieldEditingComponent prop shouldn't pass to NextLink

sc-keithbauer:
- [x] Changelog updated

## Description / Motivation
To support Tailwind in Pages field helpers, we are using passing a custom `emptyFieldEditingComponent` function that adds our component classes to the output: 
```
<span classNames={classNames}>[No text in field]</span>
```

Overall, this approach works, but it does throw an error when used in the Link component from `@sitecore-jss/sitecore-jss-nextjs` which is:

> Warning: React does not recognize the `emptyFieldEditingComponent` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `emptyfieldeditingcomponent` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

This is because the `emptyFieldEditingComponent` is not being removed from the `htmlLinkProps` object before being added to `NextLink` from `next/link`. 

This PR removes the `emptyFieldEditingComponent` props before it is passed to `NextLink`. Also added a unit test to detect that this warning no longer appears when the `emptyFieldEditingComponent` prop is used on a link that results in `NextLink` usage.

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Our versioning strategy is as follows:
   * The `Form` component must be registered in the app to enable non-BYOC Forms support.
   * `SitecoreContext` now supports an `api` property for passing XM Cloud Edge endpoint settings, enabling the `Form` component to access the configured endpoint.
   * Added shared `Form` functionality via the `sitecore-jss/form` submodule.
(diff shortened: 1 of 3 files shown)


addy-pathania (commented):
@sc-keithbauer thankyou for you contribution, this PR looks good.
Just one comment since this is a bugfix, can you please add a new section for bugfix and add your changelog entry there?

addy-pathania (approved):

PR #2081, opened 2025-05-14
[sitecore-jss-react] Introduce a flag to control rendering of Suspense

addy-pathania:
- [x] Changelog updated

## Description / Motivation
Introduced `disableSuspence` flag to the placeholder component  to control rendering of Suspense. 

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Our versioning strategy is as follows:
 
 ### 🐛 Bug Fixes
 
-* `[sitecore-jss-react]` Fix hydration error by introducing a flag `disableSuspense` to control rendering of Suspense. ([#2081](https://github.com/Sitecore/jss/pull/2081))
+* `[sitecore-jss-react]` Add an optional `disableSuspense` flag for Placeholder component that prevents Suspense from being rendered for components. This can help avoid hydration issues in connected mode. ([#2081](https://github.com/Sitecore/jss/pull/2081))
 
 ### 🛠 Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Our versioning strategy is as follows:
 
 ### 🐛 Bug Fixes
 
-* `[sitecore-jss-react]` Fix BYOC form hydration error ([#2081](https://github.com/Sitecore/jss/pull/2081))
+* `[sitecore-jss-react]` Fix hydration error by introducing a flag `disableSuspense` to control rendering of Suspense. ([#2081](https://github.com/Sitecore/jss/pull/2081))
 
 ### 🛠 Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Our versioning strategy is as follows:
 
 ### 🐛 Bug Fixes
 
-* `[sitecore-jss-react]` Add an optional `disableSuspense` flag for Placeholder component that prevents Suspense from being rendered for components. This can help avoid hydration issues in connected mode. ([#2081](https://github.com/Sitecore/jss/pull/2081))
+* `[sitecore-jss-react]` Add an optional `disableSuspense` flag to the Placeholder component to prevent error boundaries from rendering Suspense which helps contain errors for components. This can help avoid hydration issues in connected mode. ([#2081](https://github.com/Sitecore/jss/pull/2081))
 
 ### 🛠 Breaking Changes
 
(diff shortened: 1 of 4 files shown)


packages/sitecore-jss-react/src/components/Placeholder.test.tsx
```diff
@@ -440,7 +439,7 @@ describe('<Placeholder />', () => {
       byocWrapperStub.restore();
     });
 
-    it('should render ErrorBoundary without Suspense for byoc wrapper', () => {
+    xit('should render ErrorBoundary without Suspense for byoc wrapper', () => {
```
art-alexeyenko: this should be enabled right?

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -338,6 +340,29 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
     return transformedComponents;
   }
 
+  hasDynamicComponents(placeholderData: (ComponentRendering | HtmlElementRendering)[]): boolean {
```
art-alexeyenko: our usecase isn't so much about dynamic components, as it is about rendering errorboundary. I.e. byoc components are not always dynamic - only the client-side ones are.
Let's rename this to `shouldSkipSuspense`

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -338,6 +340,29 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
...
+      if (!componentRendering?.componentName) return false;
+
+      const component = this.getComponentForRendering(componentRendering);
+
+      const isDynamicComponent = !!(component as JssComponentType)?.render?.preload;
+
+      const isByocWrapper = componentRendering.componentName === BYOC_WRAPPER_RENDERING_NAME;
+
+      const nestedPlaceholders = componentRendering?.placeholders;
+      if (nestedPlaceholders) {
```
art-alexeyenko: this will disable error boundary for **all** the nested components that eventually contain byoc, both dynamic and not? It would be a regression then.
Does the issue occur still without this part?
As an alternative, could you we wrap `<FEAAS.ExternalComponent>` in `<Suspense>` in `BYOCComponent.tsx`?

art-alexeyenko (changes requested):
I left some comments for consideration, let's refine this approach.
With that said, outside of Components team fixing something inside FEAAS.ExternalComponent and then us adjusting here, it's probably our best shot at a fix.

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
addy-pathania: without this part it will definitely occur since it will then wrap all components with suspense.
as discussed the latter suggested approach will bring us back to the same problem.

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -284,7 +286,7 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
           this.props.modifyComponentProps ? this.props.modifyComponentProps(finalProps) : finalProps
         );
 
-        if (!isEmpty) {
+        if (!isEmpty && !containsDynamicComponents) {
```
yavorsk: isn't it better to use the isDynamicComponent prop below to skip rendering only the Suspense boundary? This way it can keep catching client side errors

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Our versioning strategy is as follows:
 
 ## Unreleased
 
+### 🐛 Bug Fixes
+
+* `[sitecore-jss-react]` Fix hydration error by introducing a flag `disableSuspense` to control rendering of Suspense. ([#2081](https://github.com/Sitecore/jss/pull/2081))
```
art-alexeyenko:
```suggestion
* `[sitecore-jss-react]` Add an optional `disableSuspense` flag for Placeholder component that prevents ErrorBoundary be rendered for components. This can help avoid hydration issues in connected mode, when components using nextjs `dynamic` are included in placeholder.([#2081](https://github.com/Sitecore/jss/pull/2081))
```

art-alexeyenko (approved):
One small suggestion for changelog, and the rest is good. 👍

CHANGELOG.md
addy-pathania: Updated it slightly and I think we should not mention any specific causes of hydration since it can be due to anything.

art-alexeyenko: You must mention error boundary specifically, however, because that's our feature and developers need to fully understand the tradeoff

PR #2093, opened 2025-05-28
[sitecore-jss-react-forms] Upgrade to React 19 and rework unit tests

yavorsk:
- [x] Changelog updated

## Description / Motivation

Upgrade React to version 19 and Nextjs to version 15

- upgrade React and Nextjs dependencies
- with React 19, JSX is in the 'react' namespace and therefore 'react' needs to be imported befoore using JSX. 
- `react-test-renderer` has been deprecated in react 19. additionaly `enzyme` is not supported anymore so all unit tests have been reworked to use [@testing-library/react](https://testing-library.com/docs/react-testing-library/intro/)

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate) - test by creating react forms

## Types of changes


CHANGELOG.md
```diff
@@ -17,7 +17,7 @@ Our versioning strategy is as follows:
 
 ### 🛠 Breaking Changes
 
-* `[Next.js]` `[React]`: `[sitecore-jss]` `[sitecore-jss-react]` `[sitecore-jss-nextjs]` `[create-sitecore-jss]` Upgrade React to version 19 and Nextjs to version 15 ([#2078](https://github.com/Sitecore/jss/pull/2078))([#2084](https://github.com/Sitecore/jss/pull/2084))([#2090](https://github.com/Sitecore/jss/pull/2090)):
+* `[Next.js]` `[React]`: `[sitecore-jss]` `[sitecore-jss-react]` `[sitecore-jss-nextjs]` `[sitecore-jss-react-forms]` `[create-sitecore-jss]` Upgrade React to version 19 and Nextjs to version 15 ([#2078](https://github.com/Sitecore/jss/pull/2078))([#2084](https://github.com/Sitecore/jss/pull/2084))([#2090](https://github.com/Sitecore/jss/pull/2090)) ([#2093](https://github.com/Sitecore/jss/pull/2093)):
```
MenKNas, line 20: More of a question trying to understand the structure? Is this a new way of using tags on `CHANGELOG.md`. Talking about this part:

`[Next.js]` `[React]`: `[sitecore-jss]` `[sitecore-jss-react]` `[sitecore-jss-nextjs]` `[sitecore-jss-react-forms]` `[create-sitecore-jss]`

packages/sitecore-jss-react-forms/src/components/field-templates/button.test.tsx
MenKNas, line 1: Do we need all these fields when we define `p` since we don't really use everything on the test?

packages/sitecore-jss-react-forms/src/components/field-templates/file-upload.tsx
```diff
@@ -116,6 +116,15 @@ class FileUpload extends Component<ValueFieldProps<FileInputViewModel>> {
...
         <FieldValidationErrors {...this.props} />
+        {process.env.TEST && (
+          <script id="file-getEnabledValidation">
+            {JSON.stringify(
+              field.model.validationDataModels.map((validationDM) =>
+                this.getEnabledValidation(validationDM.itemId)
+              )
+            )}
+          </script>
+        )}
```
MenKNas, line 127: Why do we need this?

packages/sitecore-jss-react-forms/src/components/form.tsx
```diff
@@ -119,8 +119,16 @@ export class Form extends Component<FormProps, FormState & FieldStateCollection>
...
+        inert={isInert}
       >
+        {process.env.TEST && (
+          <>
+            <script id="test-form-state">{JSON.stringify(this.state)}</script>
+            <script id="test-collect-fields">
+              {JSON.stringify(this.collectCurrentFieldValues())}
+            </script>
+          </>
+        )}
```
MenKNas, line 131: Why do we need this?

CHANGELOG.md
yavorsk: You mean the addition of the first two tags? it was suggested by Illia to differentiate what was changed for react/next vs angular. And since this is react/next upgrade we make it more visible that there is change for Angular as well

packages/sitecore-jss-react-forms/src/components/form.tsx
yavorsk: testing-library/react has a different concept with regards to testing components compared to enzyme. In Enzyme you can directly access the class component isntance, set and update state, access react components that do not produce html (such as Suspense) and so on. In testing-library you can't do that - the idea there is that test should resemble more the real world usage of the component. So if you want to test if some state value has changed you need to check its effects on the produced html. 
So this is why we render some things only during dunning unit tests.

packages/sitecore-jss-react-forms/src/components/field-templates/file-upload.tsx
yavorsk: see my answer below.

packages/sitecore-jss-react-forms/src/components/field-templates/button.test.tsx
yavorsk: probably not, but i wanted to change as little as possible to the test data

MenKNas (approved):

PR #2112, opened 2025-06-27
[sitecore-jss] Fix forwarding of `Set-Cookie` headers

addy-pathania:
- [x] Changelog updated

## Description / Motivation
Previously, cookies returned by Sitecore (such as `__RequestVerificationToken`) were not being set in the browser when using `NativeDataFetcher`. This caused CSRF validation to fail on form submissions due to missing verification tokens. This fix ensures that any `Set-Cookie` headers returned by the Layout Service,  including required anti-forgery cookie `__RequestVerificationToken` are forwarded from the server response to the browser during SSR.

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/sitecore-jss/src/layout/rest-layout-service.ts
+++ b/packages/sitecore-jss/src/layout/rest-layout-service.ts
@@ -256,10 +256,16 @@ export class RestLayoutService extends LayoutServiceBase {
 
     const headers = serverRes.headers;
 
-    if (headers instanceof Headers) {
-      const setCookieHeader = headers.get('set-cookie');
-      if (setCookieHeader) {
-        res.setHeader('set-cookie', setCookieHeader);
+    if (headers instanceof Headers && headers.has('set-cookie')) {
+      const rawSetCookie = headers.get('set-cookie');
+
+      if (rawSetCookie) {
+        const cookies =
+          rawSetCookie.includes(', ') && rawSetCookie.includes(';')
+            ? rawSetCookie.split(/,(?=\s*\w+=)/)
+            : [rawSetCookie];
+
+        res.setHeader('Set-Cookie', cookies);
       }
     }
 
--- a/packages/sitecore-jss/src/native-fetcher.ts
+++ b/packages/sitecore-jss/src/native-fetcher.ts
@@ -28,7 +28,7 @@ export interface NativeDataFetcherResponse<T> {
   /** Response content */
   data: T;
   /** Response headers */
-  headers?: HeadersInit;
+  headers?: HeadersInit | Headers;
 }
 
 /**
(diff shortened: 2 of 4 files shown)


packages/sitecore-jss/src/native-fetcher.ts
```diff
@@ -103,7 +103,12 @@ export class NativeDataFetcher {
         data: respData,
       });
 
-      return { ...response, data: respData as T };
+      return {
+        data: respData as T,
+        status: response.status,
+        statusText: response.statusText,
+        headers: response.headers,
```
art-alexeyenko, line 110: Why was the original approach with spread not working here? Functionally two formats should be identical (the original would pass more data, potentially)

packages/sitecore-jss/src/layout/rest-layout-service.ts
```diff
@@ -256,10 +256,16 @@ export class RestLayoutService extends LayoutServiceBase {
...
-      if (setCookieHeader) {
-        res.setHeader('set-cookie', setCookieHeader);
+    if (headers instanceof Headers && headers.has('set-cookie')) {
+      const rawSetCookie = headers.get('set-cookie');
+
+      if (rawSetCookie) {
+        const cookies =
+          rawSetCookie.includes(', ') && rawSetCookie.includes(';')
+            ? rawSetCookie.split(/,(?=\s*\w+=)/)
+            : [rawSetCookie];
```
art-alexeyenko, line 266: This may cause `cookies` to be a double array ([[content]] instead of []) if set-cookies is already parsed with Array type.
You can conditionally parse the string value, similar to the way we do it in other places:
https://github.com/Sitecore/jss/blob/1dfc9ccd1e3fa6e3249ef82dd05136f544734bb0/packages/sitecore-jss-nextjs/src/editing/editing-render-middleware.ts#L369

art-alexeyenko (approved):
Approving in advance, but please check out the comments

packages/sitecore-jss/src/layout/rest-layout-service.ts
art-alexeyenko: Please also add a unit test to ensure `set-cookies` is handled correctly when it's parsed as array

addy-pathania: Nextjs `res.getHeader('Set-Cookie');` returns a string[] for multiple Set-Cookie values.

The Fetch API `res.headers.get('set-cookie');`  doesn’t expose multiple Set-Cookie headers as an array rather it combines them into a single string. Check the test I have added.

This logic attempts to handle multiple Set-Cookie values incorrectly combined in a string, so it will not create a double array.

packages/sitecore-jss/src/native-fetcher.ts
addy-pathania: The original response spreads the native Response object into a plain object, but native fetch Response's headers are stored using a symbol ([Symbol(headers)]) and not copied by ...spread.
So when you try to access `serverRes.headers` later in `setupResHeaders` function in `RestLayoutService` service, it’s undefined.

Also this is the data which we need, as it matches the return type.

art-alexeyenko: I modified my comment since then, sorry

PR #2120, opened 2025-08-05
[sitecore-jss-nextjs] Add support for displayName based routing with special character encoding

addy-pathania:
- [x] Changelog updated

## Description / Motivation
Introduced support for using `displayName` values in route generation for statically built pages in JSS Next.js apps. This enables localized and user-friendly URLs when SXA link provider is configured to use display names.

### Changes
- Extended GraphQL query to include `route.displayName`
- Updated sitemap transformation logic to generate additional paths based on `displayName`
- Ensured `displayName` paths are UTF-8 encoded using `encodeURIComponent`
- `enableDisplayNameRouting` (optional, boolean): When set to true, the service will generate static paths using the displayName of Sitecore items in addition to their itemName. Default is false.

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,6 @@ Our versioning strategy is as follows:
 
 * `[sitecore-jss-nextjs]` Introduced support for using `displayName` values in route generation for statically built pages in JSS Next.js apps. This enables localized and user-friendly URLs when SXA link provider is configured to use display names.  Ensured `displayName` paths are UTF-8 encoded. ([#2120](https://github.com/Sitecore/jss/pull/2120))
 
-
 ## 22.8.0
 
 ### 🎉 New Features & Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,11 @@ Our versioning strategy is as follows:
 
 * `[sitecore-jss-nextjs]` Ensure Redirect Middleware handles case-insensitive path matching to prevent missed redirects due to casing differences ([#2114](https://github.com/Sitecore/jss/pull/2114))
 
+### 🎉 New Features & Improvements
+
+* `[sitecore-jss-nextjs]` Introduced support for using `displayName` values in route generation for statically built pages in JSS Next.js apps. This enables localized and user-friendly URLs when SXA link provider is configured to use display names.  Ensured `displayName` paths are UTF-8 encoded. ([#2120](https://github.com/Sitecore/jss/pull/2120))
+
+
 ## 22.8.0
 
 ### 🎉 New Features & Improvements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ Our versioning strategy is as follows:
 
 ### 🎉 New Features & Improvements
 
-* `[sitecore-jss-nextjs]` Introduced support for using `displayName` values in route generation for statically built pages in JSS Next.js apps. This enables localized and user-friendly URLs when SXA link provider is configured to use display names.  Ensured `displayName` paths are UTF-8 encoded. ([#2120](https://github.com/Sitecore/jss/pull/2120))
+* `[sitecore-jss-nextjs]` Introduced support for using `displayName` values in route generation for statically built pages in JSS Next.js apps. This enables localized and user-friendly URLs when SXA link provider is configured to use display names.  Ensured `displayName` paths are UTF-8 encoded. This feature can be enabled by setting the `enableDisplayNameRouting` flag to true. By default, it is set to false. ([#2120](https://github.com/Sitecore/jss/pull/2120))
 
 ## 22.8.0
 
--- a/packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
+++ b/packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
@@ -256,18 +256,32 @@ export abstract class BaseGraphQLSitemapService {
   ): Promise<StaticPath[]> {
     const aggregatedPaths: StaticPath[] = [];
 
-    // Step 1: Create a map of item segment -> displayName
+    // Build a map of the last segment of each path to its encoded display name.
+    // This is used later to substitute the final segment with a display name if available.
     const displayNameMap = new Map<string, string>();
     for (const item of sitePaths) {
+      if (!item || typeof item.path !== 'string') continue;
+
       const segments = item.path.replace(/^\/|\/$/g, '').split('/');
       const lastSegment = segments[segments.length - 1];
       const displayName = item.route?.displayName;
+
       if (displayName) {
         displayNameMap.set(lastSegment, encodeURIComponent(displayName));
       }
     }
 
-    // Step 2: Generate combinations recursively
+    /**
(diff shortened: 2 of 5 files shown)


packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
```diff
@@ -258,14 +262,27 @@ export abstract class BaseGraphQLSitemapService {
...
+      }
 
+      // Personalization variants
       const variantIds = item.route?.personalization?.variantIds?.filter(
         (variantId) => !variantId.includes('_') // exclude component A/B test variants
       );
       if (variantIds?.length) {
         aggregatedPaths.push(
-          ...variantIds.map((varId) => formatPath(getPersonalizedRewrite(item.path, [varId])))
+          ...variantIds.map((varId) => formatPath(getPersonalizedRewrite(itemNamePath, [varId])))
```
art-alexeyenko: display name paths should also be personalize-able

packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
```diff
@@ -258,14 +262,27 @@ export abstract class BaseGraphQLSitemapService {
...
       if (!item) return;
 
-      aggregatedPaths.push(formatPath(item.path));
+      // ItemName-based path
+      const itemNamePath = item.path;
+      aggregatedPaths.push(formatPath(itemNamePath));
+
+      // DisplayName-based path
+      const displayName = item.route?.displayName;
+      if (typeof displayName === 'string' && displayName.trim().length > 0) {
```
art-alexeyenko: This will only cover the paths with last part being displayName, right?
I.e. consider this item structure
-Home
  --Page1 (displayName: Orders)
    ---Page2 (displayName: Details)
    
This logic will only make aggregated path for /Page1/Details, right?

CHANGELOG.md
```diff
@@ -15,6 +15,10 @@ Our versioning strategy is as follows:
 
 * `[sitecore-jss-nextjs]` Ensure Redirect Middleware handles case-insensitive path matching to prevent missed redirects due to casing differences ([#2114](https://github.com/Sitecore/jss/pull/2114))
 
+### 🎉 New Features & Improvements
+
+* `[sitecore-jss-nextjs]` Introduced support for using `displayName` values in route generation for statically built pages in JSS Next.js apps. This enables localized and user-friendly URLs when SXA link provider is configured to use display names.  Ensured `displayName` paths are UTF-8 encoded. ([#2120](https://github.com/Sitecore/jss/pull/2120))
```
art-alexeyenko: What about display name route resolution for SSR?

art-alexeyenko (changes requested):

packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
addy-pathania: During implementation, I found that Sitecore Personalize resolves variants only when the request path matches the original itemName-based route. When a personalized page is accessed via its display name, the page renders correctly, but personalization is not applied and the default content is shown instead. 

Given this limitation, I've chosen not to generate personalized static paths for display name–based routes in our sitemap. We should consider documenting this limitation and push this further on platform to support personalization resolution via display name paths.

packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
addy-pathania: Our implementation covers all nested routes, just like item name based routes.

addy-pathania: I tested this, works fine.
Next.js will generate that page statically, and as long as
 - The route exists in the sitemap (getStaticPaths)
 - The path matches during navigation
Then it will render successfully.

CHANGELOG.md
addy-pathania: SSR doesn’t care whether it’s `itemName` or `displayName`, as long as the route exists in Sitecore and matches the URL, it works. Display name routes were already working for SSR.

packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
art-alexeyenko: It works fine for routing, but paths with all display names are not generated. From my latest test, here are the pages generated
```
├   ├ /en/_site_base/About/Page-3 (968 ms)
├   ├ /en/_site_base/About/DisplayPage (958 ms)
├   ├ /en/_site_base/DisplayAbout (958 ms)
├   ├ /en/_site_base (958 ms)
├   ├ /en/_site_base/About (957 ms)
```
But there also should be the page for `/en/_site_base/DisplayAbout/DisplayPage`

CHANGELOG.md
art-alexeyenko: I wrongly assumed the display name routing should also be implemented, thanks for verifying it.

packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
art-alexeyenko: Still this needs to be addressed. At the very least we need to ensure personalization is not failing due to logic we have ownership of.

addy-pathania: I have refactored the logic and from my last test the personalized display name based page worked fine as well.
Maybe previously i was testing stale data.

packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
addy-pathania: Refactored and all combinations gets pre generated now.

packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
```diff
@@ -250,62 +254,99 @@ export abstract class BaseGraphQLSitemapService {
...
+     * - The item name segment (default)
+     * - Or the display name (if available in the map)
+     *
+     * For example: if path is ['about', 'team'] and displayName for 'team' is 'Team-Page',
+     * it will generate:
+     * - ['about', 'team']
+     * - ['about', 'Team-Page']
+     * @param segments
+     */
+    const generateCombinations = (segments: string[]): string[][] => {
```
art-alexeyenko, line 295: One final suggestion I'd add is to make displayName SSG optional. It adds a lot of overhead for static paths and not everyone will like it.
It is up to your consideration however.

art-alexeyenko (approved):
Approved - with one optional suggestion which is up to you.

packages/sitecore-jss-nextjs/src/services/base-graphql-sitemap-service.ts
addy-pathania: Yes i was thinking of adding this myself. Build will degrade if there are lot of routes.

Added.

PR #2131, opened 2025-08-18
[core] Update package.json to add form-data dependency version 2.5.4

faiqamin:
- [x] Changelog updated

## Description / Motivation

- the `critical` warning stems from the `form-data` package for versions <2.5.4, which is a transitive dependency of the `request` package.
- the `request` package itself is deprecated, so we can't update its version
- `@2` in the resolution is a selector: it applies only when a package asks for form-data in the 2.x range.
- `request` declares "form-data": "~2.3.2", which is 2.x, so Yarn resolves that to 2.5.4 due to the resolution.
- Anything that asks for form-data 4.x won’t match `@2`, so it keeps using 4.x. The code in request does a normal require('form-data'); the exact version is chosen by the package manager based on package.json ranges and the resolution.

## Testing Details

- [ ] Unit Test Added

## Types of changes


art-alexeyenko:
@faiqamin this will only have effect within monorepo. If you check an app newly scaffolded with `npx create-sitecore-jss`, and run `npm audit --json` the vulnerabilities would still be present.

packages/create-sitecore-jss/src/templates/angular/package.json
```diff
@@ -109,8 +107,13 @@
...
     "move-cli": "^2.0.0",
     "npm-run-all2": "^7.0.1",
-    "protractor": "^7.0.0",
+    
     "ts-node": "~10.9.2",
     "typescript": "~5.4.0"
+  },
+  "overrides": {
+    "request": {
+      "form-data": "2.5.4"
```
art-alexeyenko: Please use `^2.5.5` to avoid this warning
image

art-alexeyenko (approved):

PR #2159, opened 2025-11-12
[sitecore-jss-nextjs] Add regex variable substitution for absolute and external URL redirects

addy-pathania:
- [x] Changelog updated

## Description / Motivation
Fixes regex-based redirects with variable extraction when the target URL is external or absolute.

Resolves issue where redirects like:
- Source: `^/redirect/docs/([^/]+)/([^/]+)$`
- Target: `https://museum.olympics.com/docs/$1/$2$`
Would redirect to `https://museum.olympics.com/docs/$1/$2` instead of the expected URL with substituted values.

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,9 @@ Our versioning strategy is as follows:
 
 ### 🐛 Bug Fixes
 
-* `[Next.js]` [Redirects] Preserve default locale in external absolute urls ([#2142](https://github.com/Sitecore/jss/pull/2142))
+* `[sitecore-jss-nextjs]` Preserve default locale in external absolute urls ([#2142](https://github.com/Sitecore/jss/pull/2142))
 * `[React]` Custom properties are not applied to empty field in editing metadata mode ([#2141](https://github.com/Sitecore/jss/pull/2141))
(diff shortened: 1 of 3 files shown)


MenKNas:
@addy-pathania  

Everything looks good to me, so this is approved.

Just a small footnote which isn't specific to your PR, but since we're in this context, could you verify one edge case?

When a request has a mixed-case pathname with no query string (for example:  `/Redirect/Docs/ABC/123`), `normalizeUrl` doesn't lowercase the pathname as far as I can tell. The pattern should still match (due to the `/i` flag), but could you please confirm that variable substitution works correctly in this scenario?

MenKNas (approved):

addy-pathania:
> @addy-pathania
> Everything looks good to me, so this is approved.
> Just a small footnote which isn't specific to your PR, but since we're in this context, could you verify one edge case?
> When a request has a mixed-case pathname with no query string (for example: `/Redirect/Docs/ABC/123`), `normalizeUrl` doesn't lowercase the pathname as far as I can tell. The pattern should still match (due to the `/i` flag), but could you please confirm that variable substitution works correctly in this scenario?

@MenKNas I tested the mentioned edge case and i confirm the middleware handles it correctly since we use the pattern matching `/i` flag

PR #2177, opened 2026-01-12
Upgrade to Next 16

MenKNas:
- [x] Changelog updated

## Description / Motivation
With this PR, we are introducing the Upgrade JSS to Next.js 16. Updated package versions, replaced `middleware.ts` with `proxy.ts`, removed `NextFetchEvent` from middleware signatures, fixed `NextURL` type extraction, defined `STATIC_PROPS_ID` and `SERVER_PROPS_ID` locally, and added `--webpack` to build scripts to support existing webpack configurations.

## Testing Details

- [ ] Unit Test Added
- [x] Manual Test/Other -> Scaffolded samples, apps, and made sure everything is building and running properly.

## Types of changes


packages/sitecore-jss-vue/src/components/__snapshots__/HiddenRendering.test.ts.snap
```diff
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
```
art-alexeyenko, line 1: let's exclude jss-vue files from this PR

art-alexeyenko (approved):
Looks good - one thing I'd suggest is to adjust FEAASScripts the same way you did for Content SDK, but it's up to you, it won't break much either way

packages/sitecore-jss-vue/src/components/__snapshots__/HiddenRendering.test.ts.snap
MenKNas: This is a mandatory change for the Vue package tests to pass since Jest has changed the way it handles the URLS on the snapshots. I don't know why this was passing/working earlier, to be honest, but now the tests were failing. If you wanna dig deeper, you can have a look at the link below: 

https://jestjs.io/docs/upgrading-to-jest30#update-broken-documentation-link

PR #2186, opened 2026-03-09
Migrate ESLint to v9 and flat config

MenKNas:
- [x] Changelog updated

## Description / Motivation
This PR introduces ESLint v9 and flat config syntax. All scaffolded apps (Angular, React, Next.js, Vue) and related samples now use the new flat config format (`eslint.config.js` / `eslint.config.mjs`) instead of legacy `.eslintrc` and `.eslintignore`. 

For this purpose, a new `eslint-configs/flat/` folder provides a shared flat config for the whole repo: `index.mjs` exports a `createConfig(files, options)` factory that composes base rules, TypeScript, optional React, test overrides, and ignores. Packages (and the root) use it by importing `createConfig` from `eslint-configs/flat/index. mjs` in their `eslint.config.mjs`, so lint is consistent across the repo without duplicating config.

## Testing Details

- [ ] Unit Test Added

## Types of changes


packages/create-sitecore-jss/src/templates/angular/eslint.config.js
```diff
@@ -0,0 +1,63 @@
...
+
+module.exports = tseslint.config(
+  {
+    ignores: [
+      'projects/**',
+      'node_modules',
+      'dist',
+      'src/environments/environment.js',
+      'src/environments/environment.prod.js',
+      'src/app/components/app-components.module.ts',
```
art-alexeyenko, line 20: js files are understandable, but why exclude this file?

packages/create-sitecore-jss/src/templates/react/scripts/http-renderer.js
```diff
@@ -9,12 +9,12 @@ const express = require('express');
 /* eslint-disable no-console */
 
 startRenderHostTunnel('localhost', { port: 5000 })
-  .then((tunnelUrl) => {
+  .then((_tunnelUrl) => {
```
art-alexeyenko: if _tunnelUrl and other underscored variables are not used, perhaps they can just be removed to avoid confusion?

packages/create-sitecore-jss/src/templates/vue/package.json
```diff
@@ -35,7 +35,7 @@
     "start:vue": "vue-cli-service serve --open --host localhost",
     "start:proxy": "node scripts/disconnected-mode-proxy.js",
     "start:watch-components": "node scripts/generate-component-factory.js --watch",
-    "build:client": "cross-env-shell BUILD_TARGET_ENV=client PUBLIC_URL=$npm_package_config_sitecoreDistPath \"vue-cli-service build\"",
```
art-alexeyenko, line 38: seems like `vue` template was not updated, please re-verify if it needs to be

packages/sitecore-jss-dev-tools/src/package-deploy.ts
```diff
@@ -209,10 +209,10 @@ export async function watchJobStatus(options: PackageDeployOptions, taskName: st
 
   req.on('response', (res) => {
     if (res.statusCode !== 200) {
-      let errorData = '';
+      let _errorData = '';
 
       res.on('data', (chunk) => {
-        errorData += chunk;
+        _errorData += chunk;
```
art-alexeyenko: It's better to log error data here instead of leaving it as is. It's a bit out of the task scope, but shouldn't be a large change

package.json
```diff
@@ -3,7 +3,7 @@
   "scripts": {
     "reset": "lerna clean --yes && yarn dlx rimraf node_modules && yarn install && yarn build",
     "build": "lerna run build",
-    "lint-packages": "lerna run lint",
+    "lint-packages": "lerna run lint --scope='@sitecore-jss/*' --scope='create-sitecore-jss' --ignore='@sitecore-jss/sitecore-jss-angular'",
```
art-alexeyenko: Angular package should not be ignored, we should still lint it alongside others

tsconfig.scripts.json
```diff
@@ -0,0 +1,11 @@
+{
```
art-alexeyenko: I don't think this file is needed - we don't use ts-node at root level

packages/sitecore-jss-vue/src/components/Placeholder.test.ts
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable vue/one-component-per-file */
 /* eslint-disable @typescript-eslint/no-empty-function */
 /* eslint-disable no-unused-expressions */
```
art-alexeyenko: could you check if eslint-disable directives can be removed here and across the monorepo? Some directives were already removed in other packages - which is great

art-alexeyenko (changes requested):
Great work and a big effort - thank you.
Leaving some comments to help bring it to completion

packages/create-sitecore-jss/src/templates/angular/eslint.config.js
MenKNas: `app-components.module.ts` is auto-generated by scripts/generate-component-factory when components are added/removed (and is in .gitignore). We ignore it in ESLint so we don’t lint generated code; I’ve made that explicit in the config comment.

packages/create-sitecore-jss/src/templates/vue/package.json
MenKNas: Vue template is updated now.

packages/sitecore-jss-angular/eslint.config.mjs
```diff
@@ -0,0 +1,108 @@
...
+        {
+          format: ['PascalCase'],
+          selector: 'typeLike',
+          custom: { regex: '^I[A-Z]', match: false },
+        },
+      ],
+      '@typescript-eslint/member-ordering': 'error',
+      '@typescript-eslint/no-use-before-define': ['error', { functions: false, variables: false }],
+      '@typescript-eslint/typedef': 'error',
+      '@typescript-eslint/no-require-imports': 'off',
```
art-alexeyenko: I think these rules should not be off as it breaks away from repo-wide conventions. Were there any issues with them?

art-alexeyenko (approved):

PR #2190, opened 2026-03-30
[Personalize] Support dynamic token replacement from decision tables

bransbury:
- [x] Changelog updated

## Description / Motivation

This PR adds support for dynamic token replacement from Sitecore Personalize decision table responses across the JSS personalization pipeline.

### Problem

When Sitecore Personalize decision tables return tokens (e.g. `{ firstName: "Bob", city: "Portland" }`), there is currently no mechanism in JSS to apply these values to layout data. Content authors can configure `{{firstName}}` placeholders in their content, but the values are never substituted at render time.

### Solution

**New `token-replace` utility** (`sitecore-jss/src/personalize/token-replace.ts`):
- `replaceTokens()`: substitutes `{{key}}` and `{{key|fallback}}` mustache-style tokens in strings
- `replaceTokensInObject()`: recursively walks layout data and applies token replacement to all string values
- Supports fallback values (`{{name|Valued Customer}}`), whitespace-trimmed keys, `removeUnmatched` option, and `onUnmatched` callback
- Includes prototype pollution guards (`__proto__`, `constructor`, `prototype` keys are skipped)
- ReDoS-safe regex design (no polynomial backtracking)

**Next.js middleware** (`personalize-middleware.ts`):
- Collects tokens from each personalize execution's decision table response
- Encodes tokens as base64 JSON in the `x-sc-personalize-tokens` response header
- Tokens are keyed by variant friendly ID for multi-execution scenarios

**Next.js page-props plugin** (`personalize.ts` template):
- Reads the `x-sc-personalize-tokens` header from the incoming request
- Merges per-variant token maps with collision detection/logging
- Applies `replaceTokensInObject()` to layout data after `personalizeLayout()` runs
- Validates all token values are strings; guards against prototype pollution

**Proxy PersonalizeHelper** (`sitecore-jss-proxy`):
- `getVariantIds()` now returns `{ variantIds, tokens }` to pass tokens through
- Applies token replacement in `personalizeLayoutData()` after layout personalization

### 🛠 Breaking Changes

`PersonalizeHelper.getVariantIds()` in `sitecore-jss-proxy` changed its return type from `Promise<string[]>` to `Promise<{ variantIds: string[]; tokens: Record<string, string> }>`. This is a `protected` method — only subclasses that override it need to update their signature. This is unlikely to affect most consumers since the proxy package is used by non-Next.js rendering hosts and this method is not part of the public API surface.

## Testing Details

- [x] Unit Test Added

### New test coverage:
- **`token-replace.test.ts`**: 25+ test cases covering: single/multiple token replacement, fallback values, whitespace trimming, `removeUnmatched`, `onUnmatched` callback, nested object traversal, array handling, null/undefined safety, prototype pollution guard, ReDoS guard, HTML/URL content, special regex characters in values
- **`personalize-middleware.test.ts`**: updated tests for token forwarding via header, multi-execution token aggregation, base64 encoding
- **`PersonalizeHelper.test.ts`**: updated tests for token application to layout data, empty tokens handling, collision detection

## Types of changes

## Packages affected

| Package | Changes |
|---|---|
| `sitecore-jss` | New `token-replace.ts` utility, exported from `personalize` subpath |
| `sitecore-jss-nextjs` | Middleware collects & forwards tokens; re-exports `PERSONALIZE_TOKENS_HEADER` |
| `sitecore-jss-proxy` | `PersonalizeHelper` applies tokens to layout data (⚠️ breaking: `getVariantIds` return type) |
| `create-sitecore-jss` | Updated XM Cloud template `personalize.ts` plugin to consume tokens |

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "editor.fontSize": 13,
-}
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ packages/*/*.tgz
 samples/
 
 .idea
+.vscode/settings.json
 .next
 
 .npmrc
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Changelog
+
 All notable changes to this project will be documented in this file. The format (starting with 18.0.0) is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 This project does NOT strictly adhere to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). Major versions of this project may include breaking changes in core packages but also denote compatibility with Sitecore Platform versions. Refer to the "Headless Services" section in the Sitecore modules compatibility table ([Sitecore XP 7.5 - 9.3](https://support.sitecore.com/kb?id=kb_article_view&sysparm_article=KB0541788), [Sitecore XP 10.0 and later](https://support.sitecore.com/kb?id=kb_article_view&sysparm_article=KB1000576)) or the [Headless Rendering download page](https://dev.sitecore.net/Downloads/Sitecore_Headless_Rendering.aspx) for details. Minor versions may also include breaking changes in framework packages.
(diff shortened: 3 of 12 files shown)


.vscode/settings.json
bransbury, line 1: Looks like this was committed erroneously in the past so I've removed and added to .gitignore. This is unrelated to the main changes

PR #2204, opened 2026-07-04
[sitecore-jss-react] RichText: stabilize dangerouslySetInnerHTML reference across re-renders

### LINKED ISSUE #2203: RichText builds a new dangerouslySetInnerHTML object every render, causing full DOM replacement and dropped internal-link listeners
### Packages / versions
- `@sitecore-jss/sitecore-jss-react`: 22.11.0
- `@sitecore-jss/sitecore-jss-nextjs`: 22.11.0
- React: 19.2.7 / React DOM: 19.2.7
- Next.js: 15.5.20
- Node: 22.23.1

### What's happening

`RichText` in `sitecore-jss-react` builds its `dangerouslySetInnerHTML` prop inline on every render:

https://github.com/Sitecore/jss/blob/v22.11.0/packages/sitecore-jss-react/src/components/RichText.tsx#L36-L44

```js
const htmlProps = {
  dangerouslySetInnerHTML: {
    __html: field.editable && editable ? field.editable : field.value,
  },
  ref,
  ...otherProps,
};

return React.createElement(tag || 'div', htmlProps);
```

This creates a brand-new object reference every time the component function runs, even when `field.value` is unchanged.

React DOM's host-component prop diffing for plain elements (`div`, `span`, `a`, `p`, etc. — i.e. anything that falls through to the generic diff path rather than a specialized one like `<input>`/`<textarea>`) compares the entire `dangerouslySetInnerHTML` prop **by reference**, not by the nested `.__html` string:

```js
// react-dom-client.development.js, updateProperties()
for (lastProp in nextProps)
  (_propKey8 = nextProps[lastProp]),
    (propKey = lastProps[lastProp]),
    !nextProps.hasOwnProperty(lastProp) ||
      _propKey8 === propKey ||               // <- reference equality on the whole {__html} object
      (null == _propKey8 && null == propKey) ||
      setProp(domElement, tag, lastProp, _propKey8, nextProps, propKey);
```

```js
// setProp(), case "dangerouslySetInnerHTML"
domElement.innerHTML = value.__html;         // unconditional re-set once reference mismatches
```

Because `RichText` never stabilizes the object reference, **every parent re-render forces a full `innerHTML` replacement**, even when the rendered content hasn't changed. This destroys and recreates every child DOM node in the rich text.

That's a real problem in combination with `sitecore-jss-nextjs`'s `NextRichText`, which attaches internal-link click handlers imperatively, once, via a `useEffect` keyed only on `hasText`:

https://github.com/Sitecore/jss/blob/v22.11.0/packages/sitecore-jss-nextjs/src/components/RichText.tsx#L41-L44

```js
useEffect(() => {
  if (hasText && !isEditing) {
    initializeLinks(); // L58: node.querySelectorAll + L91: link.addEventListener('click', routeHandler)
  }
}, [hasText]);
```

Since `hasText` doesn't change across re-renders, this effect never reruns — so when an ancestor re-renders and `innerHTML` gets blown away and rebuilt (per the above), the freshly-created `<a>` nodes never get their `click`/prefetch listeners reattached. In practice this means:

- Internal links inside rich text silently lose their SPA routing/prefetch behavior after any unrelated parent re-render.
- If a re-render happens while a user is mid-interaction with a link (e.g. `mousedown` already fired), the DOM node is replaced out from under the pointer, breaking the click.
- It's also a needless perf cost: a full `innerHTML` rewrite on every incidental re-render, regardless of whether content changed.

### To Reproduce

Render `<RichText field={{ value: '<a href="/foo">bar</a>' }} />` inside a parent with some unrelated state that ticks (e.g. `setInterval` triggering a re-render). Watch the DOM node for the anchor via `MutationObserver`, or attach a marker/listener to it after mount — it gets replaced on every parent re-render even though `field.value` never changes.

### Expected Behavior

Memoize the `dangerouslySetInnerHTML` object in `RichText` (e.g. `useMemo` keyed on the actual html string / `field.editable` / `field.value`) so its reference stays stable across renders when content hasn't changed. That alone would prevent the unnecessary `innerHTML` churn and, transitively, the listener loss in `NextRichText`.

### Possible Fix

We're currently wrapping `RichText` in our app with a `React.memo` that skips re-rendering unless `field.value`, `field.editable`, `tag`, `className`, `editable`, or `onClick` actually change — but a fix upstream (stabilizing the `dangerouslySetInnerHTML` reference) would remove the need for consumers to do this themselves.

### Provide environment information

* Sitecore Version: n/a
* JSS Version: 22.11.0
* Browser Name and version: n/a
* Operating System and version (desktop or mobile): n/a
* Link to your project (if available): n/a

pzi:
- [x] Changelog updated

## Description / Motivation

Fixes [#2203](https://github.com/Sitecore/jss/issues/2203).

`RichText` in `sitecore-jss-react` built its `dangerouslySetInnerHTML` prop inline on every render, creating a new object reference each time the component function ran — even when the rendered html was unchanged.

React DOM compares the `dangerouslySetInnerHTML` prop **by reference** (not by the nested `__html` string) for plain host elements, and unconditionally re-sets `domElement.innerHTML` when the reference changes. As a result, every parent re-render forced a full `innerHTML` replacement, destroying and recreating every child DOM node inside the rich text.

This is especially problematic in combination with `sitecore-jss-nextjs`'s `NextRichText`, which attaches internal-link `click`/prefetch listeners imperatively via a `useEffect` keyed only on `hasText`. Since `hasText` doesn't change across re-renders, the effect never reruns — so after any unrelated ancestor re-render, the freshly recreated `<a>` nodes silently lose their SPA routing/prefetch behavior.

Symptoms:

- Internal links inside rich text lose SPA routing/prefetch after any unrelated parent re-render (full page load instead of client-side navigation).
- DOM nodes replaced mid-interaction (e.g. between `mousedown` and `click`) break clicks.
- Needless perf cost: full `innerHTML` rewrite on every incidental re-render.

### Fix

Memoize the `dangerouslySetInnerHTML` object in `RichText` with `useMemo`, keyed on the resolved html string (`field.editable`/`field.value` depending on `editable`). The object reference now stays stable across re-renders when content is unchanged — including when the `field` object itself is a new instance with the same value — so React DOM skips the `innerHTML` re-set and existing DOM nodes (and their listeners) are preserved. When the html actually changes, behavior is unchanged: content is re-rendered as before.

The memoization is placed before the early `null` return to comply with the Rules of Hooks.

### Test-infrastructure change (`sitecore-jss-nextjs/test/setup.js`)

This is the first hook used by a `sitecore-jss-react` component that `sitecore-jss-nextjs` unit tests render. Because the monorepo uses `nmHoistingLimits: workspaces`, each workspace resolves its **own physical copy** of `react` — so the workspace-linked `sitecore-jss-react` dist ran its hooks against a different React instance than the `react-dom` rendering the tests, throwing `TypeError: Cannot read properties of null (reading 'useMemo')`.

The mocha setup now redirects `react`/`react-dom` resolution to a single copy for the test run. This affects unit tests only; consuming apps have a single React via peer dependencies and are unaffected.

### Changes

- `packages/sitecore-jss-react/src/components/RichText.tsx` — memoize `dangerouslySetInnerHTML` keyed on the resolved html string.
- `packages/sitecore-jss-react/src/components/RichText.test.tsx` — new tests: DOM content is not replaced on re-render with an unchanged value (same or new `field` object), and *is* replaced when the value changes.
- `packages/sitecore-jss-nextjs/test/setup.js` — dedupe `react`/`react-dom` module resolution in the mocha test environment (see above).

## Testing Details

- New test reproducing the bug fails without the fix (`AssertionError: expected <a href="/foo"></a> to equal <a href="/foo"></a>` — anchor node replaced on re-render) and passes with it. A second test verifies the DOM *is* replaced when the field value actually changes.
- `packages/sitecore-jss-react`: `yarn test` — 288 passing, `yarn lint` clean, `yarn build` clean.
- `packages/sitecore-jss-nextjs`: `yarn test` — 324 passing, `yarn lint` clean.

- [x] Unit Test Added

## Types of changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)


illiakovalenko (approved):
@pzi Thank you for your contribution! We really appreciate it. We’ll include this fix in the JSS 23.0.0 release

PR #2218, opened 2026-08-07
[sitecore-jss-react] Hydration failed because the server rendered HTML didn't match the client

sc-naveenhedallaarachchi:
- [x] Changelog updated

## Description / Motivation

- Fixes Experience Editor hydration mismatch warnings caused by Sitecore chrome mutating the DOM after SSR (e.g. `sc-part-of`, `scEnabledChrome`, editable field chrome).
- Adds `suppressHydrationWarning` only in SDK components so apps do not need it on every rendering.
## Changes
- `ErrorBoundary`: in edit mode, wrap renderings with a `display: contents` node that has `suppressHydrationWarning` (central fix for rendering-root EE tags).
- `Placeholder` / `PlaceholderCommon`: suppress on empty-placeholder wrapper and EE raw chrome elements (`phkey` / `<code>`).
- Field components (`RichText`, `Text`, `Image`, `Link`): suppress on editable chrome markup paths only.
- `HiddenRendering`: suppress on root (not wrapped by ErrorBoundary).
- Tests updated for the ErrorBoundary edit-mode wrapper.

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
+++ b/packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
@@ -54,6 +54,13 @@ class ErrorBoundary extends React.Component<ErrorBoundaryProps> {
     );
   }
 
+  isPageEditing(): boolean {
+    return (
+      !!this.props.sitecoreContext?.pageEditing ||
+      this.props.sitecoreContext?.pageState === LayoutServicePageState.Edit
+    );
+  }
+
   render() {
     if (this.state.error) {
       if (this.props.errorComponent) {
@@ -81,17 +88,32 @@ class ErrorBoundary extends React.Component<ErrorBoundaryProps> {
     }
 
     // do not apply suspense when suspense is disabled or when on already dynamic components
+    let content: ReactNode;
     if ((this.props.disableSuspense ?? true) || this.props.isDynamic) {
-      return this.props.children;
+      content = this.props.children;
+    } else {
+      content = (
+        <Suspense
+          fallback={<h4>{this.props.componentLoadingMessage || this.defaultLoadingMessage}</h4>}
+        >
+          {this.props.children}
+        </Suspense>
+      );
     }
 
-    return (
-      <Suspense
-        fallback={<h4>{this.props.componentLoadingMessage || this.defaultLoadingMessage}</h4>}
-      >
-        {this.props.children}
-      </Suspense>
-    );
+    // Experience Editor tags the outermost DOM node of each rendering (sc-part-of,
+    // scEnabledChrome) after SSR. Own that node in the SDK during editing so hydration
+    // mismatches can be suppressed without requiring suppressHydrationWarning on every
+    // app component. `display: contents` keeps SXA/Bootstrap column layout intact.
+    if (this.isPageEditing()) {
+      return (
+        <div suppressHydrationWarning style={{ display: 'contents' }}>
+          {content}
+        </div>
+      );
+    }
+
+    return content;
   }
 }
 
--- a/packages/sitecore-jss-react/src/components/Link.tsx
+++ b/packages/sitecore-jss-react/src/components/Link.tsx
@@ -56,6 +56,8 @@ export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
 
       const htmlProps = {
         className: 'sc-link-wrapper',
+        // Experience Editor rewrites this chrome markup client-side after SSR.
+        suppressHydrationWarning: true,
         dangerouslySetInnerHTML: {
           __html: markup,
         },
--- a/packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
+++ b/packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
@@ -339,6 +339,10 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
       props.ref = this.addRef; // only need ref for placeholder containers, trying to add it to other components (e.g. stateless components) may result in a warning.
     }
 
+    // Editing middleware renames phkey→key in the HTML EE loads; EE may also mutate chrome
+    // attributes before React hydrates. Suppress the expected mismatch on these SDK nodes.
+    props.suppressHydrationWarning = true;
+
     return React.createElement(elem.name, props);
   }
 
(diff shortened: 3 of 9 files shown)


packages/sitecore-jss-react/src/components/Link.tsx
```diff
@@ -56,6 +56,8 @@ export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
 
       const htmlProps = {
         className: 'sc-link-wrapper',
+        // Experience Editor rewrites this chrome markup client-side after SSR.
+        suppressHydrationWarning: true,
```
illiakovalenko, line 60: Do we need to add this property? do we have a case when we have a hydration warning related to Link component?

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -339,6 +339,10 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
       props.ref = this.addRef; // only need ref for placeholder containers, trying to add it to other components (e.g. stateless components) may result in a warning.
     }
 
+    // Editing middleware renames phkey→key in the HTML EE loads; EE may also mutate chrome
```
illiakovalenko: Editing middleware renaming should not affect it and should not produce hydration mismatch, unless I miss something
```suggestion
    // EE may mutate chrome
```

packages/sitecore-jss-react/src/components/RichText.tsx
```diff
@@ -40,6 +40,8 @@ export const RichText = forwardRef(
     const htmlProps = {
       dangerouslySetInnerHTML,
       ref,
+      // Experience Editor rewrites editable chrome markup client-side after SSR.
+      ...(field.editable && editable ? { suppressHydrationWarning: true } : {}),
```
illiakovalenko, line 44: Do we need to add this property? do we have a case when we have a hydration warning related to RichText component?

packages/sitecore-jss-react/src/components/Text.tsx
```diff
@@ -80,6 +80,11 @@ export const Text: React.FC<TextProps> = ({
     htmlProps.dangerouslySetInnerHTML = {
       __html: output,
     };
+
+    // Experience Editor rewrites editable chrome markup client-side after SSR.
+    if (isEditable) {
+      htmlProps.suppressHydrationWarning = true;
```
illiakovalenko: Do we need to add this property? do we have a case when we have a hydration warning related to Text component?

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
```diff
@@ -339,6 +339,10 @@ export class PlaceholderCommon<T extends PlaceholderProps> extends React.Compone
...
+    // attributes before React hydrates. Suppress the expected mismatch on these SDK nodes.
+    props.suppressHydrationWarning = true;
```
illiakovalenko: Is placeholder affected by hydration warnings? I think only the html blocks rendered by placeholder

packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
```diff
@@ -81,17 +88,32 @@ class ErrorBoundary extends React.Component<ErrorBoundaryProps> {
...
-        {this.props.children}
-      </Suspense>
-    );
+    // Experience Editor tags the outermost DOM node of each rendering (sc-part-of,
+    // scEnabledChrome) after SSR. Own that node in the SDK during editing so hydration
+    // mismatches can be suppressed without requiring suppressHydrationWarning on every
+    // app component. `display: contents` keeps SXA/Bootstrap column layout intact.
+    if (this.isPageEditing()) {
+      return (
+        <div suppressHydrationWarning style={{ display: 'contents' }}>
```
illiakovalenko, line 110: suppressHydrationWarning only works on the element carrying it, not on what's inside it
are we sure this change affects and can supress hydration warnings?

illiakovalenko (changes requested):

packages/sitecore-jss-react/src/components/Link.tsx
sc-naveenhedallaarachchi: Yes. the original EE hydration stack showed a mismatch on sc-link-wrapper, where editable chrome was rewritten (code/scpm → input) between SSR and hydrate. Suppress is only on the editable chrome path.

packages/sitecore-jss-react/src/components/RichText.tsx
sc-naveenhedallaarachchi: Yes. same stack showed RichText __html changing from scChromeData markup to an input field chrome. Suppress is only when rendering field.editable

packages/sitecore-jss-react/src/components/Text.tsx
sc-naveenhedallaarachchi: Confirmed by testing that removing it causes no warning, so it’s removed.

packages/sitecore-jss-react/src/components/ErrorBoundary.tsx
sc-naveenhedallaarachchi: Correct, suppressHydrationWarning does not cascade to children, and this wrap is not meant to cover nested mismatches.

The remaining warning was EE adding sc-part-of / scEnabledChrome on each rendering’s outermost DOM node (app roots like .component). In edit mode ErrorBoundary owns that outermost node in the SDK, so EE tags it instead of every app component, and suppress applies on that node. Nested field/placeholder chrome is handled separately.

Also verified by testing: removing this wrap brings those app-root warnings back; with it they are gone. display: contents is only to keep SXA/Bootstrap layout intact.

packages/sitecore-jss-react/src/components/PlaceholderCommon.tsx
sc-naveenhedallaarachchi: Yes, via the EE chrome HTML blocks Placeholder renders (createRawElement for chrometype="placeholder"), where phkey→key causes the mismatch. Moved suppress inside that check so it only applies there; removing it brings the warning back.

illiakovalenko (approved):
Looks good 👍

